PR #19, opened 2024-03-08
Remove CLIENT_GEN_VERSION and CONTROLLER_GEN_VERSION from devel image

ArangoGutierrez:
It is better to leave each repo to set CLIENT_GEN_VERSION and CONTROLLER_GEN_VERSION and simply use `k8s-test-infra` as a base image

### DIFF
--- a/images/devel/Dockerfile
+++ b/images/devel/Dockerfile
@@ -15,17 +15,12 @@
 ARG GOLANG_VERSION=x.x.x
 FROM golang:${GOLANG_VERSION}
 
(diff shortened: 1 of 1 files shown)


klueska:
Hmm. I don't think we necessarily need to remove these tools from this Dockerfile. If we know a specific pinned version of these tools works across all of our projects for a given golang version, then we should perfer to pull in this image. If this image doesn't work for a specific repo, it should have a fallback to be able to build a local image until a suitable image can be created here.

PR #35, opened 2024-04-05
add controller-gen and client-gen versions to 1.22 builds

foolusion:
This should add the proper controller-gen and client-gen versions to build with go 1.22. I'm not very familiar with github actions so let me know if there is a better way to include the defaults.


ArangoGutierrez:
Hey @foolusion thanks for the contribution, could you follow https://docs.pi-hole.net/guides/github/how-to-signoff/ to signoff your commit so the DCO check passes?

.github/workflows/image.yaml
```diff
@@ -91,4 +98,6 @@ jobs:
           platforms: ${{ env.IMAGE_PLATFORMS }}
           build-args: |
             GOLANG_VERSION=${{ matrix.go-version }}
+            CONTROLLER_GEN_VERSION=${{ matrix.controller-gen-version || "v0.9.2" }}
```
ArangoGutierrez, line 101: The workflow is not valid. .github/workflows/image.yaml (Line: 99, Col: 23): Unexpected symbol: '"v0'. Located at position 34 within expression: matrix.controller-gen-version || "v0.9.2"

ArangoGutierrez (changes requested):

.github/workflows/image.yaml
foolusion: Thanks, looks like you have to use single quotes inside the expressions.

PR #61, opened 2024-07-05
Add minimal ubi8 container

elezar:
This change adds tooling to produce a minimal container image.

The `redhat/ubi8-minimal` image is queried for a list of packages and the `nvidia/cuda:12.5.0-base-ubi8` image is updated to match the packages in the minimal image. 

This is done by:
* removing the `cuda-*` packages and repositories
* installing `microdnf`
* removing `dnf` and `python` packages
* removing packages that are not in the list of minimal packages


deployments/container/Dockerfile.ubi8
```diff
@@ -0,0 +1,43 @@
...
+    rm -rf /var/cache/dnf
+
+COPY --from=minimal /tmp/minimal-list.txt /tmp/minimal-list.txt
+COPY --from=minimal /tmp/minimal-list.versions /tmp/minimal-list.versions
+
+COPY cleanup.sh .
+RUN ./cleanup.sh
+
+RUN microdnf upgrade
+RUN rpm -qa | sort -u > /tmp/package-list.versions
```
tariq1890, line 43: Let's add an EOL here

deployments/container/Dockerfile.ubi8
```diff
@@ -0,0 +1,43 @@
...
+    dnf update -y && \
+    rm -rf /var/cache/dnf
+
+COPY --from=minimal /tmp/minimal-list.txt /tmp/minimal-list.txt
+COPY --from=minimal /tmp/minimal-list.versions /tmp/minimal-list.versions
+
+COPY cleanup.sh .
+RUN ./cleanup.sh
+
+RUN microdnf upgrade
```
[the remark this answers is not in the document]
elezar: I've updated things to do everything in the cleanup.sh script.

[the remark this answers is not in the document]
elezar: Moved into the `cleanup.sh` script`.

deployments/container/Dockerfile.ubi8
```diff
@@ -0,0 +1,32 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# We use the ubi8-minimal image as a reference image in removing unneeded dependencies.
+FROM redhat/ubi8-minimal:latest AS minimal
+
+RUN rpm -qa --queryformat='^%{NAME}-\[0-9\].*\.%{ARCH}$\n' | sort -u > /tmp/minimal-list.txt
```
tariq1890, line 18: Can we merge these two ?

deployments/container/cleanup/cleanup.sh
```diff
@@ -0,0 +1,48 @@
...
+microdnf remove $(rpm -q --whatrequires dnf)
+rpm -e dnf
+
+microdnf remove \
+    $(rpm -q --whatrequires /usr/libexec/platform-python) \
+    $(rpm -q --whatrequires 'python(abi)') \
+    python* \
+    dnf*
+
+microdnf remove \
```
shivamerla, line 40: does this ensure there are no broken dependencies? i.e remaining packages in the cuda:12.5.0-base-ubi8 might have dependencies on packages removed with the minimal list as a reference?

elezar: In my testing, `microdnf` still respects package dependencies, meaning that if `x` depends on `y`, one cannot remove `y`.

For example, let's say we want to remove `openssl` we see the following in the image:
```
$ microdnf remove openssl*
error: Could not depsolve transaction; 1 problem detected:
 Problem: package curl-7.61.1-34.el8.x86_64 from @System requires libcrypto.so.1.1()(64bit), but none of the providers can be installed
  - package curl-7.61.1-34.el8.x86_64 from @System requires libssl.so.1.1()(64bit), but none of the providers can be installed
  - conflicting requests
  - problem with installed package curl-7.61.1-34.el8.x86_64
```

deployments/container/Dockerfile.ubi8
elezar: I would prefer not to as I find two different commands more readable. I understand that we're generating two separate layers here, but this is in an intermediate image that doesn't affect the final image that we produce.

If we're really concerned about layers in the images, we could squash the image as a final step in our pipelines.

tariq1890 (approved):

elezar:
I have created https://github.com/NVIDIA/k8s-device-plugin/pull/813 with a POC for this work. Let's keep this open while we iterate on that change.

ArangoGutierrez:
> I have created [NVIDIA/k8s-device-plugin#813](https://github.com/NVIDIA/k8s-device-plugin/pull/813) with a POC for this work. Let's keep this open while we iterate on that change.

since  [NVIDIA/k8s-device-plugin#813](https://github.com/NVIDIA/k8s-device-plugin/pull/813) is merged, do we want to re ignite the conversation here, and potentially move from draft to ready-for-review?

PR #66, opened 2024-07-12
Add image for gh runners

ArangoGutierrez:


.github/workflows/image.yaml
```diff
@@ -29,6 +29,48 @@ on:
...
+          images: |
+            ghcr.io/nvidia/k8s-test-infra
+      - name: Build and push
+        uses: docker/build-push-action@v6
+        with:
+          push: ${{ env.IMAGE_PUSH }}
+          context: .
+          outputs: ${{ env.IMAGE_OUTPUTS }}
+          platforms: ${{ env.IMAGE_PLATFORMS }}
+          tags: ghcr.io/nvidia/k8s-test-infra:gh-runner
```
elezar, line 72: Are we interested in the version at all?

ArangoGutierrez: I was thinking that no, since we want latest versions from both holodeck and the runner.

ArangoGutierrez: how about now?

ArangoGutierrez: I like how it reads now

images/gh_runner/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+  lsb-release \
+  python3 \
+  sudo \
+  unzip \
+  zip \
+  && rm -rf /var/lib/apt/lists/*
+
+COPY --from=holodeck /go/bin/holodeck /usr/local/bin/holodeck
+
+USER runner
```
tariq1890, line 27: Is this a predefined user ?

ArangoGutierrez: Yes, is defined on the base image, see the link on the comment above the FROM

images/gh_runner/Dockerfile
```diff
@@ -0,0 +1,27 @@
+# Cache Holodeck image
+FROM ghcr.io/nvidia/holodeck:v0.1.1 AS holodeck
```
tariq1890, line 2: Can we use build args to provide the versions instead of hardcoding here?

.github/workflows/image.yaml
```diff
@@ -29,6 +29,52 @@ on:
...
+          if [ "${{github.event_name == 'pull_request'}}" = "true" ]; then
+            echo "IMAGE_PUSH=false" >> $GITHUB_ENV
+            echo "IMAGE_OUTPUTS=type=docker,dest=/tmp/k8s-test-infra-image.tar" >> $GITHUB_ENV
+            echo "IMAGE_PLATFORMS=linux/amd64" >> $GITHUB_ENV
+          else
+            echo "IMAGE_PUSH=true" >> $GITHUB_ENV
+            echo "IMAGE_OUTPUTS=type=registry,push=true" >> $GITHUB_ENV
+            echo "IMAGE_PLATFORMS=linux/amd64,linux/arm64" >> $GITHUB_ENV
+          fi
+          HOLODECK_VERSION=$(sed -n 's/.*holodeck:v\([0-9.]*\).*/\1/p' images/gh_runner/Dockerfile)
```
tariq1890, line 49: I think we should just define the versions as constants in `*.mk` or `Makefile` files and pass them as build args to the docker containers

images/gh_runner/Dockerfile
tariq1890: Got it, thanks

images/gh_runner/Dockerfile
ArangoGutierrez: The idea is to rely on dependabot for version management

.github/workflows/image.yaml
ArangoGutierrez: We want to rely on dependabot

PR #130, opened 2025-02-16
Update repo to serve as Go library

ArangoGutierrez:


images/devel/Dockerfile
elezar, line 1: Just a general question: How many of our projects depend on this image already? Have they all been migrated off it again?

images/gh_runner/Dockerfile
elezar, line 1: What was the purpose of this image?

versions.mk
```diff
@@ -11,4 +11,4 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-GOLANG_VERSION ?= 1.23.5
+GOLANG_VERSION ?= 1.24.0
```
elezar, line 14: Question: Could we use `hack/golang-version.sh` here instead?

elezar: (This would mean that we have a golang Dockerfile in `deployments/devel`).

ArangoGutierrez: sure, a way to leverage dependabot, do we agree that said image wouldn't require an `image` action, right?

ArangoGutierrez: Done, we now have a `deployments/devel` structure

ArangoGutierrez:
@tariq1890  @elezar  PTAL

images/devel/Dockerfile
ArangoGutierrez: The last one was DRA, and now all repos have in-repo solutions

ArangoGutierrez:
@tariq1890 @elezar PTAL

versions.mk
ArangoGutierrez: I have removed this file in favor of the `deployments/devel` structure

Makefile
```diff
@@ -50,3 +48,32 @@ vendor:
...
+	    echo "Verifying $$mod..."; ( \
+	        cd $$(dirname $$mod) && go mod verify | sed 's/^/  /g' \
+	    ) || exit 1; \
+	done
+
+check-modules: modules
+	@echo "- Checking if go.mod and go.sum are in sync..."
+	@git diff --exit-code -- $$(find . -name go.mod -name go.sum)
+	@echo "- Checking if the go mod vendor dir is in sync..."
+	@git diff --exit-code -- $$(find . -name vendor)
```
tariq1890, line 79: add EOL
```suggestion
	@git diff --exit-code -- $$(find . -name vendor)

```

ArangoGutierrez: thanks! done

tariq1890 (approved):

PR #193, opened 2025-11-24
Add mock libnvidia-ml.so implementation

ArangoGutierrez:
Implements a CGo-based mock NVML library. Includes Docker-based
build support for cross-platform development.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,6 @@
+# Build artifacts
+*.so
+*.dll
+*.dylib
+*.out
+*.test


pkg/gpu/mocknvml/engine/engine.go
```diff
@@ -0,0 +1,308 @@
...
+	if e.initialized {
+		e.initCount++
+		return nvml.SUCCESS
+	}
+
+	// Create base mock server
+	server := dgxa100.New()
+
+	// Apply CDI configuration if present
+	if e.config.CDISpecPath != "" {
```
elezar: Why does the NVML mock implementation require ANY CDI logic?

pkg/gpu/mocknvml/engine/engine.go
```diff
@@ -0,0 +1,308 @@
...
+
+	"github.com/google/uuid"
+
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+	"github.com/NVIDIA/go-nvml/pkg/nvml/mock/dgxa100"
+)
+
+// Engine manages the mock NVML implementation and provides a bridge between
+// C and Go NVML interfaces.
+type Engine struct {
```
elezar: Is the intent that `Engine` also implements the `nvml.Interface` interface?

pkg/gpu/mocknvml/bridge/bridge.go
```diff
@@ -0,0 +1,271 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+/*
+#include <stdlib.h>
+#include <string.h>
+
+typedef int nvmlReturn_t;
```
elezar: Why don't we reuse the `nvml.h` definitions that we already have?

pkg/gpu/mocknvml/engine/handles.go
```diff
@@ -0,0 +1,90 @@
...
+	counter uintptr
+	mu      sync.RWMutex
+}
+
+// NewHandleTable creates a new HandleTable.
+func NewHandleTable() *HandleTable {
+	return &HandleTable{
+		devices: make(map[uintptr]nvml.Device),
+		reverse: make(map[nvml.Device]uintptr),
+		counter: 1, // Start from 1, 0 reserved for NULL
```
elezar: When is this counter actually used? Why is it not sufficient to check `len(devices)` and `len(reverse)`?

tests/mocknvml/main.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	return strings.TrimRight(string(bytes), "\x00")
+}
+
+func main() {
+	log.Println("Starting Mock NVML Test")
+	log.Println("=================================")
+
+	// Initialize NVML
+	log.Println("Initializing NVML...")
```
ArangoGutierrez: This is an integration test to check that we can use `go-nvml` with the mock `libnvidia-ml.so`

PR #300, opened 2026-04-08
docs: streamline demo guides and remove manual node labelling

giuliocalzo:
## Summary

- Move Kind cluster config to a standalone `kind.yaml` file referenced from `demo.sh`, removing inline heredoc
- Apply node pool labels directly in the Kind config instead of a separate manual labelling step
- Mark the local image build step as optional, with a tip for overriding chart defaults
- Pass FGO topology inline via Helm values, removing the redundant manual ConfigMap apply step
- Renumber steps after consolidation (7 steps → 6 steps)

## Test plan

- [ ] Run `docs/demo/standalone/demo.sh` end-to-end on a fresh Kind cluster
- [x] Follow `docs/demo/with-fgo/README.md` steps manually and verify FGO integration

### DIFF
--- a/docs/demo/standalone/demo.sh
+++ b/docs/demo/standalone/demo.sh
@@ -20,17 +20,10 @@ info() { echo "==> $*"; }
 fail() { echo "ERROR: $*" >&2; exit 1; }
 
 ###############################################################################
-# Step 1 -- Create a 3-node Kind cluster
+# Step 1 -- Create a Kind cluster
 ###############################################################################
 info "Creating Kind cluster: ${CLUSTER_NAME}"
-cat <<EOF | kind create cluster --name "${CLUSTER_NAME}" --config=-
-kind: Cluster
-apiVersion: kind.x-k8s.io/v1alpha4
-nodes:
-  - role: control-plane
-  - role: worker
-  - role: worker
-EOF
+kind create cluster --name "${CLUSTER_NAME}" --config=$REPO_ROOT/docs/demo/kind.yaml
 
 ###############################################################################
 # Step 2 -- Build the nvml-mock image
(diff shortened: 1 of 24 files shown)


docs/demo/standalone/demo.sh
```diff
@@ -44,25 +37,6 @@ docker build -t "${IMAGE_NAME}" -f "${REPO_ROOT}/deployments/nvml-mock/Dockerfil
 info "Loading image into Kind cluster"
 kind load docker-image "${IMAGE_NAME}" --name "${CLUSTER_NAME}"
 
-###############################################################################
-# Step 4 -- Label worker nodes
-###############################################################################
-info "Labelling worker nodes"
-WORKERS=($(kubectl get nodes --no-headers -o custom-columns=":metadata.name" \
```
ArangoGutierrez, line 51: The summary at the end of the script still uses `${#WORKERS[@]}` (~line 115), but this was the only place `WORKERS` got defined. With `set -euo pipefail` that'll blow up at runtime.

Also worth renumbering the step comments below (5 through 9) since step 4 is gone now.

docs/demo/standalone/kind.yaml
```diff
@@ -0,0 +1,10 @@
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+  - role: control-plane
+  - role: worker
+    labels:
+      run.ai/simulated-gpu-node-pool: integration
+  - role: worker
+    labels:
+      run.ai/simulated-gpu-node-pool: scale
```
ArangoGutierrez: Nit: missing trailing newline.

docs/demo/with-fgo/README.md
```diff
@@ -23,88 +23,73 @@ apiVersion: kind.x-k8s.io/v1alpha4
...
   --set integrations.fakeGpuOperator.enabled=true \
   --set gpu.profile=h100 \
   --set gpu.count=8 \
   --set "nodeSelector.run\.ai/simulated-gpu-node-pool=integration" \
   --wait --timeout 120s
 ```
 
+> **Tip:** To use the locally built image from Step 2, add `--set image.repository=nvml-mock --set image.tag=demo` to the command above.
+
 ## Step 5 -- Install fake-gpu-operator
```
ArangoGutierrez: Should be Step 4 now that the labels step was removed — same for Step 6 (Verify) below, should be Step 5.

docs/demo/with-fgo/README.md
```diff
@@ -23,88 +23,73 @@ apiVersion: kind.x-k8s.io/v1alpha4
...
-      integration:
+echo "Installing fake-gpu-operator"
+helm upgrade --install gpu-operator  oci://ghcr.io/run-ai/fake-gpu-operator/fake-gpu-operator -n gpu-operator --create-namespace \
+  --wait --timeout 120s  -f - <<EOF
+topology:
+    nodeGroups:
+      - name: integration
         backend: mock
+        nodeSelector:
+          "nodeGroup": "gpu"
```
ArangoGutierrez: The Kind config uses `run.ai/simulated-gpu-node-pool` labels but this selector references `nodeGroup: gpu` — no node will match. Is this the right key for current FGO?

docs/demo/with-fgo/README.md
```diff
@@ -23,88 +23,73 @@ apiVersion: kind.x-k8s.io/v1alpha4
...
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: fake-gpu-operator-topology
-  namespace: gpu-operator
-data:
-  topology.yaml: |
-    nodePools:
-      integration:
+echo "Installing fake-gpu-operator"
```
ArangoGutierrez: This `echo` looks like a leftover from a script — probably doesn't belong in a README walkthrough. Also there's a stray blank line at the top of the code block.

ArangoGutierrez (changes requested):
Nice cleanup — moving labels into the Kind config is much cleaner. A few things to address though.

giuliocalzo:
@ArangoGutierrez anything else missing?

ArangoGutierrez (approved):

PR #302, opened 2026-04-10
ci: publish Helm chart as OCI artifact on merge to main

giuliocalzo:
## Summary

- Add `push` trigger on `main` to `.github/workflows/helm.yaml` so the workflow also runs when a PR is merged (not just on PRs)
- Add `publish` job that runs only on push to main, gated behind `lint` and `unittest` passing
- The publish job packages the chart, pushes it to `oci://ghcr.io/nvidia/k8s-test-infra` as an OCI artifact, and signs it with cosign (keyless)
- Extract `CHART_NAME` and `CHART_DIR` into workflow-level env vars for DRY references

After merge, the chart is installable via:
```
helm install nvml-mock oci://ghcr.io/nvidia/k8s-test-infra/nvml-mock --version <version>
```

## Test plan

- [x] Verify `lint` and `unittest` jobs still pass on PRs
- [x] Verify `publish` job is skipped on PRs (`if: github.event_name == 'push'`)
- [x] Merge a chart change to main and confirm the OCI artifact appears in GHCR packages


.github/workflows/helm.yaml
```diff
@@ -46,4 +53,46 @@ jobs:
...
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    needs: [lint, unittest]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      id-token: write
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
```
ArangoGutierrez, line 68: You are pinning SHAs here but lint/unittest still use tags. Worth aligning while you are in the file.

.github/workflows/helm.yaml
```diff
@@ -14,6 +18,9 @@ concurrency:
 permissions:
   contents: read
 
+env:
+  CHART_NAME: nvml-mock
+  CHART_DIR: deployments/nvml-mock/helm
```
ArangoGutierrez, line 23: `ct lint` still hardcodes the chart path in the lint job — might as well use `${{ env.CHART_DIR }}` there too.

.github/workflows/helm.yaml
```diff
@@ -46,4 +53,46 @@ jobs:
...
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Set up Helm
+        uses: azure/setup-helm@1a275c3b69536ee54be43f2070a358922e12c8d4 # v4
+
+      - name: Read chart version
+        id: chart
+        run: |
+          version=$(grep '^version:' ${{ env.CHART_DIR }}/${{ env.CHART_NAME }}/Chart.yaml | awk '{print $2}')
```
ArangoGutierrez: Will break on quoted versions (`version: "1.2.3"`). `yq .version` would be safer here.

.github/workflows/helm.yaml
```diff
@@ -46,4 +53,46 @@ jobs:
...
+          echo "version=${version}" >> "$GITHUB_OUTPUT"
+
+      - name: Log in to GHCR
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | helm registry login ghcr.io -u ${{ github.actor }} --password-stdin
+
+      - name: Package chart
+        run: helm package ${{ env.CHART_DIR }}/${{ env.CHART_NAME }} --destination .cr-release-packages
+
+      - name: Push chart to OCI registry
+        run: helm push .cr-release-packages/${{ env.CHART_NAME }}-${{ steps.chart.outputs.version }}.tgz oci://ghcr.io/nvidia/k8s-test-infra/chart
```
ArangoGutierrez, line 86: This pushes to `.../chart/nvml-mock` but the PR description shows `oci://ghcr.io/nvidia/k8s-test-infra/nvml-mock`. Which one are we going with?

.github/workflows/helm.yaml
```diff
@@ -46,4 +53,46 @@ jobs:
...
+
+      - name: Install cosign
+        uses: sigstore/cosign-installer@c56c2d3e59e4281cc41dea2217323ba5694b171e # v3.8.0
+
+      - name: Sign chart package
+        run: |
+          cosign sign-blob --yes \
```
ArangoGutierrez: `sign-blob` writes the .sig and .cert locally — they are gone when the runner exits. You want `cosign sign` on the OCI artifact so the signature lives in the registry and consumers can actually verify.

.github/workflows/helm.yaml
```diff
@@ -46,4 +53,46 @@ jobs:
...
+            --output-signature .cr-release-packages/${{ env.CHART_NAME }}-${{ steps.chart.outputs.version }}.tgz.sig \
+            --output-certificate .cr-release-packages/${{ env.CHART_NAME }}-${{ steps.chart.outputs.version }}.tgz.cert \
+            .cr-release-packages/${{ env.CHART_NAME }}-${{ steps.chart.outputs.version }}.tgz
+        env:
+          COSIGN_EXPERIMENTAL: "true"
```
ArangoGutierrez: Deprecated since cosign v2, keyless is default now. `--yes` is enough.

deployments/nvml-mock/Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN cd pkg/gpu/mockcuda && make clean && make
 
 FROM debian:bookworm-slim
 ARG KUBECTL_VERSION=v1.32.0
-ARG TARGETARCH=amd64
+ARG TARGETARCH
```
ArangoGutierrez, line 27: Good fix but unrelated to Helm publish — separate commit?

ArangoGutierrez (changes requested):
Nice — been wanting OCI publish for the chart. One thing to sort out: the cosign signing does not actually persist anywhere useful right now. Details inline.

.github/workflows/helm.yaml
giuliocalzo: yq is not installaed by default I've used a different approach

deployments/nvml-mock/Dockerfile
giuliocalzo: this come from a rebase... not sure why... let me fix it

.github/workflows/helm.yaml
giuliocalzo: helm plugin install does not support the sha version I can only use the tag

ArangoGutierrez (approved):
All review comments addressed:
- cosign sign on OCI artifact (not sign-blob) ✓
- --experimental removed, --yes only ✓
- CHART_DIR env var used in ct lint ✓
- helm-unittest SHA noted in comment (plugin doesn't support SHA install) ✓
- Dockerfile rebase artifact removed ✓

Minor follow-ups to be filed as issues (version parsing, OCI path docs alignment). LGTM.

PR #306, opened 2026-04-15
Mock NVML: hex UUIDs, IsMigDeviceHandle, SONAME, device visibility filtering

dims:
Four changes to make mock NVML work with the DRA driver's CDI spec generation pipeline on CPU-only nodes:

1. **Valid hex UUIDs** for all GPU profiles (B200, GB200, H100, L40S, T4). Non-hex prefixes like `GPU-GB200-` are rejected by nvidia-container-toolkit's CDI spec generator. Also fixes the Helm README example.

2. **`nvmlDeviceIsMigDeviceHandle`** implementation -- returns `SUCCESS` with `isMigDevice=0` (mock devices are always full GPUs). Without this, CDI spec generation fails with `ERROR_NOT_SUPPORTED`.

3. **ELF SONAME `libnvidia-ml.so.1`** via `-Wl,-soname` linker flag. Without the SONAME, the CDI `create-symlinks` hook only creates `.so` but not `.so.1`, and nvidia-smi (which dlopen's `libnvidia-ml.so.1`) fails inside workload pods.

4. **Device visibility filtering** by `/dev/nvidia*` presence. On real hardware, NVML filters visible GPUs via cgroup device permissions. The mock now scans for `/dev/nvidia<N>` nodes during `nvmlInit()` and filters `DeviceGetCount`/`DeviceGetHandleByIndex` accordingly. When CDI injects only `/dev/nvidia0` into a workload pod, nvidia-smi correctly shows 1 GPU instead of all 8.


pkg/gpu/mocknvml/engine/engine.go
```diff
@@ -407,3 +431,36 @@ func ResetForTesting() {
...
 }
+
+// detectVisibleDevices scans for /dev/nvidia<N> device nodes and returns a
+// slice mapping visible indices to actual device indices. This mimics real
+// NVML behavior where cgroup device permissions limit which GPUs a container
+// can see.
+//
+// Returns nil if ALL device nodes exist (no filtering needed) or if none
+// exist (host context where /dev/nvidia* may not be present but NVML should
+// still work).
```
ArangoGutierrez, line 450: Needs tests. Pure function, easy to cover with temp device nodes or at minimum verify the nil-return logic for all-present / none-present. The visibility remapping in `DeviceGetHandleByIndex` and the count override also need engine-level smoke tests -- without coverage this is fragile to refactor later.

pkg/gpu/mocknvml/engine/device.go
```diff
@@ -1236,11 +1236,30 @@ func parseComputeMode(mode string) nvml.ComputeMode {
 type MockServer struct {
 	*dgxa100.Server
 	configurableDevices [MaxDevices]*ConfigurableDevice
+
+	// visibleDevices maps "visible index" (0, 1, ...) to the actual device
+	// index in configurableDevices. When a container has only a subset of
+	// /dev/nvidia* nodes (via CDI injection), this slice contains only the
+	// devices whose device nodes exist, mimicking real NVML's cgroup-based
+	// device filtering. If nil, all devices are visible (no filtering).
+	visibleDevices []int
```
ArangoGutierrez, line 1246: `DeviceGetHandleByUUID` and `DeviceGetHandleByPciBusId` (~line 1272) iterate `configurableDevices` directly and will return devices outside the visible set. Real NVML filters all lookup paths, not just index-based. A container with only `/dev/nvidia0` shouldn't resolve GPU-1's UUID.

pkg/gpu/mocknvml/engine/device.go
```diff
@@ -1236,11 +1236,30 @@ func parseComputeMode(mode string) nvml.ComputeMode {
...
 func (s *MockServer) DeviceGetHandleByIndex(index int) (nvml.Device, nvml.Return) {
 	debugLog("[NVML] nvmlDeviceGetHandleByIndex(%d)\n", index)
+
+	// Map through visibility filter if active
+	if s.visibleDevices != nil {
+		if index < 0 || index >= len(s.visibleDevices) {
+			return nil, nvml.ERROR_INVALID_ARGUMENT
+		}
+		actual := s.visibleDevices[index]
+		return s.configurableDevices[actual], nvml.SUCCESS
```
ArangoGutierrez, line 1264: Unfiltered path below checks `configurableDevices[index] == nil` -> `ERROR_NOT_FOUND`. This path skips it. Add the nil guard for parity.

pkg/gpu/mocknvml/Makefile
```diff
@@ -53,8 +53,12 @@ VERSION_ALIASES := \
 	-Wl,--defsym,nvmlVgpuInstanceGetLicenseInfo=nvmlVgpuInstanceGetLicenseInfo_v2 \
 	-Wl,--defsym,nvmlDeviceGetDriverModel=nvmlDeviceGetDriverModel_v2
 
+# Set SONAME so nvidia-container-toolkit CDI spec generation creates the
+# libnvidia-ml.so.1 symlink (nvidia-smi dlopen's this soname).
+SONAME_FLAG := -Wl,-soname,$(LIB_SONAME)
+
 $(LIB_REALNAME):
-	CGO_LDFLAGS_ALLOW='--defsym' CGO_LDFLAGS='$(VERSION_ALIASES)' \
+	CGO_LDFLAGS_ALLOW='--defsym|.*-soname.*' CGO_LDFLAGS='$(VERSION_ALIASES) $(SONAME_FLAG)' \
```
ArangoGutierrez: `.*-soname.*` is wider than needed -- `'--defsym|-soname'` is sufficient.

ArangoGutierrez (changes requested):
Good stuff -- each fix clearly motivated by a real CDI failure, PR description is solid.

Two things to address, plus a couple nits. Also `docs/examples.md:163` still has `GPU-A100-0000` (non-hex UUID), same bug class you're fixing here.

pkg/gpu/mocknvml/engine/engine.go
dims: Adding tests in for detectVisibleDevices nil-return logic (using temp nodes) and engine-level smoke tests for count/index/UUID/PCI visibility filtering.

pkg/gpu/mocknvml/engine/device.go
dims: Fixed — both lookups now check isDeviceVisible(i) and return ERROR_NOT_FOUND for non-visible devices. Tests added.

pkg/gpu/mocknvml/engine/device.go
dims: Adding nil guard for parity with the unfiltered path.

pkg/gpu/mocknvml/Makefile
dims: Tightening to '--defsym|-soname'

ArangoGutierrez (approved):

PR #315, opened 2026-04-27
feat(nvml-mock): add InfiniBand sysfs mock

giuliocalzo:
## What This PR Does

Adds a fake InfiniBand fabric to the nvml-mock DaemonSet so real userspace tools (`ibstat`, `ibstatus`, `iblinkinfo`, libibverbs consumers, ...) report a realistic per-HCA topology on hosts with no IB hardware. Same model as the existing `nvidia-smi` mock — real binary backed by a fake kernel surface.

```bash
$ kubectl exec nvml-mock-xxxxx -- ibstat
CA 'mlx5_0'
        CA type: MT4129
        Number of ports: 1
        Firmware version: 28.39.2048
        Node GUID: 0xa088c20300ab0000
        Port 1:
                State: Active
                Physical state: LinkUp
                Rate: 400
                Link layer: InfiniBand
... (one CA per GPU)
```

### Architecture

Two pieces in `pkg/network/mockibsysfs/`:

- **`libibmocksys.so` (LD_PRELOAD shim, C)** — intercepts ~15 libc filesystem functions (`open`/`openat`/`opendir`/`scandir`/`stat`/`statx`/`chdir`/...) and rewrites paths under `/sys/class/infiniband*`, `/sys/class/infiniband_mad/`, `/sys/class/infiniband_verbs/`, and `/dev/infiniband` to point at `$MOCK_IB_ROOT`. Everything else passes through to real libc via `dlsym(RTLD_NEXT, ...)`. Set `MOCK_IB_DISABLE=1` to bypass.
- **`render-ib-sysfs` (Go binary)** — reads the new `infiniband:` block of each mock-nvml profile YAML and writes a kernel-faithful tree: per-CA `node_type`/`node_guid`/`sys_image_guid`/`fw_ver`/`hw_rev`/`board_id`/`hca_type`/`node_desc`; per-port `state`/`phys_state`/`rate`/`lid`/`sm_lid`/`cap_mask`/`link_layer`/`port_guid`/`gids/0`/`gid_attrs/`/`pkeys/0`/`counters/*`; plus `infiniband_mad/{abi_version,umad*,issm*}` for libibumad and `infiniband_verbs/{abi_version,uverbs*}` for libibverbs.

Why C for the shim (not Go): a Go `c-shared` library would inject the entire Go runtime (scheduler, GC, signal handlers, ~5–10 MB resident) into every process loaded with LD_PRELOAD, hijack signal handlers (SIGSEGV/SIGURG/SIGPROF), and risk re-entrancy with the runtime's own `open`/`stat` calls during init. Variadic and function-pointer signatures (`open(..., flags, ...)`, `scandir`, `statx`) cannot be expressed cleanly through cgo. C trampolines forwarding to `dlsym(RTLD_NEXT, ...)` are the textbook shape for this. The renderer (one-shot, complex logic, runs at startup) stays in Go.

### Wiring

- Dockerfile builds both pieces and installs `infiniband-diags` + `rdma-core` for the real userspace tools.
- `setup.sh` invokes `render-ib-sysfs` after the GPU mock setup.
- DaemonSet exports `LD_PRELOAD=/usr/local/lib/libibmocksys.so.1` and `MOCK_IB_ROOT=/var/lib/nvml-mock/ib` so every process in the container — including `kubectl exec ... ibstat` — sees the fake fabric.

### Profile defaults

| Profile  | Enabled | HCA model              | Speed        | HCAs / GPU |
|----------|---------|------------------------|--------------|------------|
| `a100`   | yes     | ConnectX-6 (`MT4123`)  | HDR 200 Gb/s | 1          |
| `h100`   | yes     | ConnectX-7 (`MT4129`)  | NDR 400 Gb/s | 1          |
| `b200`   | yes     | ConnectX-7 (`MT4129`)  | NDR 400 Gb/s | 1          |
| `gb200`  | yes     | ConnectX-7 (`MT4129`)  | NDR 400 Gb/s | 1          |
| `l40s`   | no      | —                      | —            | —          |
| `t4`     | no      | —                      | —            | —          |

All knobs (`hca_type`, `fw_version`, `link_layer`, `rate_gbps`, `port_state`, `phys_state`, `hcas_per_gpu`, `hca_count`, `guid_prefix`, `node_desc_template`) can be overridden per profile or via `gpu.customConfig`.

## Why

The mock-nvml stack covers GPU discovery very thoroughly, but most NCCL / training workloads also probe the InfiniBand fabric (`ibv_get_device_list`, `ibstat`, NCCL's `nccl-tests`, k8s GPU Operator's network operator validators). Today these probes either fail or are skipped in CI, which means we can't exercise GPU + IB topology end-to-end against the mock. This PR closes that gap with the same "real-tool-on-fake-kernel" approach that already works for `nvidia-smi`.

## Rebase notes

The branch was rebased on `upstream/main` (latest: `87ae8025`) to clear the `BLOCKED` state:

- **#326 (DRA quick-start: co-locate dev nodes under driver root)** — accepted upstream's `DEV_ROOT=$DRIVER_ROOT/dev` change in `setup.sh` and re-applied the IB rendering block on top.
- **#323 (dynamic metrics: temperature, power, utilization)** — kept upstream's `MOCK_NVML_CONFIG` env var and `/etc/nvml-mock` configmap mount; layered the IB-mock `LD_PRELOAD` / `MOCK_IB_ROOT` env vars and the second `host-nvml-mock` mount alongside.
- **`helm-unittest --verify=false`** — landed upstream as `ce6dcfc0`, so the previously-included CI commit dropped automatically during rebase. This PR is now scoped purely to the InfiniBand mock.
- **#306 (Mock NVML: hex UUIDs, IsMigDeviceHandle, SONAME, device visibility filtering)** — refreshed the helm `__snapshot__/*.yaml.snap` files to match the post-#306 / post-#323 daemonset shape. The diff that looks like "snapshot churn" is the merge of three orthogonal changes (UUID format, `MOCK_NVML_CONFIG` env, IB env + double-mount) into a single rendered baseline.

## Review feedback addressed (commit `8b0059d9`)

Blockers from the changes-requested review (#315):

- **`gid_attrs` as directory** (`render.go`): real Linux sysfs has it as a directory containing per-GID attribute files. Was previously written as an empty regular file (would yield `ENOTDIR` for libibverbs / iblinkinfo `opendir()`); now created via `mkdirAll`. Test added.
- **`port_guid` wired in** (`render.go`): `perHCAPortGUID` is now written to `ports/1/port_guid` instead of being computed and discarded. Real mlx5 ports expose this alongside `node_guid` (lower 16 bits differ to mimic the U/L bit). Test added.
- **Error wrapping** (`render.go`): `mkdirAll` and `writeFile` now return `fmt.Errorf("mkdir %s: %w", rel, err)` / `fmt.Errorf("write %s: %w", rel, err)` so a permission-denied at deep paths surfaces with call-site context instead of a bare `os` error.
- **`setup.sh` failure handling**: removed the `|| echo "WARNING: render-ib-sysfs failed (non-fatal)"` swallowing; failures are fatal under `set -e` (a profile-YAML typo would otherwise silently produce an empty IB tree). Also fixed the post-rebase `--config` path to `/etc/nvml-mock/config.yaml`.

Non-blocking nits also addressed:

- **Deterministic file order**: `caFiles` / `portFiles` converted from `map[string]string` to slices of `(name, value)` pairs (Go map iteration is randomized).
- **`strconv.Itoa`**: replaced the hand-rolled `itoa` test helper.
- **`shim.c`**: replaced the glibc-internal `__O_TMPFILE` with public `O_TMPFILE` matching its full bit pattern (it includes `O_DIRECTORY`); switched the `va_arg` mode extraction to `(mode_t)va_arg(ap, unsigned int)` matching glibc's own `fcntl.h` inline `open()` pattern.
- **Integration test**: now also asserts on per-HCA counter files (`port_xmit_data`, `port_rcv_data`, `port_xmit_packets`) and `infiniband_mad/abi_version` so non-`ibstat` consumers (perfquery, iblinkinfo, libibumad init) are exercised end-to-end.
- **`TestRender_Disabled`**: now populates every `Infiniband` field except `Enabled`, so a regression where `Enabled` is ignored and `Render()` proceeds based on the rest of the struct alone is now caught.
- **`Chart.yaml`**: added `kubeVersion: ">= 1.28.0-0"`. The two-mountPath pattern for one hostPath has worked since k8s 1.0, but pinning a floor surfaces the requirement. The `daemonset.yaml` comment now points at `kubeVersion` rather than reading as "this is unusual".

## Tests

- Unit tests for the renderer (defaults, GUID normalization, count override, rate-string mapping, error paths, `gid_attrs` directory, `port_guid`, disabled-with-fully-populated-struct) — `go test ./pkg/network/mockibsysfs/...` (passes).
- Integration test (`-tags=integration`, Linux-only, skipped otherwise) builds the shim, renders a tree, runs real `ibstat` from `infiniband-diags`, and asserts on output, per-HCA counter files, and `infiniband_mad/abi_version`. Verified passing in `linux/amd64` Docker.
- Helm unittest snapshots regenerated; all 79 tests / 14 snapshots pass post-rebase.
- End-to-end smoke against the actual built image: `setup.sh` followed by `ibstat` / `ibstatus` produces correct output for the H100 default (8x ConnectX-7 NDR HCAs).

## Checklist

- [x] Tests pass (`go test -race ./pkg/network/mockibsysfs/... ./cmd/render-ib-sysfs/...`)
- [x] Helm unittest passes (`helm unittest deployments/nvml-mock/helm/nvml-mock` — 79 tests, 14 snapshots)
- [ ] Linter passes (`golangci-lint run`) — not installed locally; relying on CI
- [x] Documentation updated (Helm chart README, `pkg/network/mockibsysfs/README.md`, standalone + with-fgo demos)
- [x] CHANGELOG.md updated under `[Unreleased]`

### DIFF
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -50,7 +50,7 @@ jobs:
         uses: azure/setup-helm@dda3372f752e03dde6b3237bc9431cdc2f7a02a2 # v5
 
       - name: Install helm-unittest plugin
-        run: helm plugin install https://github.com/helm-unittest/helm-unittest.git --version v1.0.3 # 6f82a998e0b5461762ca959f87f5dd344af5e4eb
+        run: helm plugin install --verify=false https://github.com/helm-unittest/helm-unittest.git --version v1.0.3 # 6f82a998e0b5461762ca959f87f5dd344af5e4eb
 
       - name: Run helm unittest
         run: helm unittest ${{ env.CHART_DIR }}/${{ env.CHART_NAME }}
--- a/.github/workflows/nvml-mock-e2e.yaml
+++ b/.github/workflows/nvml-mock-e2e.yaml
@@ -130,6 +130,15 @@ jobs:
             "${{ steps.gpu-info.outputs.gpu-name }}" \
             2
 
+      - name: Validate ibstat / ibstatus (InfiniBand mock)
+        run: |
+          chmod +x tests/e2e/validate-ibstat.sh
+          POD=$(kubectl get pods -l app.kubernetes.io/name=nvml-mock \
+            -o jsonpath='{.items[0].metadata.name}')
+          # gpu.count=2 in this job; profiles ship hcas_per_gpu=1, so
+          # IB-enabled profiles are expected to expose 2 HCAs.
+          ./tests/e2e/validate-ibstat.sh "$POD" "${{ matrix.gpu-profile }}" 2
+
       - name: Deploy NVIDIA device plugin
         run: |
           kubectl apply -f tests/e2e/device-plugin-mock.yaml
@@ -703,6 +712,18 @@ jobs:
           echo "=== Worker-2 (T4) ==="
           docker exec "${{ steps.nodes.outputs.worker2 }}" ls -la /var/lib/nvml-mock/driver/usr/lib64/libnvidia-ml.so.* | head -2
           docker exec "${{ steps.nodes.outputs.worker2 }}" grep "NVIDIA T4" /var/lib/nvml-mock/config/config.yaml | head -1
+
+      - name: Validate ibstat / ibstatus on both workers
+        run: |
+          chmod +x tests/e2e/validate-ibstat.sh
+          # A100 release: IB enabled, expect gpu.count=2 HCAs.
+          A100_POD=$(kubectl get pods -l app.kubernetes.io/instance=nvml-mock-a100 \
+            -o jsonpath='{.items[0].metadata.name}')
+          ./tests/e2e/validate-ibstat.sh "$A100_POD" a100 2
+          # T4 release: IB disabled by default, expect zero HCAs.
+          T4_POD=$(kubectl get pods -l app.kubernetes.io/instance=nvml-mock-t4 \
+            -o jsonpath='{.items[0].metadata.name}')
+          ./tests/e2e/validate-ibstat.sh "$T4_POD" t4 2
 
       - name: Deploy NVIDIA device plugin
         run: |
--- a/.github/workflows/nvml-mock-e2e.yaml
+++ b/.github/workflows/nvml-mock-e2e.yaml
@@ -131,6 +131,15 @@ jobs:
             "${{ steps.gpu-info.outputs.gpu-name }}" \
             2
 
+      - name: Validate ibstat / ibstatus (InfiniBand mock)
+        run: |
+          chmod +x tests/e2e/validate-ibstat.sh
+          POD=$(kubectl get pods -l app.kubernetes.io/name=nvml-mock \
+            -o jsonpath='{.items[0].metadata.name}')
+          # gpu.count=2 in this job; profiles ship hcas_per_gpu=1, so
+          # IB-enabled profiles are expected to expose 2 HCAs.
+          ./tests/e2e/validate-ibstat.sh "$POD" "${{ matrix.gpu-profile }}" 2
+
       - name: Deploy NVIDIA device plugin
         run: |
           kubectl apply -f tests/e2e/device-plugin-mock.yaml
@@ -756,6 +765,18 @@ jobs:
           echo "=== Worker-2 (T4) ==="
           docker exec "${{ steps.nodes.outputs.worker2 }}" ls -la /var/lib/nvml-mock/driver/usr/lib64/libnvidia-ml.so.* | head -2
           docker exec "${{ steps.nodes.outputs.worker2 }}" grep "NVIDIA T4" /var/lib/nvml-mock/config/config.yaml | head -1
+
+      - name: Validate ibstat / ibstatus on both workers
+        run: |
+          chmod +x tests/e2e/validate-ibstat.sh
+          # A100 release: IB enabled, expect gpu.count=2 HCAs.
+          A100_POD=$(kubectl get pods -l app.kubernetes.io/instance=nvml-mock-a100 \
+            -o jsonpath='{.items[0].metadata.name}')
+          ./tests/e2e/validate-ibstat.sh "$A100_POD" a100 2
+          # T4 release: IB disabled by default, expect zero HCAs.
+          T4_POD=$(kubectl get pods -l app.kubernetes.io/instance=nvml-mock-t4 \
+            -o jsonpath='{.items[0].metadata.name}')
+          ./tests/e2e/validate-ibstat.sh "$T4_POD" t4 2
 
       - name: Deploy NVIDIA device plugin
         run: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- InfiniBand mock: real `ibstat`, `ibstatus`, `iblinkinfo`, and other
+  `infiniband-diags` / `rdma-core` tools now work inside the nvml-mock
+  DaemonSet without IB hardware. Implementation: `LD_PRELOAD` shim
+  (`libibmocksys.so`) redirects libc filesystem accesses against
+  `/sys/class/infiniband*` and `/dev/infiniband` to a fake tree rendered
+  at startup by `render-ib-sysfs` from each profile's new `infiniband:`
+  block. Defaults: `a100` -> ConnectX-6 HDR; `h100` / `b200` / `gb200`
+  -> ConnectX-7 NDR; `l40s` / `t4` -> disabled.
+
 ## [0.1.0] - 2026-04-07
 
 ### Added
--- a/cmd/render-ib-sysfs/main.go
+++ b/cmd/render-ib-sysfs/main.go
@@ -1,0 +1,75 @@
+// Copyright 2026 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+// Command render-ib-sysfs reads the `infiniband:` block from a mock-nvml
+// profile YAML and writes a fake InfiniBand sysfs tree under --output.
+//
+// Pair with libibmocksys.so (LD_PRELOAD) so that real userspace tooling
+// (ibstat, ibstatus, iblinkinfo, libibverbs consumers, ...) reads from the
+// rendered tree instead of /sys directly.
+//
+// Usage:
+//
+//	render-ib-sysfs \
+//	    --config /config/config.yaml \
+//	    --gpu-count "$GPU_COUNT" \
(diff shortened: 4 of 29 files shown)


ArangoGutierrez (changes requested):
Solid LD_PRELOAD design — the C-shim choice is well-justified, the engine/renderer separation is clean, and the mock is convincing enough to pass real `ibstat`/`ibstatus` integration. The rate/state-format helpers and GUID normalization are tight. Three things keep this from approve: (1) hard merge conflict with #326's setup.sh refactor (your branch reverts the `DEV_ROOT=$DRIVER_ROOT/dev` change — needs rebase + reconcile, this is the BLOCKED you're seeing); (2) the `helm plugin install --verify=false` change is unrelated supply-chain regression that should be its own PR with a real upstream-issue link; (3) a small cluster of code-quality items in render.go (`gid_attrs` as file vs dir, dead `perHCAPortGUID`, missing error wrapping, swallowed setup.sh failure). Tests are real (not theater) and the integration test against actual `ibstat` is exactly the right shape. Snapshot regeneration to fix #306's drift is appropriate but should be called out in the PR body.

---

## Blockers

**1. `deployments/nvml-mock/scripts/setup.sh` (around line 13)**

Hard merge conflict with #326 (already merged). Branch base reverts #326's `DEV_ROOT=$DRIVER_ROOT/dev` co-location and the corresponding CDI hostPath edits. This is the actual cause of the BLOCKED state, not just a stale base. Rebase on `upstream/main`, keep #326's DRA-driver path co-location, and append the IB_ROOT rendering block at the bottom only.

**2. `.github/workflows/helm.yaml` (around line 53)**

Adding `--verify=false` to `helm plugin install` is an unrelated CI/supply-chain change with no link to an upstream issue or plan to re-enable. This silently weakens supply-chain integrity and violates 'one concern per PR'. Two requests: (a) extract the helm.yaml change into its own PR with a link to the upstream helm-unittest signing issue and a re-enable plan; (b) keep this PR scoped to InfiniBand mocking. Per security.md ('Sigstore/cosign for prod images'), removing signature verification needs more justification than 'avoid CI failures' — name the failure mode (cert chain? key rotation? upstream not signing?).

**3. `pkg/network/mockibsysfs/render/render.go` (around line 1561)**

`portFiles["gid_attrs"] = ""` writes a regular empty file at `.../ports/1/gid_attrs`. In real Linux sysfs, `gid_attrs` is a directory containing per-GID attribute files. `opendir(gid_attrs)` on this mock will return ENOTDIR. `ibstat` doesn't read it, but `libibverbs`/`iblinkinfo` may. Either drop the entry entirely or use `mkdirAll(portDir, "gid_attrs")`. Tests do not currently cover the gid_attrs branch.

**4. `pkg/network/mockibsysfs/render/render.go` (around line 1714)**

`portGUID := perHCAPortGUID(...)` is computed and immediately discarded with `_ = portGUID`. Dead code. Either wire it into a port-specific file (e.g., `ports/1/port_guid` which real mlx5 ports do expose) or remove the function and the discard. Leaving it as scaffold for 'future use' violates YAGNI.

**5. `pkg/network/mockibsysfs/render/render.go` (around line 1722)**

`mkdirAll` and `writeFile` return raw os errors with no call-site context. When `os.WriteFile` fails inside `renderHCA`, the user sees `open /tmp/.../sys/class/infiniband/mlx5_3/cap_mask: permission denied` with no hint that this came from the IB mock. Wrap with `fmt.Errorf("write %s: %w", rel, err)` per go-conventions.md.

**6. `deployments/nvml-mock/scripts/setup.sh` (around line 207)**

`|| echo "WARNING: render-ib-sysfs failed (non-fatal)"` swallows render failures silently. The rest of `setup.sh` uses `set -e` and exits on errors. Either (a) remove the `|| echo` and let render failures abort startup (failing fast is correct for a startup script), or (b) document why a partial render is acceptable. As written, a profile YAML typo will silently produce an empty IB tree and `ibstat` will report zero HCAs with no signal in pod logs except a single warning line.

---

## Non-blocking nits

**1. `pkg/network/mockibsysfs/render/render.go` (around line 1561)**

`caFiles` and `portFiles` are `map[string]string` iterated for file creation. Map iteration order is randomized in Go, so file creation order is non-deterministic. Functionally fine, but consider a slice of `{name, value}` pairs for reproducible build artifacts.

**2. `pkg/network/mockibsysfs/render/render_test.go` (around line 1885)**

Hand-rolled `itoa` helper duplicates `strconv.Itoa`. Replace with `strconv.Itoa(i)`.

**3. `pkg/network/mockibsysfs/shim.c` (around line 2033)**

`__O_TMPFILE` is a glibc-internal symbol. Cleaner: `(flags & O_TMPFILE) == O_TMPFILE`. Also: `extract_mode` takes `va_arg(int)` — POSIX mode is `mode_t`. Default arg promotion makes this work, but `(mode_t)va_arg(ap, unsigned int)` matches glibc's own pattern.

**4. `pkg/network/mockibsysfs/render/integration_test.go` (around line 1376)**

Integration test does not assert on per-HCA counter files (`counters/port_xmit_data`, etc.) or on `/sys/class/infiniband_mad/abi_version`. Adding 1-2 grep checks would strengthen end-to-end coverage of those branches.

**5. `pkg/network/mockibsysfs/render/render_test.go` (around line 1749)**

`TestRender_Disabled` would still pass if `Render()` were stubbed to `return nil` for any input. It only validates the disabled branch by chance. Consider asserting that Output is unchanged when `Enabled=false` but other fields are populated (catch a regression where `Enabled` is ignored).

**6. `deployments/nvml-mock/helm/nvml-mock/templates/daemonset.yaml` (around line 65)**

Mounting the same hostPath volume at two paths (`/host/var/lib/nvml-mock` and `/var/lib/nvml-mock`) works on k8s 1.32 (CI confirms) but is unusual. Add a brief comment in the template (already present) and verify against the minimum supported k8s version in the chart's Chart.yaml.

giuliocalzo:
@ArangoGutierrez sorted the comments

ArangoGutierrez (dismissed):
Updated review: rebased onto current main (the original setup.sh conflict with #326 auto-resolved cleanly post-#326 merge); the --verify=false helm plugin install change is already in main via #323 so no longer in this PR diff. The follow-up commit 33366b4a addresses all 6 PE blockers and 6 nits I flagged: gid_attrs as directory, perHCAPortGUID wired into ports/N/port_guid, error wrapping with rel paths, deterministic file ordering via slices, fail-fast on render failures in setup.sh, O_TMPFILE / mode_t fixes in shim.c, TestRender_Disabled strengthened, integration test asserts on counters and abi_version, kubeVersion floor in Chart.yaml. Authorship preserved across rebase. Approving for merge train.

PR #316, opened 2026-04-27
ci: split Helm chart publish into a dedicated workflow

giuliocalzo:
## Summary

- Move the `publish` job out of `.github/workflows/helm.yaml` into a new `.github/workflows/helm-publish.yaml`, mirroring the structure of `nvml-mock-publish.yaml`.
- The new workflow is triggered on pushes to `main` and `v*` tags with path filters on `deployments/nvml-mock/helm/**` and the workflow file itself, replacing the previous in-job `if:` gate.
- Bump `sigstore/cosign-installer` to `v4.1.1` to match the version pinned in `nvml-mock-publish.yaml`.
- `helm.yaml` now only runs `lint` and `unittest` on PRs and pushes to `main`.

## Test plan

- [ ] Open the PR and confirm only `helm / lint` and `helm / unittest` run (no `publish` job).
- [ ] After merge to `main`, confirm `helm Publish / publish` runs and pushes the chart + cosign signature to GHCR.


.github/workflows/helm-publish.yaml
```diff
@@ -0,0 +1,57 @@
+# Copyright 2026 NVIDIA CORPORATION
+# SPDX-License-Identifier: Apache-2.0
+name: helm Publish
+
+permissions: {}
+
+on:
+  push:
```
ArangoGutierrez: Trigger semantics regression: the old job in `helm.yaml` had `needs: [lint, unittest]`, so a chart with broken lint or unit tests could not publish. The new workflow has no cross-workflow gate (GitHub Actions cannot express `needs:` across workflow files directly), and the path filter overlaps with `helm.yaml`'s lint/unittest paths, so on a push to `main` both workflows start in parallel. A chart with broken `helm unittest` will still publish to GHCR. Either (a) call the lint/unittest steps inline at the top of `helm-publish.yaml` before package/push/sign, or (b) use `workflow_run:` triggered by `helm` workflow completion with a `conclusion == 'success'` guard. As written, this weakens the publish gate — block on this.

.github/workflows/helm-publish.yaml
```diff
@@ -0,0 +1,57 @@
...
+    branches:
+      - main
+    tags:
```
ArangoGutierrez: New behavior introduced silently: `tags: ['v*']` triggers publish on tag pushes, which the old job never did (it was `ref == 'refs/heads/main'` only). If this is intentional (release tagging flow), call it out in the PR body and add semver-aware version derivation — `helm show chart` reads `Chart.yaml` `version:`, which is independent of the git tag, so a `v0.2.0` tag push will publish whatever version Chart.yaml currently has. Recommend dropping `tags:` for this PR and doing the release-tag flow as a follow-up with proper version reconciliation.

.github/workflows/helm-publish.yaml
```diff
@@ -0,0 +1,57 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      id-token: write  # needed for keyless cosign signing
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
```
ArangoGutierrez, line 35: `azure/setup-helm@1a275c3b... # v4` — same repo's `helm.yaml` lint job and `nvml-mock-e2e.yaml` use `dda3372f... # v5`. Pin drift inherited from the old block; align to v5 for consistency. Not a blocker.

.github/workflows/helm-publish.yaml
```diff
@@ -0,0 +1,57 @@
...
+    paths:
+      - 'deployments/nvml-mock/helm/**'
+      - '.github/workflows/helm-publish.yaml'
+
+env:
+  CHART_NAME: nvml-mock
+  CHART_DIR: deployments/nvml-mock/helm
+
+jobs:
+  publish:
```
ArangoGutierrez, line 22: No `concurrency:` block. `nvml-mock-publish.yaml` also lacks one, so consistency is preserved, but two simultaneous tag pushes (or main pushes) would race the cosign sign step. Recommend `concurrency: { group: helm-publish-${{ github.ref }}, cancel-in-progress: false }` (do not cancel in progress for publish workflows — let the first finish).

.github/workflows/helm-publish.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+      - name: Log in to GHCR
+        run: echo "${{ secrets.GITHUB_TOKEN }}" | helm registry login ghcr.io -u ${{ github.actor }} --password-stdin
+
+      - name: Package chart
+        run: helm package ${{ env.CHART_DIR }}/${{ env.CHART_NAME }} --destination .cr-release-packages
+
+      - name: Push chart to OCI registry
+        run: helm push .cr-release-packages/${{ env.CHART_NAME }}-${{ steps.chart.outputs.version }}.tgz oci://ghcr.io/nvidia/k8s-test-infra/chart
+
```
ArangoGutierrez, line 53: Push path is `oci://ghcr.io/nvidia/k8s-test-infra/chart` and sign target is `ghcr.io/nvidia/k8s-test-infra/chart/${CHART_NAME}:...`. PR #319 (open, modifying the same job pre-deletion) argues this should be `oci://ghcr.io/nvidia/k8s-test-infra` because `helm push` appends the chart name automatically. Verify against what is actually live on GHCR; if #319 is correct, this PR perpetuates the bug.

ArangoGutierrez (changes requested):
Mechanical move is clean and `actionlint` passes, but the trigger semantics regress in two ways: (1) the cross-job gate `needs: [lint, unittest]` is lost when the publish job leaves `helm.yaml`, so a chart with failing tests can still publish; (2) `tags: ['v*']` is added without a corresponding tag-vs-Chart.yaml version reconciliation, so tag-triggered publishes will use whatever `version:` is in Chart.yaml regardless of the git tag. Address those two before merge. Also coordinate with PR #319 (registry path / jq parsing / keyless flag hardening on the same job) — whichever lands second must carry the other's improvements forward into helm-publish.yaml.

ArangoGutierrez (approved):
Updated review: amended this PR to address the original blockers. Trigger now uses workflow_run on the helm workflows completion with success guard, preserving the lint+unittest cross-workflow gate. Dropped tags:[v*] trigger and added non-cancelling concurrency block. Also propagated the post-train fixes to keep helm-publish.yaml in sync with helm.yaml: jq parsing (#319), setup-helm@v5, cosign-installer@v4.1.2 (#330). Authorship preserved (Giulio Calzolari). Approving for merge train.

PR #311, opened 2026-04-16
docs: update helm install commands to use OCI registry

giuliocalzo:
## Summary

- Replace local chart path (`deployments/nvml-mock/helm/nvml-mock`) with `oci://ghcr.io/nvidia/k8s-test-infra/chart/nvml-mock` in all documentation
- Now that the chart is published as an OCI artifact (#302), docs should point users to the registry install method
- CI/test scripts intentionally left unchanged as they need the local chart path for testing

### Files updated
- `README.md`
- `CONTRIBUTING.md`
- `docs/README.md`
- `docs/quickstart.md`
- `docs/examples.md`
- `docs/integrations/fake-gpu-operator.md`
- `docs/demo/with-fgo/README.md`
- `deployments/nvml-mock/helm/nvml-mock/README.md`
- `deployments/nvml-mock/helm/nvml-mock/Chart.yaml` (version bump to 0.1.1)

## Test plan

- [ ] Verify all markdown links and code blocks render correctly
- [ ] Verify `helm install nvml-mock oci://ghcr.io/nvidia/k8s-test-infra/chart/nvml-mock` works after chart publish

PR #319, opened 2026-05-03
ci(helm): harden chart version parsing in publish workflow

### LINKED ISSUE #310: ci(helm): harden chart version parsing in publish workflow
Follow-up from #302 review.

The Read chart version step uses grep + awk to extract the version from Chart.yaml. This breaks on quoted versions like `version: "1.2.3"`. While unlikely in our chart today, using `helm show chart ... -o json | jq -r .version` would be more robust (jq is pre-installed on GitHub runners).

Related: helm-unittest plugin install only supports tag-based pinning, not SHA. A comment with the SHA was added for traceability — no action needed unless the plugin adds SHA support.

theonlychant:
## What This PR Does
Hardens chart version parsing in the Helm publish workflow to handle 
edge cases that could cause the CI pipeline to fail or produce incorrect 
version strings.

## Why
Closes #310

## Checklist

### DIFF
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -75,7 +75,7 @@ jobs:
       - name: Read chart version
         id: chart
         run: |
-          version=$(helm show chart ${{ env.CHART_DIR }}/${{ env.CHART_NAME }} | grep '^version:' | awk '{print $2}')
+          version=$(helm show chart ${{ env.CHART_DIR }}/${{ env.CHART_NAME }} -o json | jq -er .version)
           echo "version=${version}" >> "$GITHUB_OUTPUT"
 
       - name: Log in to GHCR


.github/workflows/helm.yaml
```diff
@@ -83,12 +83,12 @@ jobs:
...
       - name: Push chart to OCI registry
-        run: helm push .cr-release-packages/${{ env.CHART_NAME }}-${{ steps.chart.outputs.version }}.tgz oci://ghcr.io/nvidia/k8s-test-infra/chart
+        run: helm push .cr-release-packages/${{ env.CHART_NAME }}-${{ steps.chart.outputs.version }}.tgz oci://ghcr.io/nvidia/k8s-test-infra
 
       - name: Install cosign
         uses: sigstore/cosign-installer@c56c2d3e59e4281cc41dea2217323ba5694b171e # v3.8.0
 
-      - name: Sign OCI artifact
+      - name: Sign OCI artifact (keyless)
         run: |
```
ArangoGutierrez, line 94: `cosign sign --keyless` is not a valid flag — running `cosign sign --help` shows no `--keyless` option in any cosign version. Keyless signing is the default when OIDC ambient credentials are present (this workflow has `id-token: write`); no flag is needed. As written this fails with `unknown flag: --keyless`, and removing `--yes` would also cause cosign to hang on the interactive confirmation in CI. This change is also out of scope for issue #310. Please drop it and restore `--yes`.

.github/workflows/helm.yaml
```diff
@@ -83,12 +83,12 @@ jobs:
         run: helm package ${{ env.CHART_DIR }}/${{ env.CHART_NAME }} --destination .cr-release-packages
 
       - name: Push chart to OCI registry
-        run: helm push .cr-release-packages/${{ env.CHART_NAME }}-${{ steps.chart.outputs.version }}.tgz oci://ghcr.io/nvidia/k8s-test-infra/chart
+        run: helm push .cr-release-packages/${{ env.CHART_NAME }}-${{ steps.chart.outputs.version }}.tgz oci://ghcr.io/nvidia/k8s-test-infra
```
ArangoGutierrez: Removing `/chart` from the push path silently relocates the artifact from `oci://ghcr.io/nvidia/k8s-test-infra/chart/nvml-mock` to `oci://ghcr.io/nvidia/k8s-test-infra/nvml-mock`. This breaks 20+ user-facing install commands across `deployments/nvml-mock/helm/nvml-mock/README.md`, `docs/quickstart.md`, `docs/examples.md`, `docs/README.md`, and `docs/demo/with-fgo/README.md`. The path was established by #302 and documented by #311. This change is out of scope for issue #310 — please revert.

.github/workflows/helm.yaml
```diff
@@ -83,12 +83,12 @@ jobs:
...
       - name: Install cosign
         uses: sigstore/cosign-installer@c56c2d3e59e4281cc41dea2217323ba5694b171e # v3.8.0
 
-      - name: Sign OCI artifact
+      - name: Sign OCI artifact (keyless)
         run: |
-          cosign sign --yes \
-            ghcr.io/nvidia/k8s-test-infra/chart/${{ env.CHART_NAME }}:${{ steps.chart.outputs.version }}
+          cosign sign --keyless \
+            ghcr.io/nvidia/k8s-test-infra/${{ env.CHART_NAME }}:${{ steps.chart.outputs.version }}
```
ArangoGutierrez: Sign target path must match the push path. If the push path stays as `.../chart/nvml-mock` (which it must — see comment on line 86), the sign argument must also stay `ghcr.io/nvidia/k8s-test-infra/chart/${{ env.CHART_NAME }}:...`. Please revert.

.github/workflows/helm.yaml
```diff
@@ -73,7 +73,7 @@ jobs:
       - name: Read chart version
         id: chart
         run: |
-          version=$(helm show chart ${{ env.CHART_DIR }}/${{ env.CHART_NAME }} | grep '^version:' | awk '{print $2}')
+          version=$(helm show chart ${{ env.CHART_DIR }}/${{ env.CHART_NAME }} -o json | jq -r .version)
```
ArangoGutierrez: Optional: use `jq -er .version` (the `-e` flag) so the step fails loudly if `version` is absent instead of emitting the literal string `null` into `$GITHUB_OUTPUT`. Not required by issue #310.

ArangoGutierrez (changes requested):
Thanks for picking up #310. The version-parsing change on line 76 is correct and matches the issue exactly. However, this PR also includes three undocumented changes that are out of scope and break things: (1) `cosign sign --keyless` is not a real flag — keyless signing is automatic when OIDC ambient creds are present, and removing `--yes` would cause cosign to hang on a confirmation prompt in CI; (2) removing `/chart` from the OCI push path silently moves the artifact away from the documented location and breaks 20+ install commands across our README and docs; (3) the matching change to the sign-target path is wrong for the same reason. Please reduce the diff to the single `helm show chart ... -o json | jq -r .version` change. Heads-up: BLOCKED state is from `required_signatures` on main — your head commit `1798319c` reports unsigned. Please re-sign with GPG (`git commit -S`) or re-commit via the GitHub web editor, which will satisfy the signature gate.

.github/workflows/helm.yaml
theonlychant: ok now it uses cosign sign --yes and does not include --keyless. I think it is all good now

ArangoGutierrez (approved):
Updated review for the new state of this PR — the diff is now clean (single line: helm show chart -o json | jq -er .version), the bogus --keyless / /chart changes are gone, and the commits have been squashed and re-signed (verified: true). Approving for merge train. Thanks for picking this up @theonlychant — addressed the original review feedback exactly.

Issue #245, opened 2026-03-13
feat(nvml-mock): GPU failure injection (device lost, Xid errors)


ArangoGutierrez:
## Context

nvml-mock supports `simulate_gpu_lost` and `failure_probability` per device — useful for testing how consumers handle GPU failures, Xid errors, and degraded hardware.

Our mock always reports healthy GPUs.

## Scope

Add failure injection configuration to YAML profiles:
- `failure.mode`: `healthy` (default), `lost`, `fallen_off_bus`, `ecc_uncorrectable`
- `failure.probability`: float 0.0-1.0 for stochastic failures
- `failure.after_calls`: trigger failure after N NVML calls (deterministic testing)

When triggered:
- `GetDeviceHandleByIndex` returns `ERROR_GPU_IS_LOST`
- Health check functions report appropriate error codes
- Xid error injection via `GetViolationStatus`

## Acceptance Criteria

- [ ] Device can be configured to report as "lost" via YAML
- [ ] Stochastic failure mode works with configurable probability
- [ ] Deterministic failure-after-N-calls mode for reproducible tests
- [ ] Engine unit tests cover all failure modes
- [ ] GPU Operator/device-plugin handles failure gracefully (doesn't crash)

## Priority

Low — useful for resilience testing but not blocking any current workflow.

giuliocalzo:
@ArangoGutierrez I supposed can be closed as https://github.com/NVIDIA/k8s-test-infra/pull/328 is merged

ArangoGutierrez:
Done — delivered by #328 (merged 2026-05-05). All acceptance criteria met: lost / fallen_off_bus / ecc_uncorrectable modes, probability + after_calls triggers, Xid critical events through the standard NVML event set, engine unit tests, Helm `gpu.failureInjection` block with schema validation, and an end-to-end demo at `docs/demo/failure-injection/run.sh`. Thanks @giuliocalzo!

PR #328, opened 2026-05-05
feat(nvml-mock): GPU failure injection (lost, fallen-off-bus, ECC, Xid)

giuliocalzo:
## Summary

Implements #245: opt-in GPU failure injection for `nvml-mock`. Consumers (device plugin, GPU operator, monitoring stacks) can now exercise lost / fallen-off-bus / uncorrectable-ECC GPUs and Xid critical errors without real broken hardware. When `failure:` is absent the engine reports healthy hardware exactly as before — zero behaviour change for existing users.

This PR has been rebased onto `main` now that #323 (dynamic metrics) has merged; the diff is the failure-injection feature only.

### Engine

- `FailureInjectionConfig` (mode, probability, after_calls, seed, xid) and `XidErrorConfig` types in `engine/config_types.go`, plus a `Failure` field on `DeviceConfig`.
- `engine/failure_injector.go` implements a sticky `failureInjector` with seeded `math/rand/v2`, atomic call counter, and three composable triggers:
  - **mode-only** → trip on the first guarded call
  - **`after_calls: N`** → deterministic trip on the Nth guarded call
  - **`probability: p`** → stochastic per-call roll
- `ConfigurableDevice` calls `tickFailure` at the entry of every guarded metric / counter getter (temperature, power, utilization, memory, clock, ECC counters, remapped rows, violation status). `lost` / `fallen_off_bus` modes propagate `ERROR_GPU_IS_LOST`; `ecc_uncorrectable` keeps the device addressable and surfaces a strictly-monotonic uncorrectable counter.
- Identity getters (`GetUUID`, `GetName`, `GetIndex`, `GetMinorNumber`, `GetSerial`, `GetBoardPartNumber`, `GetPciInfo`, `GetBrand`) are guarded too — once a device is `lost` / `fallen_off_bus` they return `ERROR_GPU_IS_LOST` instead of silently succeeding, matching real NVML behaviour where the kernel driver has marked the GPU gone. Identity getters use a tick-free guard so they don't advance the `after_calls` counter.
- Server-level `DeviceGetHandleByIndex` / `ByUUID` / `ByPciBusId` consult the device tripped state and return `ERROR_GPU_IS_LOST` for lost devices, mirroring real NVML.
- **Xid is delivered through the standard NVML event set**, not via overloading `ViolationTime.ViolationTime`. The bridge advertises `NVML_EVENT_TYPE_XID_CRITICAL_ERROR` from `nvmlDeviceGetSupportedEventTypes` and `nvmlEventSetWait_v2` is wired through `Engine.PendingXidEvent`; each tripped Xid is delivered exactly once per engine lifetime (atomic CompareAndSwap on a per-device flag), matching real NVML semantics. `GetViolationStatus` is now spec-compliant — it returns an empty `ViolationTime{}` for healthy / ECC-tripped devices and `ERROR_GPU_IS_LOST` for lost / fallen-off-bus.
- `mergeDeviceOverride` propagates `Failure` so per-device overrides compose with `device_defaults`.

### Helm chart

- New `gpu.failureInjection` values block (disabled by default) with mode, probability, after_calls, seed, and xid.code.
- **Default `mode: healthy`** is intentionally inert: even with `enabled: true`, engaging failure injection requires explicitly selecting one of `lost`, `fallen_off_bus`, `ecc_uncorrectable`. This eliminates the `--set gpu.failureInjection.enabled=true` footgun where the previous defaults would have tripped every GPU on the very first NVML call.
- `_helpers.tpl` overlays the `failure:` block under `device_defaults:` only when `gpu.failureInjection.enabled=true`. Composes with the existing `dynamicMetrics` overlay; both can be enabled simultaneously.
- **`values.schema.json`** validates the `failureInjection` block at install / upgrade time. Typos like `mode: healhty`, out-of-range `probability: 1.5`, or negative `after_calls` are rejected by Helm before the chart renders, surfacing actionable schema errors instead of silent runtime coercion. The `profile` field is deliberately left un-enumerated so the existing template-level validation (which points users at `gpu.customConfig`) keeps producing its friendly error.
- DaemonSet operability:
  - `checksum/config` annotation hashes the rendered GPU config so any change to `gpu.dynamicMetrics` / `gpu.failureInjection` automatically rolls every pod.
  - `updateStrategy.rollingUpdate.maxUnavailable: 25%` (was the upstream default of 1) for faster but still bounded rollouts; demos override to 100% when they want every node to flip at once.
  - `terminationGracePeriodSeconds: 10` (preStop is a few `rm -rf` + `kubectl label` calls), so rollouts don't wait the upstream 30s default per pod.

### Per-mode behaviour

| mode                | guarded API calls return | handle lookup returns | identity getters    | ECC counters         | event set                       |
| ------------------- | ------------------------ | --------------------- | ------------------- | -------------------- | ------------------------------- |
| `healthy` (default) | normal values            | normal handle         | normal values       | zero                 | empty                           |
| `lost`              | `ERROR_GPU_IS_LOST`      | `ERROR_GPU_IS_LOST`   | `ERROR_GPU_IS_LOST` | error                | empty                           |
| `fallen_off_bus`    | `ERROR_GPU_IS_LOST`      | `ERROR_GPU_IS_LOST`   | `ERROR_GPU_IS_LOST` | error                | empty                           |
| `ecc_uncorrectable` | normal values            | normal handle         | normal values       | strictly-increasing  | one `XID_CRITICAL_ERROR` if xid |

### Unit tests

- `go test -race ./...` — all packages green. New engine cases cover each mode, probability bounds, `after_calls` determinism, sticky-trip semantics, seed reproducibility, concurrent-`Tick` safety, override merge, identity-getter pre/post-trip behaviour, `PendingXidEvent` deliver-once + Xid-less-config no-event semantics, and `GetViolationStatus` staying spec-compliant after a Xid trip.
- `helm unittest deployments/nvml-mock/helm/nvml-mock` — **94 / 94 cases, 14 / 14 snapshots green**. New cases verify disabled default, enabled injection, per-field overrides, no `enabled` flag leakage into the rendered YAML, combined `dynamicMetrics`+`failureInjection` rendering, schema rejection of bad `mode` / `probability` / `after_calls` inputs, and the new DaemonSet annotations / updateStrategy / terminationGracePeriodSeconds wiring.

### End-to-end demo

A dedicated demo at `docs/demo/failure-injection/run.sh` spins up a Kind cluster with the chart's H100 profile (8 GPUs per node) and walks the deployment through every mode via successive `helm upgrade --reuse-values`, asserting the `nvidia-smi` surface for each. Output of the latest run:

```
==> Scenario 1: healthy baseline (failureInjection.enabled=false)
    DaemonSet pod ready: nvml-mock-c95rw
    ✓ ConfigMap has no failure: block (as expected)
    ✓ nvidia-smi -L lists 8 GPU(s) (healthy baseline)
    ✓ Healthy ECC baseline: 0

==> Scenario 2: ecc_uncorrectable + Xid 79 (after_calls=1)
    helm upgrade -> ecc_uncorrectable
    Pod after rollout: nvml-mock-7wxgs
    ✓ ConfigMap contains: mode: ecc_uncorrectable
    ✓ nvidia-smi -L still lists 8 GPUs (device addressable)
    ECC uncorrectable per-GPU readings:
      1
      1
      1
      1
      1
      1
      1
      1
    ✓ ECC uncorrectable max = 1 (>0 confirms trip)

==> Scenario 3: lost (after_calls=1)
    helm upgrade -> lost
    Pod after rollout: nvml-mock-fk9b2
    ✓ ConfigMap contains: mode: lost
    nvidia-smi temperature query output:
      [GPU is lost]
      [GPU is lost]
      [GPU is lost]
      [GPU is lost]
      [GPU is lost]
      [GPU is lost]
      [GPU is lost]
      [GPU is lost]
    ✓ lost mode propagates an error marker through nvidia-smi

==> Scenario 4: fallen_off_bus + Xid 79 (after_calls=1)
    helm upgrade -> fallen_off_bus
    Pod after rollout: nvml-mock-8pk57
    ✓ ConfigMap contains: mode: fallen_off_bus
    ✓ ConfigMap contains: code: 79
    nvidia-smi temperature query output:
      [GPU is lost]
      [GPU is lost]
      [GPU is lost]
      [GPU is lost]
      [GPU is lost]
      [GPU is lost]
      [GPU is lost]
      [GPU is lost]
    ✓ fallen_off_bus propagates an error marker through nvidia-smi

==> All four failure-injection scenarios verified.

   Scenario 1  healthy            : nvidia-smi -L lists 8 GPU(s); ECC = 0
   Scenario 2  ecc_uncorrectable  : device addressable; ECC uncorrectable > 0
   Scenario 3  lost               : nvidia-smi metric query returns error markers
   Scenario 4  fallen_off_bus     : nvidia-smi metric query returns error markers; Xid 79 queued
```

The Xid critical-error event itself is delivered through `NVML_EVENT_TYPE_XID_CRITICAL_ERROR`, which `nvidia-smi` does not subscribe to. Real consumers (NVIDIA device plugin health monitor, dcgm-exporter) read it via `nvmlEventSetWait_v2` and will surface `Xid 79` / mark the GPU `Unhealthy` against this same cluster.

## Notes for reviewers

- Addresses every "out-of-diff observation" raised in the previous review:
  1. Xid via `NVML_EVENT_TYPE_XID_CRITICAL_ERROR` (no more `ViolationTime` overload).
  2. Identity getters guarded.
  3. `values.schema.json` validation.
  4. Default `mode: healthy` removes the immediate-trip footgun.
  5. Immutability comment on `failureInjector.cfg`.
- The injector counter is **per-process**, not per-pod-lifetime. A fresh `nvidia-smi` invocation starts a brand-new `failureInjector`. Both READMEs document this caveat and recommend either `after_calls: 1` for narrow queries or richer queries (`nvidia-smi -q`) that issue several guarded calls per GPU.
- Once tripped a device stays tripped for the lifetime of the engine, matching real lost / fallen-off-bus GPUs that don't recover without a reboot.
- `ecc_uncorrectable` reports the running call count as the uncorrectable count, so consumers correlating polls see a strictly-increasing value (not a synthetic constant).

### DIFF
--- a/deployments/nvml-mock/helm/nvml-mock/README.md
+++ b/deployments/nvml-mock/helm/nvml-mock/README.md
@@ -496,11 +496,11 @@ helm install nvml-mock oci://ghcr.io/nvidia/k8s-test-infra/chart/nvml-mock \
 | `gpu.dynamicMetrics.power.*` | see `values.yaml` | `base_mw`, `variance_mw` for the power-draw generator (clamped to `power.min_limit_mw`/`max_limit_mw` from the profile). |
 | `gpu.dynamicMetrics.utilization.*` | see `values.yaml` | `pattern` (`idle` \| `busy` \| `burst` \| `steady`), `gpu_min/max`, `memory_min/max`, `burst_period_sec`. |
 | `gpu.failureInjection.enabled` | `false` | Enable simulated GPU failures (lost / fallen off bus / uncorrectable ECC). See [Failure Injection](#failure-injection) below. |
-| `gpu.failureInjection.mode` | `lost` | Failure mode: `healthy`, `lost`, `fallen_off_bus`, or `ecc_uncorrectable`. |
-| `gpu.failureInjection.probability` | `0.0` | Per-call probability (0..1) for stochastic failure activation. |
+| `gpu.failureInjection.mode` | `healthy` | Failure mode: `healthy` (default, no-op), `lost`, `fallen_off_bus`, or `ecc_uncorrectable`. With the inert default, `enabled: true` alone produces a healthy device — you must set `mode` explicitly to engage failures. |
+| `gpu.failureInjection.probability` | `0.0` | Per-call probability `[0, 1]` for stochastic failure activation. |
 | `gpu.failureInjection.after_calls` | `0` | Activate failure deterministically after N guarded NVML calls (0 = disabled). |
 | `gpu.failureInjection.seed` | `0` | RNG seed for probability rolls; `0` uses a time-based seed. |
-| `gpu.failureInjection.xid.code` | `0` | Xid error code surfaced via `GetViolationStatus` once tripped. `0` = no Xid. |
+| `gpu.failureInjection.xid.code` | `0` | Xid error code delivered via the NVML event set (`NVML_EVENT_TYPE_XID_CRITICAL_ERROR`) once tripped. `0` = no Xid. |
 | `image.repository` | `ghcr.io/nvidia/nvml-mock` | Container image repository |
 | `image.tag` | `latest` | Container image tag |
 | `image.pullPolicy` | `IfNotPresent` | Image pull policy |
@@ -732,12 +732,19 @@ gpu:
 
 Per-mode behaviour:
 
-| mode                | guarded API calls return | handle lookup returns | ECC counters         | `GetViolationStatus`              |
-| ------------------- | ------------------------ | --------------------- | -------------------- | --------------------------------- |
-| `healthy` (default) | normal values            | normal handle         | zero                 | empty                             |
-| `lost`              | `ERROR_GPU_IS_LOST`      | `ERROR_GPU_IS_LOST`   | error                | error                             |
-| `fallen_off_bus`    | `ERROR_GPU_IS_LOST`      | `ERROR_GPU_IS_LOST`   | error                | error                             |
-| `ecc_uncorrectable` | normal values            | normal handle         | strictly-increasing  | non-zero `ViolationTime` if `xid` |
+| mode                | guarded API calls return | handle lookup returns | identity getters    | ECC counters         | event set                       |
+| ------------------- | ------------------------ | --------------------- | ------------------- | -------------------- | ------------------------------- |
+| `healthy` (default) | normal values            | normal handle         | normal values       | zero                 | empty                           |
+| `lost`              | `ERROR_GPU_IS_LOST`      | `ERROR_GPU_IS_LOST`   | `ERROR_GPU_IS_LOST` | error                | empty                           |
+| `fallen_off_bus`    | `ERROR_GPU_IS_LOST`      | `ERROR_GPU_IS_LOST`   | `ERROR_GPU_IS_LOST` | error                | empty                           |
+| `ecc_uncorrectable` | normal values            | normal handle         | normal values       | strictly-increasing  | one `XID_CRITICAL_ERROR` if xid |
+
+Values rendered into the ConfigMap are validated against
+[`values.schema.json`](./values.schema.json) at install / upgrade time:
+typos like `mode: healhty` or out-of-range values like `probability: 1.5`
+are rejected by Helm before the chart renders, so misconfigurations
+surface as actionable schema errors instead of silent runtime
+coercion.
 
 Failure injection composes with `gpu.dynamicMetrics`: with both enabled the
 device returns dynamic readings while healthy and switches to the configured
@@ -763,7 +770,7 @@ kubectl exec ds/nvml-mock -- nvidia-smi --query-gpu=name,uuid --format=csv
 kubectl exec ds/nvml-mock -- nvidia-smi -q                # "GPU is lost"
 
 # mode: ecc_uncorrectable  ─  device stays addressable; counters grow and
-# GetViolationStatus surfaces the configured Xid code.
+# nvmlEventSetWait_v2 delivers the configured Xid once per trip.
 kubectl exec ds/nvml-mock -- nvidia-smi -q -d ECC
 kubectl exec ds/nvml-mock -- nvidia-smi \
   --query-gpu=ecc.errors.uncorrected.aggregate.total --format=csv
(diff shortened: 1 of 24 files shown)


ArangoGutierrez (dismissed):
@giuliocalzo solid work. Tests are real (specific NVML return codes, real engine wiring via `newTestDeviceWithConfig`, no theater), concurrency story is sane (atomic counter + tripped flag, mutex only around the RNG roll), and the nil-injector pattern keeps the hot path branch-free for the default healthy case. Sticky semantics match real lost/fallen-off-bus GPUs. Helm overlay tests verify the `enabled` toggle never leaks into the rendered YAML. AfterCalls determinism, probability=0/1 boundaries, seed reproducibility, divergence between seeds, post-trip stickiness, override merge, and ECC strict monotonicity are each independently asserted. CI fully green across all 9 nvml-mock-e2e variants. Approving.

**Stack note:** this PR is stacked on #323 (dynamic-metrics). Both touch `device.go` and `config_types.go`. Train will merge #323 first; this one rebases cleanly afterward.

**Out-of-diff observations (follow-ups, not blockers):**
- `pkg/gpu/mocknvml/engine/device.go:1556` — `GetViolationStatus` overloads `ViolationTime.ViolationTime` (officially throttle ns) to carry the Xid code. Documented hack, but consumers reading the field per NVML spec (dcgm-exporter, monitoring stacks) will misinterpret. Consider follow-up to surface Xid via `nvmlEventSetWait_v2` / `NVML_EVENT_TYPE_XID_CRITICAL_ERROR`.
- `pkg/gpu/mocknvml/engine/device.go:197` — identity getters (`GetUUID`, `GetName`, `GetSerial`, `GetIndex`, `GetMinorNumber`, `GetBoardPartNumber`, `GetBrand`, `GetPciInfo`) are NOT `tickFailure`-guarded — handles resolved pre-trip return identity fields with SUCCESS post-trip. Real lost GPUs return `GPU_IS_LOST` from these too. Either document explicitly or extend the guard set.
- `deployments/nvml-mock/helm/nvml-mock/values.yaml:900` — no schema validation: `mode` accepts arbitrary strings (silently coerced to healthy), `probability` not constrained to [0,1]. Defense-in-depth gap; consider `values.schema.json`.
- Defaults `mode: lost` + `after_calls: 0` + `probability: 0.0` + `enabled: true` mean immediate first-call failure. Easy footgun; consider defaulting `after_calls: 1` or `mode: healthy`.
- `pkg/gpu/mocknvml/engine/failure_injector.go:2657` — `f.cfg` is read without sync after construction. Safe today (immutable post-newFailureInjector) but unenforced; a `// cfg is immutable after construction` comment would prevent future regressions.

giuliocalzo:
@ArangoGutierrez just improved based on your comments and with a dedicated demo scenario

PR #337, opened 2026-05-11
feat(nvml-mock): ComputeDomain simulation with fabric APIs and fake IMEX

### LINKED ISSUE #304: ComputeDomain simulation: NVML fabric APIs + fake IMEX coordination
Related: kubernetes-sigs/dra-driver-nvidia-gpu#1031

## Summary

Extend nvml-mock to enable hardware-free ComputeDomain testing on KIND
clusters. Mock the NVML fabric APIs and replace nvidia-imex/nvidia-imex-ctl
with lightweight fakes that coordinate through marker files on a shared
volume. The upstream compute-domain-controller and compute-domain-daemon
run unmodified against these mocks.

## Motivation

The DRA ComputeDomain lifecycle requires GB200/GB300 NVLink hardware, the
nvidia-imex daemon, and nvidia-imex-ctl for readiness probes. Most
contributors and CI environments lack access to this hardware, blocking
local validation of controller and daemon changes.

The upstream issue (kubernetes-sigs/dra-driver-nvidia-gpu#1031) proposes
`--dry-run` flags to skip IMEX. This proposal takes a complementary
approach: provide the mock infrastructure so the real code exercises its
full paths, including peer coordination, failure detection, and recovery.

### Goals

- Upstream daemon and controller code runs unmodified on KIND
- Full lifecycle fidelity: startup, peer discovery, readiness, failure, recovery
- Single container image bundles mock NVML + fake IMEX binaries
- Cluster topology declared in a single ConfigMap

### Non-Goals

- Upstream code changes (no dry-run flags, no fake kubelet plugin)
- CDI device injection for IMEX channels (separate concern)
- Performance or scale testing
- Multi-cluster topologies

## Architecture

```
+-----------------------------------------------------------+
|  Real upstream compute-domain-controller + daemon          |
|  (unmodified -- deployed via their own Helm chart)         |
+-----------------------------------------------------------+
|  Fake IMEX binaries (nvidia-imex, nvidia-imex-ctl)         |
|  - In $PATH inside daemon pod (custom image layer)         |
|  - File-per-daemon coordination on shared hostPath         |
|  - nvidia-imex-ctl reads nodes.cfg + checks shared state   |
+-----------------------------------------------------------+
|  Mock NVML with fabric APIs                                |
|  - nvmlDeviceGetGpuFabricInfo / GpuFabricInfoV             |
|  - Returns clusterUuid, cliqueId, state from config        |
|  - Loaded via --nvidia-driver-root (existing mechanism)    |
+-----------------------------------------------------------+
|  Topology ConfigMap (cluster-level, single source of truth)|
|  - Declares domain UUID, node-to-clique assignments        |
|  - Mounted into nvml-mock pods, read by NVML engine        |
+-----------------------------------------------------------+
|  Shared hostPath: /var/lib/nvml-mock/imex-state/           |
|  (cross-node coordination -- KIND nodes share host fs)     |
+-----------------------------------------------------------+
```

The compute-domain-daemon already writes `nodes.cfg` with peer IPs filtered
by clique. The fake nvidia-imex-ctl reads that same file to know its peers,
then checks the shared volume for their marker files. No custom clique-ID
plumbing needed -- the daemon's own code already does the filtering.

### Component 1: Mock NVML Fabric APIs

New config type in `pkg/gpu/mocknvml/engine/config_types.go`:

```go
type FabricConfig struct {
    ClusterUUID string `json:"cluster_uuid"` // 16-byte hex UUID for the NVLink domain
    CliqueID    int    `json:"clique_id"`     // which clique within the domain
    State       string `json:"state"`         // "completed", "not_started", "in_progress"
    PartitionID int    `json:"partition_id"`  // partition within the clique (v2 API)
}
```

Added to `DeviceConfig` as `Fabric *FabricConfig`. All GPUs on the same
node share the same fabric config (NVLink domain is a node-level property).

The GB200 YAML profile gains a `fabric:` section:

```yaml
fabric:
  cluster_uuid: "00000000-0000-0000-0000-000000000001"
  clique_id: 0
  state: "completed"
```

The `clique_id` default is overridden per-node at runtime by the topology
ConfigMap (see Component 3).

**Config loading** -- the engine's `LoadConfig()` gains a topology overlay:

1. Load YAML profile as today (gets default fabric config)
2. Check for topology file at `MOCK_TOPOLOGY_CONFIG` env var
   (default: `/config/topology.yaml`)
3. If present, find entry matching `NODE_NAME` env var
4. Override `clique_id` and `cluster_uuid` from the topology entry
5. Node not in topology -> `FabricConfig` stays nil ->
   `GetGpuFabricInfo` returns `NOT_SUPPORTED` (non-GB200 behavior)

**Device methods** -- two new methods on `ConfigurableDevice`:

- `GetGpuFabricInfo()` -- v1 struct with clusterUuid, cliqueId, state.
  Returns `NOT_SUPPORTED` when fabric config is nil.
- `GetGpuFabricInfoV()` -- v2 struct adds partitionId.

Plus a CGo bridge export with version aliasing, following the existing
pattern (14 aliases already exist for nvidia-smi compatibility).

### Component 2: Fake IMEX Binaries

Two Go programs under `cmd/fake-imex/`, compiled in the nvml-mock
Dockerfile and installed at `/usr/bin/nvidia-imex` and
`/usr/bin/nvidia-imex-ctl`.

**nvidia-imex** (fake daemon, ~100 lines)

The real compute-domain-daemon spawns this as a subprocess and monitors it.

Inputs:
- `POD_IP` env var (Kubernetes downward API)
- `IMEX_NODES_CONFIG` env var (default: `/imexd/nodes.cfg`) -- peer IP
  list, one per line, written by the real daemon's `writeDaemonsConfig()`
- `IMEX_STATE_DIR` env var (default: `/var/lib/nvml-mock/imex-state/`)

Behavior:
1. Accept any CLI flags (`-c /imexd/imexd.cfg` -- accepted, config
   content ignored)
2. Write marker file: `<state-dir>/<pod-ip>` (empty file, presence = alive)
3. Poll loop (2s): re-read `nodes.cfg`, log peer count
4. `SIGUSR1`: immediate re-read of `nodes.cfg` (DNS-based update loop)
5. `SIGTERM`/`SIGINT`: remove marker file, exit 0

**nvidia-imex-ctl** (fake readiness probe, ~80 lines)

The upstream `check()` calls `nvidia-imex-ctl -c /imexd/imexd.cfg -q`
and expects `"READY\n"`.

Behavior:
1. Accept `-c <config> -q` flags (config content ignored)
2. Read `IMEX_NODES_CONFIG` -- list of peer IPs
3. For each peer IP: check if `<state-dir>/<ip>` marker exists
4. All present: print `"READY\n"`, exit 0
5. Any missing: exit 1

**Lifecycle scenarios:**

| Scenario | Behavior |
|---|---|
| Happy path | All pods start, each writes marker, ctl sees all peers, READY, controller marks domain ready |
| Pod failure | SIGTERM removes marker, surviving peers' ctl returns exit 1, controller detects not-ready |
| Peer addition (IP loop) | Daemon restarts nvidia-imex with updated nodes.cfg, new fake writes marker |
| Peer addition (DNS loop) | Daemon sends SIGUSR1, fake re-reads nodes.cfg |

### Component 3: Topology ConfigMap

Single source of truth for domain layout:

```yaml
version: 1
domains:
  - name: "test-domain-1"
    uuid: "00000000-0000-0000-0000-000000000001"
    cliques:
      - id: 0
        nodes: [kind-worker, kind-worker2]
      - id: 1
        nodes: [kind-worker3, kind-worker4]
```

Each node appears in exactly one clique within one domain.

Consumed only by the mock NVML engine (at `LoadConfig()` time). The fake
IMEX binaries do NOT read it -- they get peer info from `nodes.cfg` and
coordination state from the shared volume.

Helm values:

```yaml
topology:
  enabled: false
  domains: []

imex:
  enabled: false
  stateDir: /var/lib/nvml-mock/imex-state
```

### Component 4: Shared Volume Coordination

```
/var/lib/nvml-mock/imex-state/
  10.244.0.5    <- marker (empty file)
  10.244.1.3
  10.244.2.7
```

Markers are keyed by pod IP (globally unique in k8s). No clique
subdirectories -- `nodes.cfg` already scopes to the correct clique.

In KIND, all worker nodes mount the same host directory via extraMounts:

```yaml
extraMounts:
  - hostPath: /tmp/nvml-mock-imex-state
    containerPath: /var/lib/nvml-mock/imex-state
```

Protocol:
1. Fake nvidia-imex starts -> writes marker
2. nvidia-imex-ctl reads nodes.cfg, checks markers for all peers
3. All present -> `"READY\n"`, exit 0
4. Pod dies -> SIGTERM -> marker removed -> peers detect absence

### Component 5: Deployment Integration

The daemon pods need fake binaries in `$PATH` and access to the shared
state directory. Solution: a thin image layered on the upstream daemon:

```dockerfile
FROM registry.k8s.io/dra-driver-nvidia-gpu/compute-domain-daemon:latest
COPY --from=nvml-mock:latest /usr/bin/nvidia-imex /usr/bin/nvidia-imex
COPY --from=nvml-mock:latest /usr/bin/nvidia-imex-ctl /usr/bin/nvidia-imex-ctl
```

E2E test values override the daemon image reference. No upstream changes.

## Alternatives Considered

**Dry-run flags (upstream issue proposal)**: Adding `--dry-run` to the
controller and daemon. Simpler but skips IMEX code paths entirely --
peer coordination, failure handling, and recovery logic are never
exercised. Complementary to this proposal rather than a replacement.

**Network-based coordination**: Replace file markers with TCP health
checks between fake IMEX daemons. More realistic (real IMEX uses network)
but adds complexity (port management, firewall considerations). File-based
approach is simpler and sufficient for KIND.

**Topology-driven deterministic readiness**: Have the fake always return
READY based on ConfigMap membership, with no actual peer coordination.
Simplest, but does not test failure/recovery scenarios.

## Testing Plan

**Unit tests**: GetGpuFabricInfo returns correct values from config,
NOT_SUPPORTED when nil, topology overlay works per node name. IMEX ctl
logic: all peers present -> READY, missing peer -> exit 1, empty
nodes.cfg -> READY.

**Integration test**: Start 3 fake nvidia-imex processes with a shared
temp dir, verify ctl returns READY, kill one, verify exit 1, restart,
verify READY again.

**E2E in KIND**: Deploy nvml-mock with topology enabled (2 nodes, 1
domain, 1 clique), verify fabric info via test binary, verify IMEX
coordination from within pods. Upstream controller/daemon deployment is
a stretch goal.

## Deliverables

| File | What |
|---|---|
| `pkg/gpu/mocknvml/engine/config_types.go` | FabricConfig type |
| `pkg/gpu/mocknvml/engine/config.go` | Topology overlay loading |
| `pkg/gpu/mocknvml/engine/device.go` | GetGpuFabricInfo, GetGpuFabricInfoV |
| `pkg/gpu/mocknvml/bridge/` | CGo export for fabric API |
| `cmd/fake-imex/daemon/main.go` | Fake nvidia-imex |
| `cmd/fake-imex/ctl/main.go` | Fake nvidia-imex-ctl |
| `deployments/nvml-mock/Dockerfile` | Build fake IMEX binaries |
| `deployments/nvml-mock/Dockerfile.compute-domain-daemon` | Thin daemon layer |
| `deployments/nvml-mock/helm/nvml-mock/templates/topology-configmap.yaml` | Helm template |
| `deployments/nvml-mock/helm/nvml-mock/values.yaml` | topology + imex values |
| `deployments/nvml-mock/helm/nvml-mock/profiles/gb200.yaml` | fabric section |
| `tests/e2e/kind-compute-domain-config.yaml` | KIND config with extraMounts |

giuliocalzo:
## Summary

Closes #304. Adds end-to-end simulation support for the NVIDIA GPU
Operator's ComputeDomain lifecycle on KIND clusters by mocking the
NVML fabric APIs and providing a file-based replacement for
\`nvidia-imex\` / \`nvidia-imex-ctl\`.

- **Mock NVML fabric APIs**: new \`FabricConfig\` type, per-device
  \`GetMockFabricInfo\` / \`GetMockFabricInfoV\` methods and a
  cluster-level topology overlay applied per \`NODE_NAME\` at
  \`LoadConfig\` time. CGo bridge exposes
  \`nvmlDeviceGetGpuFabricInfo[V]\` with ABI-compatible C structs.
- **Fake IMEX**: \`cmd/fake-imex/{daemon,ctl}\` plus a shared
  \`pkg/imexcoord\` package coordinate peers through marker files on
  a shared hostPath volume — suitable as a drop-in for
  \`nvidia-imex\` / \`nvidia-imex-ctl\` in tests.
- **Helm**: new topology ConfigMap template, \`imex\` / \`topology\`
  values, default \`fabric\` block in the \`gb200\` profile and
  daemonset wiring for the new volumes and env vars.
- **Demo**: \`docs/demo/compute-domain/\` with \`run.sh\`, README,
  \`topology.yaml\` and a dedicated KIND config exercising 4 workers
  across 2 cliques (verifies in-clique peer readiness, cross-clique
  isolation and a topology rebind).
- **Tests**: unit tests for fabric overlay + \`imexcoord\`, plus a
  fake-imex integration test.

## Test plan

- [x] \`go test ./pkg/gpu/mocknvml/engine/... ./pkg/imexcoord/... ./cmd/fake-imex/...\`
- [x] \`docs/demo/compute-domain/run.sh\` on a local KIND cluster
  (4 workers, gb200 profile, 2 cliques): per-clique readiness +
  cross-clique \`fake-imex-ctl\` failures + topology rebind all pass
- [ ] Reviewer: rebuild the chart image and re-run the demo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@
 *.out
 *.test
 
+# Stray binaries from `go build ./cmd/...` at repo root
+/check-fabric
+
 # Agentic working artifacts
 .agents/
 AGENTS.md
--- a/cmd/check-fabric/main.go
+++ b/cmd/check-fabric/main.go
@@ -1,0 +1,108 @@
+// Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// check-fabric prints the NVLink fabric attributes of every visible GPU
+// by calling nvmlDeviceGetGpuFabricInfo / nvmlDeviceGetGpuFabricInfoV
+// through go-nvml. It is intended as the demo-time consumer for the
+// ComputeDomain simulation work (NVIDIA/k8s-test-infra#304): with the
+// mock NVML library on LD_LIBRARY_PATH (or via the standard
+// /usr/local/lib install), the binary reports the cluster UUID, clique
+// ID, and registration state that the topology overlay assigned to
+// this Kubernetes node.
+//
+// Exit codes:
+//
+//	0  - every visible GPU reports fabric info
+//	2  - at least one GPU returned ERROR_NOT_SUPPORTED (no fabric)
+//	1  - any other NVML failure
+package main
+
+import (
+	"encoding/hex"
+	"fmt"
+	"os"
+
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+)
+
+func main() {
+	if ret := nvml.Init(); ret != nvml.SUCCESS {
+		fmt.Fprintf(os.Stderr, "nvmlInit: %v\n", ret)
+		os.Exit(1)
+	}
+	defer func() { _ = nvml.Shutdown() }()
+
+	count, ret := nvml.DeviceGetCount()
+	if ret != nvml.SUCCESS {
+		fmt.Fprintf(os.Stderr, "DeviceGetCount: %v\n", ret)
+		os.Exit(1)
+	}
+
+	notSupported := 0
+	fmt.Printf("Discovered %d GPU(s)\n", count)
+	for i := 0; i < count; i++ {
+		dev, ret := nvml.DeviceGetHandleByIndex(i)
+		if ret != nvml.SUCCESS {
+			fmt.Fprintf(os.Stderr, "handle[%d]: %v\n", i, ret)
+			os.Exit(1)
+		}
+		uuid, _ := dev.GetUUID()
+
+		info, ret := dev.GetGpuFabricInfo()
+		switch ret {
+		case nvml.SUCCESS:
+			fmt.Printf("GPU %d (%s)\n", i, uuid)
+			fmt.Printf("  clusterUuid : %s\n", formatUUID(info.ClusterUuid[:]))
+			fmt.Printf("  cliqueId    : %d\n", info.CliqueId)
+			fmt.Printf("  state       : %s (%d)\n", stateName(uint8(info.State)), info.State)
+		case nvml.ERROR_NOT_SUPPORTED:
+			fmt.Printf("GPU %d (%s): fabric NOT SUPPORTED\n", i, uuid)
+			notSupported++
+		default:
+			fmt.Fprintf(os.Stderr, "GetGpuFabricInfo[%d]: %v\n", i, ret)
+			os.Exit(1)
+		}
+	}
+
+	if notSupported > 0 {
+		os.Exit(2)
+	}
+}
+
+func formatUUID(b []byte) string {
+	if len(b) != 16 {
+		return hex.EncodeToString(b)
+	}
+	return fmt.Sprintf("%s-%s-%s-%s-%s",
+		hex.EncodeToString(b[0:4]),
+		hex.EncodeToString(b[4:6]),
+		hex.EncodeToString(b[6:8]),
+		hex.EncodeToString(b[8:10]),
+		hex.EncodeToString(b[10:16]),
+	)
+}
+
+func stateName(s uint8) string {
+	switch s {
+	case 0:
+		return "not_supported"
+	case 1:
+		return "not_started"
+	case 2:
+		return "in_progress"
+	case 3:
+		return "completed"
+	default:
+		return "unknown"
+	}
+}
--- a/cmd/fake-imex/ctl/main.go
+++ b/cmd/fake-imex/ctl/main.go
@@ -1,0 +1,56 @@
+// Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// fake-imex-ctl is a drop-in replacement for `nvidia-imex-ctl` used by
+// the compute-domain-daemon's readiness probe. It supports the
+// `-c <config> -q` invocation pattern from upstream's check() helper:
+// prints "READY\n" + exit 0 iff every peer listed in nodes.cfg has a
+// marker file under the shared state directory; otherwise exits 1.
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/NVIDIA/k8s-test-infra/pkg/imexcoord"
+)
+
+func main() {
+	cfg := flag.String("c", "", "imexd config file (ignored)")
+	query := flag.Bool("q", false, "query readiness (only supported mode)")
+	flag.Parse()
+	_ = cfg
+
+	if !*query {
+		fmt.Fprintln(os.Stderr, "fake-imex-ctl: only -q (query) mode is supported")
+		os.Exit(2)
+	}
+
+	peers, err := imexcoord.ReadPeers(imexcoord.NodesConfigPath())
+	if err != nil {
+		// Match real nvidia-imex-ctl behaviour: when the daemon has not
+		// yet written nodes.cfg the probe should fail rather than say
+		// READY. Avoids a race where the controller marks a domain
+		// ready before its members have actually started.
+		fmt.Fprintf(os.Stderr, "fake-imex-ctl: read nodes.cfg: %v\n", err)
+		os.Exit(1)
+	}
+
+	ready, missing := imexcoord.AllPeersReady(imexcoord.StateDir(), peers)
+	if !ready {
+		fmt.Fprintf(os.Stderr, "fake-imex-ctl: peer %s not ready\n", missing)
+		os.Exit(1)
+	}
+	fmt.Print("READY\n")
+}
--- a/cmd/fake-imex/daemon/main.go
+++ b/cmd/fake-imex/daemon/main.go
@@ -1,0 +1,89 @@
+// Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// fake-imex-daemon is a drop-in replacement for the real `nvidia-imex`
+// binary used inside the compute-domain-daemon pod. It accepts (and
+// ignores) the standard CLI flags, writes a marker file under the
+// shared hostPath state directory, and re-reads nodes.cfg on SIGUSR1
+// (matching the upstream daemon's DNS-update signal). See
+// NVIDIA/k8s-test-infra#304 for the full ComputeDomain simulation
+// design.
+package main
+
+import (
+	"flag"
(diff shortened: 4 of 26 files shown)


ArangoGutierrez (changes requested):
Really nice work — the version-dispatch in `nvmlDeviceGetGpuFabricInfoV` and the `GetMockFabricInfo*` naming to dodge the embedded `dgxa100.Device` shadow are both good calls. Demo runs all three scenarios cleanly.

Two things I'd want to fix before merging:

1. `overrideFabric` in `engine/config.go` creates a `FabricConfig` ex-nihilo when the profile has none. That means `topology.enabled=true` with `gpu.profile=a100` (or any non-fabric profile) will silently start reporting fabric info on every A100. The profile is the gate — if `DeviceDefaults.Fabric == nil` going in, the overlay should be a no-op (maybe a debug log so it's discoverable). Want me to push a fix on top?

2. `FabricConfig.PartitionID` is plumbed through `FabricInfo` but never written by the bridge — `nvml_types.h` doesn't have a `partitionId` field on any of v1/v2/v3, and `check-fabric` doesn't print it. Easiest path is to drop it from `FabricConfig` for now and add it back when we have a real v3 struct that carries it.

One doc gap worth covering: the chart wires `NODE_NAME`, `MOCK_TOPOLOGY_CONFIG`, and the topology mount only on the nvml-mock DaemonSet, so `check-fabric` running inside that pod works trivially. The README mentions `Dockerfile.compute-domain-daemon` for swapping the binaries on the real daemon, but the real daemon also needs those same three env/mount additions on its own pod for the fabric overlay to actually take effect. A short "if you're using this with the upstream chart, also patch its values to set `NODE_NAME` from downward API and mount the topology ConfigMap" paragraph would save the next user an afternoon.

Nice-to-haves (happy to do these as a follow-up PR on my side so this can land):
- Bridge layout-size unit test asserting `unsafe.Sizeof(nvml.GpuFabricInfo{}) == C.sizeof_nvmlGpuFabricInfo_t` (and v2/v3). Catches future header drift.
- Test for malformed `topology.yaml` — currently a typo logs a warn and silently disables the overlay.
- `overrideFabric` iterates `yamlConfig.Devices` rewriting per-device `Fabric` blocks, but no profile ships per-device fabric. Either drop the loop or `// forward-compat`.

giuliocalzo:
Thanks for the review @ArangoGutierrez — addressed all three blockers in the force-push:

1. **`overrideFabric` gate** — `applyTopologyOverlay` now short-circuits with a debug log when `DeviceDefaults.Fabric == nil`, so the profile stays the gate. Added `TestTopologyOverlay_NoFabricDefaults_NoOp` as a regression. The per-device loop survives with a `// forward-compat` comment since the cost is one nil-check and it keeps a future per-device fabric profile coherent.
2. **`PartitionID` dropped** — removed from `FabricConfig` and `FabricInfo`, from the `gb200` profile + helm snapshot, and from the v-version debug log. Will reintroduce when there's a real NVML struct version that carries it.
3. **Upstream daemon docs** — added a "Heads up — patching the upstream chart" block to `docs/demo/compute-domain/README.md` covering the three additions the real compute-domain-daemon pod needs (`NODE_NAME` from downward API, `MOCK_TOPOLOGY_CONFIG`, and the topology ConfigMap mount), so swapping in `Dockerfile.compute-domain-daemon` alone won't quietly fall back to defaults.

Local verification: `go test ./pkg/gpu/mocknvml/engine/... ./pkg/imexcoord/... ./cmd/fake-imex/...` and `helm unittest deployments/nvml-mock/helm/nvml-mock` both green; `docs/demo/compute-domain/run.sh` still passes all three scenarios end-to-end.

Happy for you to take the nice-to-haves (bridge layout-size test + malformed topology test) as a follow-up if that helps this land.

ArangoGutierrez (approved):
All three landed cleanly — the gate comment on `applyTopologyOverlay` reads well, the `TestTopologyOverlay_NoFabricDefaults_NoOp` regression is exactly right, and the "Heads up — patching the upstream chart" block calls out the three specific additions plus the silent-fallback failure mode. LGTM.

Will pick up the bridge layout-size test and the malformed-topology test as a follow-up PR.

ArangoGutierrez (changes requested):
One more thing surfaced from a deeper pass that's worth catching pre-merge — `pkg/gpu/mocknvml/bridge/fabric.go:114-117` treats `requested == 0` as v3 and writes the full v3 layout (including `healthSummary`). If a caller allocates a v2-sized buffer and casts to `*nvmlGpuFabricInfoV_t` without setting `.version`, that overruns by `sizeof(healthSummary)` + alignment. Real NVML returns `NVML_ERROR_ARGUMENT_VERSION_MISMATCH` for zero. Suggest dropping the permissive branch — only `v2` or `v3` should succeed, everything else (including zero) returns mismatch.

A small dispatch unit test covering v1/v2/v3/garbage version values would also catch any future regression here. Happy to add it as part of a follow-up PR on my side along with the imexcoord doc/hardening notes the dist-sys pass turned up — keeping this PR focused on the one defect.

ArangoGutierrez (approved):
Approving — taking ownership of the V2/V3 zero-version fix in our follow-up PR along with the dispatch test, MarkerPath validation, ctl self-check, daemon marker re-assertion, and imexcoord doc clarifications. Don't want to keep this PR in flight on one defect when there's a coherent batch of related work to land.

PR #353, opened 2026-05-19
feat(nvml-mock): add PCIe topology to profiles and render mock sysfs tree

giuliocalzo:
## Summary

Closes #263.

The DRA driver and topology-aware schedulers resolve "which PCIe root complex a GPU lives on" (`dra.k8s.io/pcieRoot`) by `readlink()`ing `/sys/bus/pci/devices/<bdf>` and parsing the path. The mock NVML library returned valid PCI BDFs via NVML but produced no corresponding `/sys/bus/pci` tree, so any consumer doing the readlink dance came back empty-handed and could not surface topology hints for scheduling.

This PR closes that gap end-to-end:

### Schema

- New `pcie_topology:` block in every profile (a100, b200, gb200, h100, l40s, t4) describing PCI root complexes, NUMA nodes, and device-to-root mapping.
- Defaults populated for all six profiles:
  | Profile | Root complexes | NUMA nodes | Layout |
  |---|---|---|---|
  | `a100` / `b200` / `h100` / `l40s` | 2 (`pci0000:00`, `pci0000:80`) | 2 (dual socket) | 4 GPUs each |
  | `gb200` | 4 (`pci0000:00`, `:40`, `:80`, `:c0`) | 4 (per Grace pair) | 2 GPUs each |
  | `t4` | 1 (`pci0000:00`) | 1 | 4 GPUs |
- Same block lives in both `pkg/gpu/mocknvml/configs/` and `deployments/.../profiles/`.

### BDF format migration (canonical Linux 4-digit-domain form)

- Profile `bus_id` fields now use `0000:07:00.0` instead of the NVML `busIdLegacy` `00000000:07:00.0`. The bridge already copied the string verbatim into `nvmlPciInfo.busId`, so this aligns with what real Linux PCI sysfs exposes and is a hard prerequisite for the renderer's readlink contract.
- `Validate()` in the schema rejects the legacy 8-digit form **explicitly** so a half-migrated profile fails loudly instead of silently producing a tree the DRA driver can't resolve.
- Default fallback in `engine.go` (used when no YAML is loaded) migrated; tests / docs / nvidia-smi output examples updated (the `nvidia-smi` Bus-Id column intentionally keeps the 8-digit form because that's what real `nvidia-smi` prints).

### Renderer + CLI

New package `pkg/system/mockpcisysfs/` (decoupled from cgo / the engine, same pattern as `pkg/network/mockibsysfs`):

- `config/types.go` — Profile / PCIeTopology / RootComplex schema with `Validate()` (rejects 8-digit BDFs, duplicate root complex IDs, BDFs under multiple roots, BDFs that have no matching `devices:` entry) + `EffectiveTopology()` (flat single-root fallback for profiles without an explicit block).
- `render/render.go` — writes:
  - `<output>/sys/devices/pciDDDD:BB/<bdf>/numa_node`
  - `<output>/sys/bus/pci/devices/<bdf>` → relative symlink `../../../devices/pciDDDD:BB/<bdf>`
  
  Mirrors what the kernel exposes so `EvalSymlinks` lands on the right path. Idempotent: re-renders replace stale symlinks so editing a topology between runs cannot leave a misattributed pcieRoot in place. BDFs lowercased on write to match the kernel.
- `cmd/render-pci-sysfs/` — CLI binary mirroring `render-ib-sysfs`. Supports `--strict` (require explicit `pcie_topology:`) and `--dry-run`.

### DaemonSet wiring

- `Dockerfile` builds the new binary and installs it at `/usr/local/bin/render-pci-sysfs`.
- `setup.sh` invokes it after the IB renderer. Failures are **fatal under `set -e`** for the same reason as the IB block — a typo would silently render an empty/malformed tree that downstream `pcieRoot` attributes would inherit.
- Output lands at `/var/lib/nvml-mock/sys/...` — the canonical hostPath the rest of the nvml-mock plumbing already mounts.

### Tests

- **Schema:** canonical accept, legacy 8-digit BDF reject, malformed root reject, duplicate-BDF reject, unknown-BDF reject, `EffectiveTopology` default + override.
- **Renderer:** full tree end-to-end (verifies the `EvalSymlinks` round-trip the deviceattribute library performs), relative-symlink target shape, idempotent rerender across topology changes, uppercase-BDF normalization, no-op when topology is nil, error when `Output` is empty.
- **Helm unittest:** all 94 tests + 14 snapshots pass after regeneration. The daemonset snapshot diff is driven entirely by the `checksum/config` annotation hashing the new profile content.
- **All existing `pkg/gpu/mocknvml/...` tests still pass** (bridge, engine, fabric).

### Acceptance criteria (from #263)

| Criterion | Status |
|---|---|
| Fix BDF format in profile YAMLs (8-digit -> 4-digit) | ✅ |
| Add `pcie_topology` section with root complex IDs, NUMA nodes, device mappings | ✅ |
| Populate PCIe topology for all profiles (A100, H100, B200, GB200, L40S, T4) | ✅ |
| Go sysfs generator creating symlinks + `numa_node` files from profile data | ✅ |
| Integrate generator into `setup.sh` init container | ✅ |
| Unit tests for sysfs tree structure validation | ✅ |

### Design note on schema placement

The schema lives in `pkg/system/mockpcisysfs/config` rather than being added to the engine's `YAMLConfig` in `pkg/gpu/mocknvml/engine/config_types.go`. This follows the established sidecar pattern from `pkg/network/mockibsysfs/config` (the IB renderer parses its own slice of the profile YAML and the engine's `YAMLConfig` doesn't carry an `Infiniband` field either). Keeping the renderer cgo-free means it builds into a small static binary for the init container.

## Test plan

Local verification on darwin (cross-build host):

- [x] `go test ./pkg/system/mockpcisysfs/... ./cmd/render-pci-sysfs/...` — pass
- [x] `go test ./pkg/gpu/mocknvml/...` — pass (no regressions)
- [x] `go build -mod=vendor ./...` — clean
- [x] `go vet ./pkg/system/mockpcisysfs/... ./cmd/render-pci-sysfs/...` — clean
- [x] `helm unittest deployments/nvml-mock/helm/nvml-mock` — 94/94 tests, 14/14 snapshots pass
- [x] `helm lint deployments/nvml-mock/helm/nvml-mock` — clean
- [x] `render-pci-sysfs --strict --dry-run` against all 12 profile YAMLs — accepts every one (2/2/4/2/2/1 root complexes, 8/8/8/8/8/4 devices respectively)
- [x] Smoke: `render-pci-sysfs --config configs/mock-nvml-config-a100.yaml --output /tmp/pci-test` produces the expected symlinks + `numa_node` files; `EvalSymlinks` of `sys/bus/pci/devices/0000:07:00.0` lands on `sys/devices/pci0000:00/0000:07:00.0`.

CI to verify on Linux:

- [ ] `make docker-build` (multi-arch) — confirm the new binary ships in the image.
- [ ] e2e: DaemonSet starts cleanly on KIND and `kubectl exec ds/nvml-mock -- ls /var/lib/nvml-mock/sys/bus/pci/devices` shows the expected BDFs per profile.
- [ ] DRA driver pointed at the mock surfaces `dra.k8s.io/pcieRoot` on ResourceSlices (follow-up if not in scope of the existing e2e jobs).

### DIFF
--- a/deployments/nvml-mock/scripts/setup.sh
+++ b/deployments/nvml-mock/scripts/setup.sh
@@ -240,6 +240,7 @@ fi
 #     malformed sysfs that downstream `dra.k8s.io/pcieRoot` attributes
 #     would inherit.
 PCI_ROOT="$HOST"
+mkdir -p "$PCI_ROOT"
 if [ -x /usr/local/bin/render-pci-sysfs ]; then
   /usr/local/bin/render-pci-sysfs \
     --config /etc/nvml-mock/config.yaml \
--- a/deployments/nvml-mock/scripts/setup.sh
+++ b/deployments/nvml-mock/scripts/setup.sh
@@ -230,4 +230,20 @@ if [ -x /usr/local/bin/render-ib-sysfs ]; then
     --output "$IB_ROOT"
 fi
 
+# 10. Render fake PCI sysfs tree (consumed by topology-aware DRA / device
+#     plugins that resolve PCIe root complex via a readlink on
+#     /sys/bus/pci/devices/<bdf>). The renderer parses the profile's
+#     `pcie_topology:` block; profiles without one get a flat default
+#     covering every device under a single root complex (`pci0000:00`,
+#     NUMA 0). Failures are fatal under `set -e` for the same reason as
+#     the IB block above — a topology typo otherwise yields silently
+#     malformed sysfs that downstream `dra.k8s.io/pcieRoot` attributes
+#     would inherit.
+PCI_ROOT="$HOST"
+if [ -x /usr/local/bin/render-pci-sysfs ]; then
+  /usr/local/bin/render-pci-sysfs \
+    --config /etc/nvml-mock/config.yaml \
+    --output "$PCI_ROOT"
+fi
+
 echo "Mock GPU environment ready: $GPU_COUNT GPUs at $HOST"
--- a/pkg/system/mockpcisysfs/config/types.go
+++ b/pkg/system/mockpcisysfs/config/types.go
@@ -1,0 +1,153 @@
+// Copyright 2026 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+// Package config defines the YAML schema for the `pcie_topology:` block
+// embedded in mock-nvml profile configs. The renderer consumes this to
+// populate a fake `/sys/bus/pci/devices` + `/sys/devices/pciDDDD:BB` tree
+// under MOCK_PCI_ROOT.
+//
+// Keeping the schema in a standalone package (no cgo dependency on the
+// mocknvml engine) mirrors `pkg/network/mockibsysfs/config` and lets the
+// renderer CLI build into a small static binary.
+package config
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+// Profile is the minimal slice of the mock-nvml profile YAML that the PCI
+// sysfs renderer cares about. It deliberately ignores every GPU detail
+// other than the per-device bus_id, which is the join key into the
+// topology block.
+type Profile struct {
+	Devices      []Device      `json:"devices"      yaml:"devices"`
+	PCIeTopology *PCIeTopology `json:"pcie_topology,omitempty" yaml:"pcie_topology,omitempty"`
+}
+
+// Device captures only the fields the renderer needs: the device index
+// (used for --gpu-count clipping) and the PCI bus_id. Every other profile
+// field is ignored at unmarshal time.
+type Device struct {
+	Index int `json:"index" yaml:"index"`
+	PCI   PCI `json:"pci"   yaml:"pci"`
+}
+
+// PCI is the inner block on each device entry. Only bus_id matters here.
+type PCI struct {
+	BusID string `json:"bus_id" yaml:"bus_id"`
+}
+
+// PCIeTopology describes the root-complex layout that the renderer
+// materializes into sysfs. Each entry under root_complexes lists the
+// devices physically attached to that root.
+type PCIeTopology struct {
+	RootComplexes []RootComplex `json:"root_complexes" yaml:"root_complexes"`
+}
+
+// RootComplex represents a single PCI host bridge (`pciDDDD:BB` in real
+// Linux sysfs) along with the NUMA node it lives on and the BDFs of
+// every device attached underneath.
+type RootComplex struct {
+	// ID is the sysfs directory name, e.g. "pci0000:00". Must start with
+	// "pci" and match the kernel's printf format ("pciDDDD:BB").
+	ID string `json:"id" yaml:"id"`
+
+	// NUMANode is the value written into each child device's `numa_node`
+	// file. Real x86 hosts use -1 for "no affinity"; the renderer accepts
+	// any int (including negatives) verbatim.
+	NUMANode int `json:"numa_node" yaml:"numa_node"`
+
+	// Devices is the list of GPU PCI BDFs (4-digit-domain form, e.g.
+	// "0000:07:00.0") that live under this root complex.
+	Devices []string `json:"devices" yaml:"devices"`
+}
+
+// bdfRE matches a canonical Linux sysfs BDF ("0000:07:00.0"). NVML's
+// `busIdLegacy` 8-digit form is intentionally rejected here so the
+// renderer fails loudly if a profile still carries it.
+var bdfRE = regexp.MustCompile(`^[0-9a-fA-F]{4}:[0-9a-fA-F]{2}:[0-9a-fA-F]{2}\.[0-9a-fA-F]$`)
+
+// rootRE matches a sysfs root-complex directory name ("pci0000:00").
+var rootRE = regexp.MustCompile(`^pci[0-9a-fA-F]{4}:[0-9a-fA-F]{2}$`)
+
+// Validate cross-checks the topology against the device list. It returns
+// the first violation encountered:
+//   - root complex IDs are well-formed and unique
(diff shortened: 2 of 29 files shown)


ArangoGutierrez (commented):
Nice work. Clean split between schema package and renderer, the EvalSymlinks round-trip and idempotent-rerender tests are exactly the right regression nets for the readlink consumer, and Validate rejecting the legacy 8-digit form explicitly means a half-migrated profile fails loudly instead of silently producing a tree the DRA driver can't resolve.

The BDF migration is the only piece with real blast radius — bridge-side it flows through `d.pciInfo.BusId` (engine/device.go:186-187), which is the modern 4-digit slot per NVML semantics, so this is a correctness fix rather than a break. `busIdLegacy` was already unset and stays unset; no regression there.

A few inline notes below. LGTM once those are sorted.

pkg/system/mockpcisysfs/config/types.go
```diff
@@ -0,0 +1,153 @@
...
+type Profile struct {
+	Devices      []Device      `json:"devices"      yaml:"devices"`
+	PCIeTopology *PCIeTopology `json:"pcie_topology,omitempty" yaml:"pcie_topology,omitempty"`
+}
+
+// Device captures only the fields the renderer needs: the device index
+// (used for --gpu-count clipping) and the PCI bus_id. Every other profile
+// field is ignored at unmarshal time.
+type Device struct {
+	Index int `json:"index" yaml:"index"`
```
ArangoGutierrez, line 34: The comment on `Device` says `Index` is "used for --gpu-count clipping" but I don't see clipping logic anywhere in this PR, and the renderer doesn't reference `Index` either. Either drop the field or update the comment so the next reader doesn't go hunting for code that isn't there.

pkg/system/mockpcisysfs/config/types.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+	seenRoot := make(map[string]struct{}, len(p.PCIeTopology.RootComplexes))
+	seenBDF := make(map[string]struct{})
+	for _, rc := range p.PCIeTopology.RootComplexes {
+		if !rootRE.MatchString(rc.ID) {
+			return fmt.Errorf("pcie_topology: invalid root complex id %q (want %q)", rc.ID, "pciDDDD:BB")
+		}
+		key := strings.ToLower(rc.ID)
+		if _, dup := seenRoot[key]; dup {
```
ArangoGutierrez, line 106: The duplicate-root check (the `seenRoot` map) has no dedicated test — `TestValidate_RejectsDuplicateBDF` covers BDFs but not roots. A one-line mirror test would close the gap cheaply.

pkg/system/mockpcisysfs/render/render.go
```diff
@@ -0,0 +1,147 @@
...
+	// <Output>/sys/... — Output itself is created if missing.
+	Output string
+}
+
+// Render writes the entire tree. It is idempotent: existing directories
+// are reused, existing files are truncated and rewritten, and existing
+// symlinks are removed and recreated so a stale relative target does not
+// linger across re-renders.
+func Render(o Options) error {
+	if o.Topology == nil || len(o.Topology.RootComplexes) == 0 {
```
ArangoGutierrez, line 54: `Render` no-ops when `Topology` is nil even if `Output == ""`, but errors when `Topology` is non-nil and `Output == ""`. Intentional per the comment ("invoked unconditionally from setup.sh"), but the asymmetry is worth a line in the godoc on `Options` so a future caller doesn't read it as a bug.

deployments/nvml-mock/scripts/setup.sh
```diff
@@ -230,4 +230,20 @@ if [ -x /usr/local/bin/render-ib-sysfs ]; then
...
+# 10. Render fake PCI sysfs tree (consumed by topology-aware DRA / device
+#     plugins that resolve PCIe root complex via a readlink on
+#     /sys/bus/pci/devices/<bdf>). The renderer parses the profile's
+#     `pcie_topology:` block; profiles without one get a flat default
+#     covering every device under a single root complex (`pci0000:00`,
+#     NUMA 0). Failures are fatal under `set -e` for the same reason as
+#     the IB block above — a topology typo otherwise yields silently
+#     malformed sysfs that downstream `dra.k8s.io/pcieRoot` attributes
+#     would inherit.
+PCI_ROOT="$HOST"
```
ArangoGutierrez, line 242: Nit: the IB block does `mkdir -p "$IB_ROOT"` before invoking `render-ib-sysfs`; this one relies on `os.MkdirAll` inside the binary. Works fine since `$HOST` already exists by step 1/2, but slightly inconsistent with the block right above.

ArangoGutierrez (approved):
LGTM — clean separation of schema and renderer, validation gating is tight, regression tests for readlink round-trip and idempotent rerender are exactly right.

shivamerla:
@ArangoGutierrez @giuliocalzo This change broke PCI bus ID parsing in go-nvlib [here](https://github.com/NVIDIA/go-nvlib/blob/8ff29bbf310da9472e5b7c7b41fe18779dd7b20d/pkg/nvlib/device/device.go#L197).

NVML returns PCI bus ID with domain as 8 hex digits (32 bit formatting), but this change returns it as 4 digits with the mock library.

```
Bus Id                            : 00000000:D8:00.0
```

This breaks when go-nvlib trims 4 digits from the bus ID here and ends up returning `:0a:00.0` and error from clients as ```error getting info for GPU 0: error getting PCI bus ID for device 0: invalid PCI Bus ID format: :0a:00.0```.

DRA driver [pipeline](https://github.com/kubernetes-sigs/dra-driver-nvidia-gpu/actions/runs/26463356255/job/77916962539) is failing due to this, cc @dims

PR #367, opened 2026-05-26
feat(nvml-mock): cross-node ibping, iblinkinfo, and ibv_devinfo on Kind without InfiniBand hardware

giuliocalzo:
## Summary

Turns the mock-IB stack into a near-complete InfiniBand diagnostic surface on Kind without any IB hardware.

For IB-enabled profiles, each nvml-mock pod gets a fake sysfs view of HCAs/ports, three `LD_PRELOAD` shims that intercept libc / `libibumad` / `libibverbs`, and an in-pod `mock-ib` daemon that synthesizes subnet-management traffic and forwards MADs to peers. Unmodified `infiniband-diags` / `rdma-core` binaries run against this stack and produce realistic output. Profiles with `infiniband.enabled=false` skip the daemon, fabric port, Service, and NetworkPolicy.

## What works now

| Tool | Status | Path |
|---|---|---|
| `ibstat`, `ibstatus` | ✅ | sysfs via `libibmocksys.so` |
| `ibping` (LID + GUID) | ✅ | UMAD via `libibmockumad.so` → `mock-ib` |
| `iblinkinfo` | ✅ (new) | SMP synthesis (NODE_DESC / NODE_INFO / PORT_INFO) |
| `ibv_devinfo -l`, `ibv_devices` | ✅ (new) | `libibverbs` → libmlx5 via kernel-grammar `modalias` |
| `ibv_devinfo` (full) | ❌ by design | needs real uverbs `ioctl()`s — see Limitations |
| `ibping -S` (server mode) | ❌ | not supported by the UMAD shim |

## What's in this PR

- **`pkg/network/mockib/subnet/`** — SMP synthesis with `libibmad`'s BITSOFFS word layout (`SetFieldSpec`) and DR-path resolution that reads `mad[128 + (i+1)]`, keeping `path[0]` reserved per IB spec Vol 1 §14.2.1.2.
- **`pkg/network/mockib/fabric/`** — in-memory port graph + per-hop neighbor selection so `libibnetdisc` finishes a multi-hop walk without "Duplicate Port" aborts.
- **Stable mock identities** — rendered GUIDs now keep the configured upper 4 bytes fixed and use the lower 32 bits for a 23-bit node hash, 8-bit HCA index, and EUI-64 U/L bit; LIDs use the configured HCA count as a stride across the unicast range to avoid HCA index wrap.
- **Helm fabric wiring** — mock-IB is enabled only when the rendered profile has `infiniband.enabled=true`; the headless ibping Service and NetworkPolicy are skipped for non-IB profiles, and Service discovery matches app name only so multi-release fleets can discover peers after restarts.
- **`libibmockverbs.so`** — third LD_PRELOAD shim proxying `/dev/infiniband/uverbsN` to `mock-ib`. Preload order is now `umad:verbs:sys`.
- **Kernel-grammar `modalias`** — `render.go` writes `pci:v000015B3d00001017sv000015B3sd00000008bc02sc00i00`, the exact pattern libmlx5's `fnmatch` table expects.
- **Image** — Dockerfile installs `ibverbs-providers` (ships libmlx5) and pre-creates `/etc/libibverbs.d/` to silence the libibverbs config-directory warning.
- **`MOCK_IB_DEBUG_SMP=1`** — daemon logs every synthesized SMP (attribute, DLID, hop count, first outbound path bytes, resolved target). The single most useful knob for debugging future DR-walk regressions.
- **E2E validators** — `tests/e2e/validate-ibping.sh`, `tests/e2e/validate-iblinkinfo.sh`, and `tests/e2e/validate-ibv-devinfo.sh`, wired into `docs/demo/standalone/demo.sh` with disabled-IB profiles skipped instead of failed.
- **`tests/redeploy.sh`** — timestamped, always-`--no-cache` rebuild + Kind reload + DaemonSet rollout helper so C-shim/renderer edits are never masked by Docker's layer cache during local dev.
- **Docs** — `pkg/network/mockib/README.md`, `deployments/nvml-mock/helm/nvml-mock/README.md`, `docs/demo/standalone/README.md` refreshed: new components, support matrix, identity contract, `modalias` contract, explicit non-support of full `ibv_devinfo`.

## How it works

1. **`mock-ib`** renders per-node sysfs (unique GUID/LID, kernel-grammar `modalias`) and, for IB-enabled profiles, runs as a daemon serving UMAD over a Unix socket, SMP synthesis over the in-memory fabric graph, uverbs RPC, and an optional TCP fabric registry for cross-pod LID/GUID resolution.
2. **`libibmocksys` / `libibmockumad` / `libibmockverbs`** intercept the three integration points so unmodified diagnostic tools route through `mock-ib`.
3. Remote MADs and SA PathRecord queries (for GUID-mode `ibping`) are resolved via the pod registry and forwarded over the cluster network. Same-release pods discover each other via the headless Service; multi-release fleets can also register peers explicitly.

See `pkg/network/mockib/README.md` for the full component map, env vars, and `iblinkinfo` / `ibv_devinfo` design notes.

## Test plan

- [x] `go test ./pkg/network/mockib/...` — renderer, daemon, subnet, fabric, gid, protocol, registry, sysfs — all green
- [x] `helm unittest deployments/nvml-mock/helm/nvml-mock` — 107/107 tests, 15/15 snapshots
- [x] `bash -n docs/demo/standalone/demo.sh tests/e2e/validate-ibping.sh deployments/nvml-mock/scripts/setup.sh deployments/nvml-mock/scripts/start-mock-ib.sh`
- [x] Kind: `tests/e2e/validate-ibping.sh` — cross-node LID + GUID, 3/3 received
- [x] Kind: `tests/e2e/validate-iblinkinfo.sh` — peer GUIDs reported, no duplicate-port aborts
- [x] Kind: `tests/e2e/validate-ibv-devinfo.sh` — `ibv_devinfo -l` lists 8 HCAs, `ibstatus` confirms ACTIVE / LinkUp on every port
- [x] `docs/demo/standalone/demo.sh` end-to-end on a 1 control-plane / 3 worker Kind cluster

## Limitations

- **Full `ibv_devinfo`** (no `-l`) is intentionally not supported. Once libibverbs claims a device, libmlx5's `verbs_open_device` issues real uverbs `ioctl()`s on `/dev/infiniband/uverbsN` that a userspace `LD_PRELOAD` shim cannot fake. The same per-port data (state, phys state, GID, LID, rate, link layer) is available through `ibstatus`; `validate-ibv-devinfo.sh` exercises both paths.
- **`ibping -S`** (server mode) is not supported by the UMAD shim; cross-node validation uses client-mode pings only.


pkg/network/mockib/daemon/verbs.go
```diff
@@ -0,0 +1,263 @@
...
+
+func (s *Server) readPortLID(caName string) (uint16, error) {
+	path := filepath.Join(s.cfg.IBRoot, "sys/class/infiniband", caName, "ports/1/lid")
+	raw, err := os.ReadFile(path)
+	if err != nil {
+		return 0, err
+	}
+	sv := strings.TrimSpace(string(raw))
+	if strings.HasPrefix(sv, "0x") {
+		v, err := strconv.ParseUint(sv, 2, 16)
```
ArangoGutierrez: Base 2 can't parse a `0x`-prefixed string, and `render.go` writes the LID as `0x%04x`, so this branch always errors and `buildQueryPortResp` ends up reporting LID 0 (the caller drops the error). Dormant today since only full `ibv_devinfo` reaches `QueryPort`, but `strconv.ParseUint(sv, 0, 16)` fixes it and makes the decimal branch below redundant.

pkg/network/mockib/c/umad_shim.c
```diff
@@ -0,0 +1,645 @@
...
+    uint32_t agent_id;
+    uint32_t status;
+    uint32_t timeout_ms;
+    uint32_t retries;
+    uint32_t length;
+    ib_mad_addr_t addr;
+    uint8_t data[0];
+} ib_user_mad_t;
+
+#define MOCK_MAX_FRAME (1 << 20)
```
ArangoGutierrez, line 66: Heads-up: several functions stack-allocate `char resp[MOCK_MAX_FRAME]` (1 MiB) — `umad_recv` here, plus `verbs_open_dev`/`verbs_read_dev` in the verbs shim, which hooks the global `read`/`write`/`open` and can run on threads with small stacks. Responses are tiny in practice; a smaller cap or a heap buffer would be safer.

pkg/network/mockib/daemon/fabric.go
```diff
@@ -0,0 +1,389 @@
...
+}
+
+func (s *Server) acceptFabric(ctx context.Context, ln net.Listener) {
+	for {
+		conn, err := ln.Accept()
+		if err != nil {
+			if ctx.Err() != nil {
+				return
+			}
+			s.log.Printf("mock-ib: fabric accept: %v", err)
```
ArangoGutierrez, line 69: A single transient `Accept` error tears down the fabric accept loop for the rest of the pod's life. Consider `continue` on temporary errors so cross-pod routing survives a blip.

pkg/network/mockib/registry/registry.go
```diff
@@ -0,0 +1,109 @@
...
+		out[k] = v
+	}
+	return out
+}
+
+// LookupByLID returns the peer and port GUID for lid when registered.
+func (r *Registry) LookupByLID(lid uint16) (Peer, string, bool) {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	for guid, p := range r.m {
```
ArangoGutierrez, line 92: Map iteration returns the first LID match nondeterministically if two peers ever share a LID. Shouldn't happen in a healthy fabric, but a deterministic tie-break (lowest PodIP, matching `Register` above) would avoid a flaky surprise.

ArangoGutierrez (dismissed):
Really nice work, Giulio — the shim/daemon split is clean and the comments explaining the *why* are excellent (the `IsSMPSend` class-byte regression, the `NETLINK_RDMA` `socket()` fallback, the `modalias` kernel grammar). They'll save the next person a lot of time. The E2E validators also check for an actual received reply and forbidden error patterns rather than just an exit code, which is exactly right.

Approving. There is one real but currently-dormant bug inline: `readPortLID` parses a hex LID with base 2, so the verbs `QueryPort` path always reports LID 0 — only matters once full `ibv_devinfo` gets wired up, but cheap to fix now. The rest are non-blocking: a couple of 1 MiB stack buffers in the C shims, the fabric accept loop bailing on the first transient error, and a nondeterministic LID lookup in the registry.

One doc nit — the TCP fabric listener binds `0.0.0.0` with no auth and the Unix socket is `0666`. Fine for an isolated Kind fixture, but since the DaemonSet runs privileged, a line in the README noting it's test-cluster-only would set expectations. Nothing here blocks merge.

giuliocalzo:
@ArangoGutierrez fixed all suggestions and improve the network security with a k8s network policy

pkg/network/mockib/subnet/synthesize.go
```diff
@@ -0,0 +1,275 @@
...
+	if len(mad) < 12 {
+		return nil, false
+	}
+
+	hdr, ok := normalizeMADHeader(mad)
+	if !ok {
+		return nil, false
+	}
+
+	cls := hdr[1]
```
ArangoGutierrez: After normalizeMADHeader's word swap, hdr[1] is the original byte 2 (ClassVersion), not MgmtClass — so this class check is really testing ClassVersion. It's the same thing you just fixed in IsSMPSend (which now reads the raw umad[+1]). It works here only because ClassVersion is 0x01 for SMI, and vendor pings get rejected downstream by the AttrID switch. A future GSI/vendor class with ClassVersion 0x01 and AttrID 0x10/0x11/0x15 would slip through. Reading MgmtClass from the raw wire byte (like IsSMPSend) would keep the two consistent.

pkg/network/mockib/subnet/synthesize.go
```diff
@@ -0,0 +1,275 @@
...
+)
+
+// SMPGetHeader carries decoded subnet GET header fields from a UMAD buffer.
+type SMPGetHeader struct {
+	AttrID  uint16
+	AttrMod uint32
+}
+
+// DecodeSMPGet parses a subnet management GET from umad (best-effort).
+func DecodeSMPGet(umad []byte) (SMPGetHeader, bool) {
```
ArangoGutierrez: DecodeSMPGet and SMPGetHeader don't look like they have any caller or test — dead code as far as I can tell. Being exported, the unused linter won't flag it. Probably safe to drop (it's also the second copy of the ClassVersion read above).

deployments/nvml-mock/scripts/setup.sh
```diff
@@ -215,19 +215,20 @@ ln -sfn /var/lib/nvml-mock/driver /host/run/nvidia/driver
...
 mkdir -p "$IB_ROOT"
-if [ -x /usr/local/bin/render-ib-sysfs ]; then
-  /usr/local/bin/render-ib-sysfs \
-    --config /etc/nvml-mock/config.yaml \
-    --gpu-count "$GPU_COUNT" \
-    --node-name "$NODE_NAME" \
-    --output "$IB_ROOT"
+if [ -x /usr/local/bin/mock-ib ]; then
+  if [ "${MOCK_IB:-0}" = "1" ]; then
+    /scripts/start-mock-ib.sh &
```
ArangoGutierrez, line 249: Backgrounding this loses the old fail-fast: render-ib-sysfs used to run synchronously and was fatal under set -e — the previous comment called out that a profile typo would otherwise give an empty tree and zero HCAs with no signal in the logs. Now a render error just exits the backgrounded daemon and setup keeps going. Could run `mock-ib -render-only` synchronously (fatal) before launching the serving daemon to keep that signal.

tests/e2e/validate-ibping.sh
```diff
@@ -0,0 +1,235 @@
...
+# Set IBPING_E2E_MODE=lid|guid|both (default: both).
+set -euo pipefail
+
+SERVER_POD="${1:?Usage: $0 <server-pod> <client-pod>}"
+CLIENT_POD="${2:?Usage: $0 <server-pod> <client-pod>}"
+
+MAX_RETRIES="${IBPING_E2E_RETRIES:-3}"
+RETRY_SLEEP="${IBPING_E2E_RETRY_SLEEP:-5}"
+
+IB_ROOT='${MOCK_IB_ROOT:-/var/lib/nvml-mock/ib}'
```
ArangoGutierrez, line 33: Single quotes make IB_ROOT a literal ${...} string; it only resolves because it's later passed into the pod's `sh -c`, which re-expands it. It works, but it's inconsistent with the double-quoted MOCK_IBPING_SOCKET just below and reads like a slip — a short comment (or double quotes) would clear it up.

pkg/network/mockib/c/verbs_shim.c
```diff
@@ -0,0 +1,423 @@
...
+#include <sys/un.h>
+#include <unistd.h>
+
+#define MOCK_MAX_FRAME (1 << 20)
+/* Daemon responses are small JSON (verbs reads decode into a 4 KiB buffer); cap
+ * stack response buffers well under MOCK_MAX_FRAME so these hooks don't reserve
+ * 1 MiB on threads with small stacks. read_frame still rejects anything larger. */
+#define MOCK_MAX_RESP (1 << 14)
+#define MOCK_MAX_VERBS 32
+#define MOCK_VERBS_FD_BASE 700
```
ArangoGutierrez, line 49: Won't block: synthetic fds at base 700 (and 100 in umad_shim) plus mock_devs/mock_ports touched outside rpc_mu are fine for the single-threaded diag tools here, but a process holding ~700 real fds with a mock device open could mis-route a real read/write. A one-line caveat near the base would be enough.

pkg/network/mockib/daemon/env.go
```diff
@@ -0,0 +1,67 @@
...
+
+const (
+	EnvMockIBConfig     = "MOCK_IB_CONFIG"
+	EnvGPUCount         = "GPU_COUNT"
+	EnvNodeName         = "NODE_NAME"
+	EnvMockIBRoot       = "MOCK_IB_ROOT"
+	EnvMockIBPingSocket = "MOCK_IB_PING_SOCKET"
+	EnvMockIBPingPort   = "MOCK_IB_PING_PORT"
+	EnvMockIBPingFabric = "MOCK_IB_PING_FABRIC"
+	EnvMockIBPeers           = "MOCK_IB_PEERS"
```
ArangoGutierrez, line 20: gofmt example: these two are aligned to a different column than the rest of the block above. `make fmt` realigns the whole const block.

ArangoGutierrez (commented):
Really nice work — this is a lot of surface and the hard parts all look right: the libibmad BITSOFFS layout, the DR path[0]-reserved indexing, the NETLINK_RDMA short-circuit, and the wire-order MgmtClass read. I especially liked that each of those is pinned by a regression test built from a real captured MAD. No blockers from me; a handful of things worth tightening, noted inline.

Two that aren't line-specific:

- gofmt isn't clean on four files (daemon/env.go, fabric/graph.go, protocol/wire.go, subnet/synthesize.go) — only const/struct alignment. `make fmt` fixes all of them. Worth calling out because CI won't catch it: the golangci-lint v2 config here has no formatters section, so the drift slips through.
- The renderer's fail-fast-on-bad-profile got lost in the move to the backgrounded daemon — inline note on setup.sh.

Thanks for the docs, and for the honest section on full ibv_devinfo explaining what doesn't work and why — that's the kind of thing that saves the next person a day.

giuliocalzo:
@ArangoGutierrez thanks for the thorough second pass — all comments are now sorted and the threads resolved. Summary of what changed (commit ac70b3d):

- **`subnet/synthesize.go`** — the MAD header (MgmtClass, Method, AttrID, AttrMod) is now parsed straight from the wire payload like `IsSMPSend`, instead of the word-swapped copy (whose byte 1 was ClassVersion). This closes the GSI/vendor-class-with-ClassVersion-0x01 hole you flagged. Dropped the now-dead `normalizeMADHeader`/`decodeAttrID` and the unused `DecodeSMPGet`/`SMPGetHeader`, and fixed the AttrID offset to wire byte 16. The synthesize/server SMP tests now build real wire-format MADs (consistent with `smp_fabric_test.go`).
- **`scripts/setup.sh`** — the IB sysfs tree is rendered synchronously (fatal under `set -e`) before the serving daemon is backgrounded, restoring the fail-fast-on-bad-profile signal.
- **`c/verbs_shim.c`** — added the caveat about the synthetic fd base (no real fd in `[700, 700+MOCK_MAX_VERBS)` for the single-threaded diag tools).
- **`tests/e2e/validate-ibping.sh`** — commented why `IB_ROOT` is single-quoted (it expands inside the pod's `sh -c`, not on the host runner).
- **gofmt** — realigned the four flagged files (`daemon/env.go`, `fabric/graph.go`, `protocol/wire.go`, `subnet/synthesize.go`) and a couple of others; `gofmt -l` is now clean.

PTAL when you have a moment.

pkg/network/mockib/daemon/sa.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+	resp := synthesizeSAPathRecordResp(sendMad, lid)
+	h.mu.Lock()
+	h.recvQ = append(h.recvQ, resp)
+	h.mu.Unlock()
+	return true
+}
+
+func isSAPathRecordGet(umad []byte) bool {
+	mad := umad[umadMADOffset:]
```
ArangoGutierrez, line 47: `handleSend` only guards `len(req.MAD) == 0`, and `TrySynthesize` returns early for `len < 72`, so a 1-55 byte MAD reaches here and `umad[56:]` panics — which kills the serveConn goroutine (no recover) and takes the daemon down. Only reachable via the local 0666 socket, and the real shim always sends >=56 bytes, so it won't fire in normal use. A `len(req.MAD) < umadMADOffset+24` guard early in `handleSend` would close it.

deployments/nvml-mock/Dockerfile
```diff
@@ -58,8 +60,13 @@ COPY --from=builder /out/check-fabric    /usr/local/bin/check-fabric
 RUN set -eux; \
     apt-get update; \
     apt-get install -y --no-install-recommends curl ca-certificates gnupg2 \
-        infiniband-diags rdma-core; \
+        infiniband-diags rdma-core ibverbs-utils ibverbs-providers python3; \
```
ArangoGutierrez: Is `python3` actually needed at runtime? Nothing in the new scripts or validators references it — looks like a fair bit of image size for no consumer. Drop it if it was a leftover.

pkg/network/mockib/fabric/graph.go
```diff
@@ -0,0 +1,137 @@
...
+	for _, p := range g.ports {
+		if p.CAName == caName && p.Local {
+			return p, true
+		}
+	}
+	return Port{}, false
+}
+
+// OutboundNeighbor picks one stable remote neighbor for local port P (lowest GUID).
+func (g *Graph) OutboundNeighbor(p Port) (Port, bool) {
```
ArangoGutierrez: `OutboundNeighbor` has no callers, and `InboundNeighborFor` below is only used from `graph_test.go` — both look superseded by `PeerAtOutbound`. Worth dropping if so.

pkg/network/mockib/daemon/loopback.go
```diff
@@ -0,0 +1,110 @@
...
+	"sync"
+
+	"github.com/NVIDIA/k8s-test-infra/pkg/network/mockib/gid"
+	"github.com/NVIDIA/k8s-test-infra/pkg/network/mockib/protocol"
+	"github.com/NVIDIA/k8s-test-infra/pkg/network/mockib/registry"
+)
+
+// Loopback matches outbound ping MADs to local port GUIDs under ib-root and synthesizes RECV payloads.
+type Loopback struct {
+	mu     sync.RWMutex
```
ArangoGutierrez: Minor: these maps are built once in `NewLoopback` and never mutated afterwards, so this mutex (taken three times per `matchesLocal` call) isn't guarding anything. Could drop it.

ArangoGutierrez (dismissed):
Read through the whole diff and ran `go test ./pkg/network/mockib/...` locally — all green. Nice work; the layout is clean and the regression tests (IsSMPSend reading the class byte, the SA method / BaseVersion fix, the DR `path[0]`-reserved indexing) are the good kind that fail when the code regresses.

A few small things, none blocking:

- `handleSend` can panic on a short MAD over the Unix socket — inline note, one-line guard closes it.
- `python3` in the Dockerfile looks unused by the new scripts.
- `gb300` is in the README/values tables but I didn't spot a profile definition in this PR — just confirm it exists.
- `OutboundNeighbor` / `InboundNeighborFor` look superseded by `PeerAtOutbound` (inline).

Approving — fold these in or wave them off as you like.

pkg/network/mockib/daemon/sa.go
giuliocalzo: Fixed in 03ea9283. Added an early guard in `handleSend` — `len(req.MAD) < umadMADOffset+ibMADCommonHdrLen` returns a `mad too short` error before any wire-offset slice, so a truncated buffer over the 0666 socket can no longer panic the `serveConn` goroutine. Added a regression test (`TestServer_handleSend_shortMADNoPanic`) that sends a 10-byte MAD and asserts the error is returned and the daemon keeps serving on the same connection. Good catch.

deployments/nvml-mock/Dockerfile
giuliocalzo: Dropped in 03ea9283 — confirmed nothing in the scripts or e2e validators references `python3`, it was a leftover. Image is a bit slimmer now.

pkg/network/mockib/fabric/graph.go
giuliocalzo: Removed both in 03ea9283. `OutboundNeighbor` had no callers and `InboundNeighborFor` was only referenced from `graph_test.go` — the live DR walk goes through `PeerAtOutbound`, so they were vestigial. Updated the graph test accordingly (kept the Build/ByLID coverage).

pkg/network/mockib/daemon/loopback.go
giuliocalzo: Dropped the mutex in 03ea9283. The lookup maps are built once in `NewLoopback` and never mutated, so reads need no synchronization; also removed the now-unused `sync` import.

giuliocalzo:
@ArangoGutierrez thanks for the approval and the thorough final pass — all of the latest comments are now sorted. Pushed in two commits on top of the approved tree:

**`03ea9283` — review nits:**
- **`daemon/sa.go` short-MAD panic** — added an early `len(req.MAD) < umadMADOffset+ibMADCommonHdrLen` guard in `handleSend` (returns `mad too short` before any wire-offset slice), so a truncated buffer over the local `0666` socket can't panic the `serveConn` goroutine. Pinned with a new regression test `TestServer_handleSend_shortMADNoPanic`.
- **`Dockerfile` python3** — dropped; nothing in the scripts/validators references it.
- **`fabric/graph.go` dead helpers** — removed `OutboundNeighbor` (no callers) and `InboundNeighborFor` (test-only); the DR walk uses `PeerAtOutbound`. Test updated.
- **`daemon/loopback.go` no-op mutex** — dropped; the maps are built once in `NewLoopback` and never mutated.

**`5731274e` — earlier follow-ups:**
- **`gb300`** — confirmed `profiles/gb300.yaml` exists with `infiniband.enabled: true`; also added the `gb300` row to the IB support matrices and documented `infiniband.mockTier` in the chart values table, plus reworked the "Disable IB" section (the old inline `gpu.customConfig` example replaced the whole profile).
- **Write deadlines** — `serveConn`/`serveFabricConn` now set a per-frame `SetWriteDeadline`, mirroring the existing read deadlines so a peer that stops reading mid-response can't pin a daemon goroutine.

`gofmt`/`go vet` clean, `go test ./pkg/network/mockib/...` and `helm unittest` all green locally. PTAL when you have a moment.

deployments/nvml-mock/helm/nvml-mock/templates/network-policy-ibping.yaml
```diff
@@ -0,0 +1,38 @@
...
+      {{- include "nvml-mock.selectorLabels" . | nindent 6 }}
+  policyTypes:
+    - Ingress
+  ingress:
+    - from:
+        # Match by app name only (not instance): the fabric spans every
+        # nvml-mock pod cluster-wide, which may be deployed as multiple Helm
+        # releases (e.g. one per node in the ibping-multinode e2e). Scoping to
+        # the release's own instance would block cross-release registration.
+        - podSelector:
```
ArangoGutierrez: A bare podSelector under `from` only matches pods in the policy's own namespace, so the cross-release case described here only works while all releases share one namespace. A fleet split across namespaces would get its REGISTER traffic dropped on enforcing CNIs. Either pair this with a namespaceSelector or narrow the comment to same-namespace releases.

deployments/nvml-mock/helm/nvml-mock/values.yaml
```diff
@@ -76,6 +76,38 @@ terminationGracePeriodSeconds: 10
 # NVML engine looks up its NODE_NAME on startup and overrides each GPU's
 # fabric cluster_uuid / clique_id accordingly. See
 # pkg/gpu/mocknvml/engine/fabric.go and issue NVIDIA/k8s-test-infra#304.
+# InfiniBand ping mock. For profiles with infiniband.enabled=true, the
```
ArangoGutierrez, line 94: This block landed directly under the NODE_NAME/fabric comment that documents `topology:`, so that comment now reads as the intro to `infiniband:` and `topology:` is left without its doc. A blank line above this one (or moving the block below `topology:`) restores the pairing.

tests/e2e/validate-iblinkinfo.sh
```diff
@@ -0,0 +1,98 @@
...
+for P in "$LOCAL_POD" "$PEER_POD"; do
+  if ! kubectl logs "$P" 2>/dev/null | grep -q 'fabric ready'; then
+    echo "WARN: pod $P logs do not show 'fabric ready' yet"
+  fi
+done
+
+# The mock fabric is a point-to-point topology (no switches): each local CA
+# picks ONE outbound neighbor — the lowest-GUID non-self port in the merged
+# graph (see fabric.OutboundNeighbor). For a control-plane pod, that ends up
+# being whichever worker pod happened to claim the lowest GUID range, which
```
ArangoGutierrez, line 36: `fabric.OutboundNeighbor` does not exist; the per-hop selection lives in `fabric.PeerAtOutbound` (pkg/network/mockib/fabric/drpath.go). Same stale name in the echo near line 96.

docs/demo/standalone/demo.sh
```diff
@@ -183,7 +192,38 @@ fi
...
+###############################################################################
+SERVER_POD=""
+CLIENT_POD=""
+if [[ "${IB_ENABLED}" == "true" ]]; then
+  SERVER_POD=$(kubectl get pods -l app.kubernetes.io/name=nvml-mock \
+    --field-selector=status.phase=Running \
+    -o jsonpath='{.items[0].metadata.name}')
+  CLIENT_POD=$(kubectl get pods -l app.kubernetes.io/name=nvml-mock \
+    --field-selector=status.phase=Running \
+    -o jsonpath='{.items[1].metadata.name}')
```
ArangoGutierrez: If only one pod is Running, jsonpath `{.items[1]}` errors out and `set -e` kills the script right here, so the friendlier checks below never run. Fetching `{.items[*].metadata.name}` into an array and checking the count first keeps the nice error message.

ArangoGutierrez (commented):
This was a big diff but a genuinely fun read. The parts that are easiest to get wrong are the best covered: the BITSOFFS word layout and DR path[0] handling are pinned by captured MAD fixtures, IsSMPSend carries a regression test documenting the exact misclassification it guards against, neighbor selection sorts before indexing so nothing depends on map iteration order, and the new GUID/LID encoding comes with collision regressions for the known FNV low-bit pair. The e2e validators check real reply counts rather than exit codes and dump useful state on failure.

Also glad to see the 0.0.0.0 listener and 0666 socket called out explicitly with a NetworkPolicy default, and the short-MAD guard in handleSend so a stray socket client cannot panic the daemon.

Four small comments inline, none blocking. CI is green including the new ibping-multinode job.

giuliocalzo:
@ArangoGutierrez thanks for the latest pass — all four inline comments are sorted in `7d543332`:

- **`network-policy-ibping.yaml`** — added a `namespaceSelector: {}` alongside the existing `podSelector` in a single ingress peer (logical AND), so the cluster-wide cross-release fabric admits `nvml-mock` peers in *any* namespace. A bare `podSelector` only matched the policy's own namespace and would silently drop cross-namespace REGISTER traffic on enforcing CNIs (Calico/Cilium). Pinned with an `exists` assertion in the chart test.
- **`values.yaml`** — moved the `infiniband:` block below `topology:` so the "ComputeDomain (NVLink fabric) simulation" comment documents `topology:` again instead of reading as the intro to `infiniband:`.
- **`tests/e2e/validate-iblinkinfo.sh`** — fixed the two stale `fabric.OutboundNeighbor` references (removed symbol) to `fabric.PeerAtOutbound` (the comment near line 35 and the echo near line 96).
- **`docs/demo/standalone/demo.sh`** — collect the Running `nvml-mock` pod names into an array and check the count before indexing, so a single-pod cluster hits the friendly `fail` message instead of aborting on a `{.items[1]}` jsonpath error under `set -e`.

`helm unittest` (111/111 + 15 snapshots) and `bash -n` on both scripts are green locally. PTAL when you have a moment.

pkg/network/mockib/fabric/drpath.go
```diff
@@ -0,0 +1,87 @@
+// Copyright 2026 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+package fabric
+
+import "sort"
+
+// NeighborsExcept returns all ports other than p in stable graph order.
+func (g *Graph) NeighborsExcept(p Port) []Port {
```
ArangoGutierrez, line 9: nit: NeighborsExcept has no callers anywhere, including tests — looks like the same vestige as OutboundNeighbor/InboundNeighborFor you removed in 03ea9283, just missed here. It's an exported method so the unused linter won't flag it. Safe to drop if PeerAtOutbound fully supersedes it.

pkg/network/mockib/registry/registry_test.go
```diff
@@ -0,0 +1,53 @@
...
+	r.Register("A088C20300AB0001", Peer{PodIP: "10.0.0.2"})
+	p, ok := r.Lookup("a088:c203:00ab:0001")
+	if !ok || p.PodIP != "10.0.0.2" {
+		t.Fatalf("lookup with normalized key failed: %+v %v", p, ok)
+	}
+}
+
+func TestRegistry_LookupByLID(t *testing.T) {
+	r := New()
+	r.Register("a088:c203:00ab:0001", Peer{PodIP: "10.0.0.2", LID: 0x0200})
```
ArangoGutierrez, line 48: nit: only one entry is ever registered here, so the deterministic lowest-PodIP tie-break you added in 24169937 isn't actually exercised — if LookupByLID regressed to map-iteration order this test would still pass. Registering a second peer that shares LID 0x0200 with a higher PodIP would pin the property.

pkg/network/mockib/daemon/verbs.go
```diff
@@ -0,0 +1,274 @@
...
+func (s *Server) readPortLID(caName string) (uint16, error) {
+	path := filepath.Join(s.cfg.IBRoot, "sys/class/infiniband", caName, "ports/1/lid")
+	raw, err := os.ReadFile(path)
+	if err != nil {
+		return 0, err
+	}
+	sv := strings.TrimSpace(string(raw))
+	// base 0 honors a "0x" prefix (render.go writes the LID as 0x%04x) and
+	// still parses bare decimal, so a single call covers both formats.
+	v, err := strconv.ParseUint(sv, 0, 16)
```
ArangoGutierrez, line 240: nit: the base-0 parse fix (was base 2, which always returned LID 0) has no regression test — readPortLID/buildQueryPortResp are uncovered, so a revert to base 2 wouldn't be caught. Low priority since only full ibv_devinfo reaches QueryPort, but a one-liner round-tripping the rendered 0x-prefixed LID would lock it down.

ArangoGutierrez (commented):
Read the whole diff and the test suite end to end. The hard parts are the best covered: captured-MAD fixtures pin IsSMPSend, the SA method/TRID handling, the DR path[0] indexing and the libibmad wire layout, and the GUID/LID encoding ships real collision regressions — all the kind that go red when the code regresses. CI was green on the prior head including ibping-multinode (the e2e jobs are just re-running on your latest push), and the mockibsysfs -> mockib migration is clean with the old package and render-ib-sysfs fully removed. Nothing blocking. Three small nits inline; also the CHANGELOG still says cross-node ibping is "always enabled in the nvml-mock Helm chart", which is no longer quite true now that the daemon and Service only come up for IB-enabled profiles.

PR #368, opened 2026-05-27
feat(mockib): non-zero IB counters + PMA MAD synth (perfquery, Prometheus exporters)

giuliocalzo:
> Stacked on top of #367 — until that lands, the diff here shows its commits too. Review only the counters / PMA work; once #367 merges, this PR's diff reduces to it.

## Summary

Before this branch every InfiniBand counter in mockib's sysfs tree was a hard-coded zero, which made `perfquery`, the Prometheus RDMA exporter, and link-quality alerts useless on Kind.

This PR makes counters live, deterministic and queryable end-to-end:

- Every port's `counters/` (16 legacy 32-bit + 8 extended 64-bit) and `hw_counters/` (23 mlx5-specific) directories now render non-zero, slowly-growing values driven by a deterministic generator keyed on `(NODE_NAME, mlx5_N, counter_name, elapsed)`.

- A ticked writer inside the `mock-ib` daemon rewrites every catalog file atomically every 5 s (configurable via `infiniband.counters.tick_seconds`).

- A Performance Management (PMA) MAD synthesizer answers `ClassPortInfo`, `PortCounters` and `PortCountersExtended` Get requests, plus the `ClearCounters` Set, so `perfquery`, `perfquery -x`, and `perfquery -R` work over the existing libibumad shim. The PMA path is hooked into `Server.handleSend` in front of the fabric/loopback branches and shares the same `Generator` and `Epochs` as the writer, so `cat .../port_xmit_data_64` and `perfquery -x` always agree, and `perfquery -R` resets both surfaces in one call.

- Behaviour is gated by a tri-state `*bool` in YAML: absent ⇒ on, `infiniband.counters.enabled: false` ⇒ render zeros, no writer, no PMA dispatch (zero behaviour change for callers that opt out). `l40s` / `t4` profiles (IB disabled) are unaffected.

Two pods on the same profile produce visibly distinct counter values without any cross-pod coordination — the generator is keyed on the rendered `NODE_NAME`.

See the new [`Counters and perfquery`](pkg/network/mockib/README.md#counters-and-perfquery) section for the full surface map, value families, and field-offset table against libibmad.

## Changes

- `pkg/network/mockib/counters/`: new package — `Catalog` (47 entries), deterministic `Generator`, per-port `Epochs` (reset state).

- `pkg/network/mockib/nodeid/`: extracted `NodeID` helper so the renderer, writer, and PMA all derive identical per-port keys.

- `pkg/network/mockib/config/`: new `Counters{Enabled *bool, TickSeconds int}` sub-block on the IB profile schema.

- `pkg/network/mockib/render/`: catalog-driven sysfs rendering (non-zero by default, zero when disabled).

- `pkg/network/mockib/daemon/counters_writer.go`: atomic per-tick file rewrites, ctx-cancel aware.

- `pkg/network/mockib/daemon/pma.go`: PMA Get/Set synthesizer for `ClassPortInfo`, `PortCounters`, `PortCountersExtended`, `ClearCounters`. Uses the shared `Generator` so the wire response and sysfs always agree.

- `pkg/network/mockib/daemon/server.go`: PMA dispatch in `handleSend`; PMA is excluded from the fabric/loopback paths the same way SMP is, so cross-pod PMA forwarding doesn't accidentally race the synthesizer.

- `pkg/network/mockib/subnet/`: export `NormalizeMADHeader` so PMA and SMP share the same word-swap helper.

- `cmd/mock-ib/main.go`: build the shared `Generator` + `Epochs`, wire them into `Server.SetCounters`, start the writer goroutine.

- `pkg/network/mockib/README.md`: new `Counters and perfquery` section (+104 lines) covering the surface map, value families, PMA field offsets, and the YAML knobs.

- `tests/e2e/validate-counters.sh`: verifies `port_xmit_data_64` grows between two reads and that `hw_counters/` exists.

- `tests/e2e/validate-perfquery.sh`: verifies `perfquery` legacy + extended return non-zero counters and that `perfquery -R` resets them.

- `.github/workflows/nvml-mock-e2e.yaml`: wires both validators into the e2e job.

## Tests

- `go build ./...` clean.

- `go vet ./pkg/network/mockib/... ./cmd/...` clean.

- `go test ./pkg/network/mockib/... -count=1 -race` — all packages pass under `-race`.

- `pma_test.go` covers `ClassPortInfo`, `PortCounters` Get, `PortCountersExtended` Get, `ClearCounters` Set, agreement with the `Generator`, and rejection of unknown CAs / AttrIDs / `GetResp`.

- `server_pma_test.go` exercises the full dispatch chain through `Server.handleSend` over a real Unix socket.

- Writer tests cover ticked growth, epoch reset semantics, atomic rename, and ctx-cancel teardown.

- Local smoke run on a Kind cluster: counter values match the generator's predicted rates within the tick window.

- CI: `validate-counters.sh` and `validate-perfquery.sh` against a Kind cluster (this PR adds both workflow steps).

## Known limitations

- `CounterSelect` (PMA request bitmask) is echoed in the response but not applied — every mapped counter is always populated. `perfquery` uses `0xffff` (all counters) in practice, so this is invisible to real callers; documented in the README PMA table.

- Per-VL counters and `XmitWait` are not modelled — out of scope and not required by `perfquery` or standard exporters.


ArangoGutierrez (commented):
Read through the counters + PMA work (setting aside the #367 base) and ran it locally: `go build`, `go vet`, and `go test ./pkg/network/mockib/{counters,config,daemon}/... -race` all clean.

Really nice shape. One catalog, one generator, one epoch store shared across render, the writer, and the PMA path, so the three surfaces can't drift, and a `-R` reset lands everywhere at once. The tests are the real kind too: the captured perfquery wire fixture pinning the MgmtClass/ClassVersion byte-order fix, the AgreesWithGenerator checks deriving the expected value independently, and the socket-level dispatch test rather than a mock. I also liked that `RenderSysfsFromConfig` applies `Defaults()` before main reads `TickSeconds`, so a minimal `{enabled: true}` block can't fatal the daemon on a zero tick.

Two small inline notes, neither blocking. One more on prose: the "cat ... and perfquery -x always agree" line is true via the shared Generator but not bit-for-bit (different read instants give different elapsed within a tick) — might be worth softening so nobody reads it as exact equality. The e2e only checks growth/non-zero/reset, which is the right thing to assert.

LGTM once you've had a look at the Set-method-byte note.

pkg/network/mockib/daemon/pma.go
```diff
@@ -0,0 +1,276 @@
...
+// full perfquery response, zeros the umad status word, sets the method
+// byte to GetResp, and zeros the PMA payload region so the caller only
+// needs to fill the fields it computes.
+func buildPMAResponse(sendMad []byte, method byte) []byte {
+	out := make([]byte, pmaUMADFullLen)
+	copy(out, sendMad)
+	// umad status word (libibumad layout: little-endian uint32 at byte 4)
+	binary.LittleEndian.PutUint32(out[4:8], 0)
+	respMAD := out[pmaUMADHeaderLen:]
+	respMAD[pmaMADMethodOff] = method | 0x80
```
ArangoGutierrez, line 208: Small spec thing: this makes a Set ClearCounters response come back with method `0x82`, but IBA only defines `GetResp` (`0x81`) as the response to both Get and Set — there's no `0x82` method. `perfquery -R` clearly accepts it (e2e is green), so this is more "worth being strict?" than a bug. Either a one-line comment noting the deviation or just returning `0x81` would do.

pkg/network/mockib/counters/generator.go
```diff
@@ -0,0 +1,117 @@
...
+func (g Generator) delta(family Family, name string, elapsed time.Duration) uint64 {
+	if elapsed <= 0 {
+		return 0
+	}
+	secs := elapsed.Seconds()
+	bps := float64(g.bytesPerSec())
+	switch family {
+	case FamilyXmitData:
+		// 70% utilization; sysfs port_xmit_data is in 4-byte words per IB spec.
+		return uint64(bps * secs * 0.7 / 4.0)
```
ArangoGutierrez, line 88: Tiny note, not asking for a change: past ~2^52 (a 64-bit data counter at 400Gbps gets there in a few hours) float64 loses exact integer precision. Determinism per call still holds and the values are synthetic, so it's harmless — just flagging since the comment above says "well within float64 precision".

pkg/network/mockib/daemon/pma.go
```diff
@@ -0,0 +1,276 @@
...
+func fillPMAClassPortInfo(payload []byte) {
+	subnet.SetFieldSpec(payload, 0, 8, 1)        // BaseVersion
+	subnet.SetFieldSpec(payload, 8, 8, 1)        // ClassVersion
+	subnet.SetFieldSpec(payload, 16, 16, 0x0700) // CapabilityMask (ALL_PORT + EXT_WIDTH + EXT_NOIETF)
+}
+
+func fillPMAPortCounters(payload []byte, caIdx int,
+	gen counters.Generator, epochs *counters.Epochs, now time.Time,
+	portSelect, counterSelect uint32) {
+	subnet.SetFieldSpec(payload, 0, 8, portSelect)
```
ArangoGutierrez, line 240: PortSelect sits at bit offset 8, not 0 — libibmad has IB_PC_PORT_SELECT_F and IB_PC_EXT_PORT_SELECT_F both as BITSOFFS(8,8); byte 0 of the attribute is reserved. As written, the request's PortSelect is read from the reserved byte (always 0) and echoed back there too (here, in the ext fill, and in the Set path), so the actual PortSelect field in every response stays 0. perfquery doesn't check the echo, and pma_test.go builds its fixtures with the same offset-0 convention, so the suite can't catch this one.

pkg/network/mockib/daemon/server.go
```diff
@@ -0,0 +1,337 @@
...
+			h.mu.Unlock()
+			return protocol.WriteMessage(c, protocol.TypeSend, protocol.SendResp{OK: true})
+		}
+	}
+	// Fabric ping + loopback echo is for vendor/ping MADs only; SMP / PMA must use synthesis.
+	if s.cfg.Fabric && !subnet.IsSMPSend(req.MAD) && !IsPMASend(req.MAD) && s.tryFabricSend(h, req.MAD) {
+		return protocol.WriteMessage(c, protocol.TypeSend, protocol.SendResp{OK: true})
+	}
+	// SMP / PMA must not fall through to loopback echo (empty payload breaks iblinkinfo / perfquery).
+	if !subnet.IsSMPSend(req.MAD) && !IsPMASend(req.MAD) {
```
ArangoGutierrez, line 281: When counters are disabled (s.epochs == nil), this gate still keeps PMA MADs out of fabric and loopback, so perfquery gets no reply at all and times out with EAGAIN — before this PR it got the loopback echo and printed zeros. That contradicts the "zero behaviour change for callers that opt out" note in the description; making the !IsPMASend exclusions conditional on s.epochs != nil would restore the old behavior, and a disabled-path test would pin whichever you intend.

pkg/network/mockib/counters/generator.go
```diff
@@ -0,0 +1,117 @@
...
+	RateGbps int    // Profile link rate; controls FamilyXmit/Rcv rate.
+}
+
+// Value returns the counter value at the given elapsed duration since
+// the per-port reset epoch. Width (32 or 64) is enforced via mod 2^Width.
+func (g Generator) Value(caIdx int, e Entry, elapsed time.Duration) uint64 {
+	seed := g.seed(caIdx, e.Name, e.Family)
+	raw := seed + g.delta(e.Family, e.Name, elapsed)
+	if e.Width == 32 {
+		return raw & 0xffffffff
```
ArangoGutierrez, line 26: IBA PortCounters saturate at all-ones instead of rolling over, but this mask wraps: at 400Gbps, port_xmit_data covers 2^32 words in roughly half a second, so every legacy perfquery read is effectively a random sample rather than a counter pinned at 0xFFFFFFFF like real hardware. Clamping with min(raw, 0xffffffff) for Width==32 would match the spec; the narrow PMA fields (16-bit SymbolErrorCounter etc.) have the same truncate-vs-saturate question, though at these error rates that only bites after weeks.

pkg/network/mockib/counters/generator.go
```diff
@@ -0,0 +1,117 @@
...
+		// 70% utilization; sysfs port_xmit_data is in 4-byte words per IB spec.
+		return uint64(bps * secs * 0.7 / 4.0)
+	case FamilyRcvData:
+		return uint64(bps * secs * 0.65 / 4.0)
+	case FamilyXmitPackets:
+		return uint64(bps * secs * 0.7 / 4096.0)
+	case FamilyRcvPackets:
+		return uint64(bps * secs * 0.65 / 4096.0)
+	case FamilyError:
+		return uint64(secs * 0.05)
```
ArangoGutierrez, line 96: Question on the default: the error families accrue at 0.05/s (~4.3k/day) on every port, so any link-quality alert wired to symbol_error will fire permanently on a mock fabric. If the point is exercising alert pipelines, that's a feature; if the baseline should look like a healthy fabric, error families near 0 would be less surprising — either way a README line stating the choice would help.

pkg/network/mockib/counters/spec.go
```diff
@@ -0,0 +1,111 @@
...
+	Family  Family
+}
+
+// Catalog is the single source of truth for which counters exist and
+// where. The renderer iterates it to lay down initial files, the daemon
+// writer iterates it to rewrite them, and pma.go indexes into it by
+// name. Keep entries deterministically ordered (slice, not map) so file
+// creation order is stable across runs.
+var Catalog = []Entry{
+	// counters/ legacy (rendered today as zero)
```
ArangoGutierrez, line 48: nit: stale once this PR lands — these render non-zero now.

ArangoGutierrez (commented):
Early pass on the WIP, mostly to close the loop on the red CI my last look didn't cover. The e2e-device-plugin a100/h100 failures aren't in your new code: this branch sits on a pre-rebase snapshot of #367 that's missing its later fixes, most relevantly af621a54 (fail the NETLINK_RDMA socket so libibverbs falls back to sysfs) — the runner's real Azure mlx5 NIC leaks into enumeration, validate-ibv-devinfo sees 1 device instead of 2, and the new counters/perfquery validators get skipped, so they've never run in CI yet. The same staleness silently reverts other #367 hardening here: the short-MAD guard in handleSend, the conn read/write deadlines, the atomic register counter, and the widened GUID/LID encoding that nodeid/ narrows back to 16 bits. So before this leaves draft: rebase onto current #367 (or main once it lands), which is also the moment to fix DCO — the 19 counters/PMA commits are missing Signed-off-by. The new code itself reads well; inline: a PortSelect wire-offset mismatch vs libibmad, a perfquery timeout regression on the counters-disabled path, 32-bit wrap-vs-saturate semantics, and two small nits.

Issue #371, opened 2026-05-29
feat(nvml-mock): NVSwitch + nvidia-fabricmanager simulation


giuliocalzo:
**Priority:** High

**Context.** NVLink fabric is currently faked only at the *identity* level (clique/cluster UUID via ComputeDomain #337/#342).

**Gap.** No `nvidia-fabricmanager` daemon, no NVSwitch device modeling, no NVLink per-link state/counters, no `nvidia-smi nvlink` / `nvidia-smi topo -m` output.

**Proposed scope.**
- NVML NVLink APIs (link state, capabilities, BW/error counters) backed by profiles.
- `nvidia-smi topo -m` matrix derived from the PCIe/NVLink topology already rendered (#353).
- Optional fake `nvidia-fabricmanager` readiness analogous to fake IMEX.
- E2E: assert `nvidia-smi topo -m` shows NVLink links and nvlink counters grow.

**Why.** Closes the data-plane half of the NVLink story whose identity half already shipped.

giuliocalzo:
example of output from a GB200
```
nvidia-smi topo -m
        GPU0    GPU1    GPU2    GPU3    CPU Affinity    NUMA Affinity   GPU NUMA ID
GPU0     X      NV18    NV18    NV18    0-71    0               N/A
GPU1    NV18     X      NV18    NV18    0-71    0               N/A
GPU2    NV18    NV18     X      NV18    72-143  1               N/A
GPU3    NV18    NV18    NV18     X      72-143  1               N/A

Legend:

  X    = Self
  SYS  = Connection traversing PCIe as well as the SMP interconnect between NUMA nodes (e.g., QPI/UPI)
  NODE = Connection traversing PCIe as well as the interconnect between PCIe Host Bridges within a NUMA node
  PHB  = Connection traversing PCIe as well as a PCIe Host Bridge (typically the CPU)
  PXB  = Connection traversing multiple PCIe bridges (without traversing the PCIe Host Bridge)
  PIX  = Connection traversing at most a single PCIe bridge
  NV#  = Connection traversing a bonded set of # NVLinks
```

giuliocalzo:
```
nvidia-smi nvlink -s
GPU 0: NVIDIA GB200 (UUID: GPU-d1b93026-6d5b-03bd-3a71-f85b47a61bde)
         Link 0: 53.125 GB/s
         Link 1: 53.125 GB/s
         Link 2: 53.125 GB/s
[...]

Issue #255, opened 2026-03-13
tracking: simulation fidelity enhancements


ArangoGutierrez:
## Overview

Advanced simulation capabilities inspired by nvml-mock. These are lower priority — no current consumer requires them, but they improve realism for monitoring, resilience testing, and detection tool compatibility.

## Sub-issues

- [ ] #245 — GPU failure injection (device lost, Xid errors)
- [ ] #246 — External metric sources (file/Redis replay)
- [ ] #247 — BSS padding to match real .so size

## Priority

Low — implement on-demand when a consumer or testing scenario requires it.

PR #387, opened 2026-06-11
feat(nvml-mock): simulate NVSwitch topology and nvidia-fabricmanager

giuliocalzo:
## Summary

Adds NVSwitch / NVLink fabric realism and a fake `nvidia-fabricmanager` to the `nvml-mock` stack so `nvidia-smi topo -m` and `nvidia-smi nvlink` produce realistic output on mock GPU nodes. Implements #371.

- **Engine**
  - NVLink switch-link **auto-expansion**: profiles declaring NVSwitches + `links_per_gpu` get a fully populated NV# matrix (GB200/GB300/H100 -> NV18, A100 -> NV12) without hand-authoring `N*links_per_gpu` entries.
  - Opt-in `fabric.state: auto` coupling: GPU registration state tracks the fake fabricmanager readiness marker via a TTL-cached provider; defaults to `COMPLETED` when fabricmanager is disabled (no ComputeDomain regression).
  - Wall-clock NVLink utilization counters anchored to `/proc/stat` btime so they grow monotonically across separate `nvidia-smi` processes.
- **Fake nvidia-fabricmanager**
  - `pkg/fmcoord` (node-local readiness marker coordination) and `cmd/fake-fabricmanager/{daemon,ctl}`, modeled on `fake-imex`. Daemon writes the marker (optional init delay), clears it on `SIGTERM`; `ctl -q` probes readiness.
- **Profiles**: GB200/GB300/H100 = NVSwitch + fabricmanager(auto) + NV18 @ ~53 GB/s; A100 = NVSwitch, no fabricmanager, NV12; standalone B200 = no NVSwitch/fabricmanager (negative control).
- **Deployment**: `Dockerfile` builds the FM binaries; `setup.sh` starts the daemon and injects the fabric-state mount + `MOCK_FABRICMANAGER_STATE_DIR` into the CDI spec so CDI-injected workloads honor the readiness gate; Helm `values.yaml` + daemonset gate it behind `fabricmanager.enabled`.

### GB200 `nvidia-smi nvlink` alignment with real 580 hardware

The GB200 profile + bundled `nvidia-smi` (pinned to 580.65.06) were validated line-for-line against a dump from a real GB200 (driver 580) and brought into agreement:

- **`--version`** renders `NVML version: 580.65` (`nvml_version: "13.580.65"`; `nvidia-smi` displays `components[1].components[2]`).
- **`nvlink -s`** shows `53.125 GB/s` per link via a new precise `bandwidth_per_link_mbps` config field (whole-GB/s couldn't express `.125`).
- **`nvlink -p` / `-R`**: switch-attached links return the `FFFFFFFF:FF:FF.0` "invalid PCI" sentinel and remote NVLink id `0`, matching hardware (`-R`: `Remote Device FFFFFFFF:FF:FF.0: Link 0`). Implemented via `nvmlDeviceGetNvLinkRemotePciInfo` + field `NVML_FI_DEV_NVLINK_REMOTE_NVLINK_ID` (146).
- **`nvlink -gLowPwrInfo`** reports `units of 50 us`, `range 1 to 1023`, `High Speed State`, `50000 us` via the low-power threshold field values (167/168/200/223/224/225).
- Device name reports `NVIDIA GB200`.
- **Known gaps (intentional)**: `nvlink -info` and `-gLWidth` resolve only through `nvmlInternalGetExportTable` / the opaque `nvmlNvLinkInfo_t` struct, whose ABI is not in the vendored headers — implementing them by guessing memory layout risks crashing `nvidia-smi`, so they remain "Not Supported".

The bundled `nvidia-smi` version is independent of the per-profile mocked driver version: a 580 binary renders older profiles (A100=550, B200=560, GB300=570) with no driver/library mismatch, so only GB200 (the profile validated against real 580 hardware) was bumped.

### Driver-version single source of truth

The Helm `DRIVER_VERSION` env (which names the on-disk `libnvidia-ml.so.<ver>` in `setup.sh`) now derives from each profile's `system.driver_version` instead of re-hardcoding it, so it can no longer drift from what NVML reports. `.Values.driverVersion` still overrides.

## Test plan

- [x] `go test ./... -race` (engine: golden NV# matrix per profile, auto-coupling, counter monotonicity, fmcoord lifecycle + drift-guard contract, new remote-id / low-power field values + switch PCI sentinel)
- [x] `go vet` / `golangci-lint`
- [x] `helm unittest` (snapshots regenerated)
- [x] Manual: rebuilt image, diffed full `nvidia-smi nvlink` (all flags) + `topo -m` against a real GB200 580 dump
- [ ] CI `nvml-mock-e2e.yaml` across a100/h100/b200/gb200/gb300 — `tests/e2e/validate-nvlink.sh` asserts exact per-profile NV# and fabricmanager readiness gating (first real-driver exercise)


deployments/nvml-mock/helm/nvml-mock/profiles/gb300.yaml
```diff
@@ -417,21 +417,25 @@ devices:
 nvlink:
   version: 5
   links_per_gpu: 18
-  bandwidth_per_link_gbps: 100      # 100 GB/s per link = 1.8 TB/s total per GPU
+  bandwidth_per_link_gbps: 53       # NVLink5 ~53.125 GB/s/link unidirectional (int); 18 links ~= 0.95 TB/s/GPU (1.8 TB/s bidir)
```
ArangoGutierrez: gb200 got the precise `bandwidth_per_link_mbps: 53125`, but gb300 stays whole-GB/s, so `nvlink -s` shows 53 here vs 53.125 on gb200 for the same NVLink5 rate. Worth adding the mbps field here too (and in the pkg/gpu/mocknvml/configs mirror), even though only gb200 was validated against hardware.

deployments/nvml-mock/scripts/setup.sh
```diff
@@ -267,4 +306,32 @@ if [ -x /usr/local/bin/render-pci-sysfs ]; then
...
+#     before its own daemon is ready. Clear it here so every pod starts in a
+#     clean IN_PROGRESS state until *this* daemon writes the marker.
+if [ "$MOCK_FM_MODE" != "off" ]; then
+  if [ -x /usr/bin/nv-fabricmanager ]; then
+    mkdir -p "$FM_STATE_DIR"
+    rm -f "$FM_STATE_DIR/fabricmanager.ready"
+    echo "Starting fake nvidia-fabricmanager (state dir: $FM_STATE_DIR)"
+    /usr/bin/nv-fabricmanager &
+  else
+    echo "WARNING: MOCK_FABRICMANAGER='$MOCK_FABRICMANAGER' set but /usr/bin/nv-fabricmanager not found" >&2
```
ArangoGutierrez: I'd exit 1 here instead of warning: if we reach this branch the env is wired, so auto profiles sit at IN_PROGRESS forever, which is painful to debug from the workload side. The typo branch above already fails hard; an image missing the binary is just as broken.

deployments/nvml-mock/scripts/setup.sh
```diff
@@ -267,4 +306,32 @@ if [ -x /usr/local/bin/render-pci-sysfs ]; then
...
+#     MOCK_FM_MODE / FM_STATE_DIR were resolved + validated earlier (near the
+#     CDI block). The readiness marker lives on a DirectoryOrCreate hostPath
+#     that survives pod restarts, and the daemon re-asserts it every 2s — so a
+#     stale marker from a prior pod could make a fresh pod report COMPLETED
+#     before its own daemon is ready. Clear it here so every pod starts in a
+#     clean IN_PROGRESS state until *this* daemon writes the marker.
+if [ "$MOCK_FM_MODE" != "off" ]; then
+  if [ -x /usr/bin/nv-fabricmanager ]; then
+    mkdir -p "$FM_STATE_DIR"
+    rm -f "$FM_STATE_DIR/fabricmanager.ready"
```
ArangoGutierrez, line 334: The engine/fmcoord contract test pins this filename on the Go side, but this shell literal isn't covered, so a rename would silently skip the stale-marker cleanup. A pointer at fmcoord.ReadyMarker in the comment would keep them traveling together.

pkg/gpu/mocknvml/engine/fabric_readiness.go
```diff
@@ -0,0 +1,115 @@
...
+
+func fabricStateDir() string {
+	fabricEnvDirOnce.Do(func() {
+		fabricEnvDir = strings.TrimSpace(os.Getenv(EnvFabricStateDir))
+	})
+	return fabricEnvDir
+}
+
+func (c *fabricReadinessCache) couplingStateDir() string {
+	if c == fabricReadiness {
```
ArangoGutierrez: The `c == fabricReadiness` identity check took me a second to parse. Capturing the resolved dir (or a dir func) on the struct at construction would make the singleton/test split explicit and drop the special case. Take it or leave it.

pkg/gpu/mocknvml/engine/topology.go
```diff
@@ -0,0 +1,712 @@
...
+	DefaultCoresPerNUMA = 64
+
+	// DefaultNVLinkDutyCycle is the fraction of line rate accrued into the
+	// utilization counters when the profile does not set one. A small
+	// positive value makes counters visibly grow between samples.
+	DefaultNVLinkDutyCycle = 0.05
+
+	// nvLinkMaxLinks bounds the per-device link index. Indices outside
+	// [0, nvLinkMaxLinks) are rejected as invalid arguments.
+	nvLinkMaxLinks = 18
```
ArangoGutierrez, line 119: Global cap: a100 has 12 links, so 12-17 come back as valid-but-DISABLED where real NVML returns INVALID_ARGUMENT for link >= links_per_gpu. Fine for current consumers; might be worth a line in the #255 fidelity backlog.

ArangoGutierrez (dismissed):
Strong PR. NodeFabric as one immutable node-level model with every getter derived from it is the right shape, and the tests genuinely guard it: full NV# matrix per shipped profile, cross-process counter growth, the engine/fmcoord contract test, and the b200 negative control wired through e2e. The SIGPIPE/pipefail notes in the validate scripts and the stale-marker cleanup in setup.sh are traps that usually bite much later; nice catches.

I checked that buildFabricInfo resolves "auto" per call (GetMockFabricInfo/V), so long-running consumers like the device plugin do see IN_PROGRESS -> COMPLETED, with the TTL cache keeping stat() off the hot path. Full e2e matrix is green including the new gb200/gb300/b200 jobs.

A few small comments inline. The two I'd like before merge: gb300 mbps precision and failing hard in setup.sh when the daemon binary is missing. The rest are optional.

giuliocalzo:
Thanks for the thorough review, @ArangoGutierrez! Addressed all five points in `ee181498`:

**Requested before merge**
- **gb300 mbps precision** — added `bandwidth_per_link_mbps: 53125` to `profiles/gb300.yaml` and the `pkg/gpu/mocknvml/configs` mirror, so `nvlink -s` now reports 53.125 GB/s like gb200 instead of a whole-GB/s 53.
- **setup.sh hard-fail** — the missing-binary branch now `exit 1`s (was a warning). If the env is wired but `nv-fabricmanager` is absent, auto-state GPUs would otherwise sit at IN_PROGRESS forever; a missing binary is a broken image, same as the unknown-mode branch.

**Optional (all taken)**
- **setup.sh:329** — the `fabricmanager.ready` literal now carries a comment pointing at `fmcoord.ReadyMarker` / `engine.FabricReadyMarker`, noting the shell path isn't covered by the contract test so a rename must update it too.
- **fabric_readiness.go:75** — captured the coupling-dir resolver on the cache struct as a `stateDir func() string`: the singleton uses the once-cached `fabricStateDir`, tests inject the live `envStateDir`. The `c == fabricReadiness` identity special case is gone.
- **topology.go:111** — documented the global `nvLinkMaxLinks` cap vs per-profile `links_per_gpu` as a `#255` fidelity note (a100 links 12–17 resolve valid-but-DISABLED vs real NVML's `INVALID_ARGUMENT`).

Verified: `go build`/`go test` for `engine` + `fmcoord`, `gofmt` clean, `helm unittest` 120 passed / 15 snapshots (regenerated the gb300 configmap snapshot for the new mbps line), `bash -n setup.sh`.

pkg/gpu/mocknvml/engine/device.go
```diff
@@ -1012,77 +1016,312 @@ func (d *ConfigurableDevice) GetGpuOperationMode() (nvml.GpuOperationMode, nvml.
...
+				setInvalidRemotePci(&pci)
+				debugLog("[NVML] nvmlDeviceGetNvLinkRemotePciInfo(link=%d) -> switch sentinel\n", link)
 				return pci, nvml.SUCCESS
 			}
+			if domain, bus, device, _, err := ParsePCIBusID(l.RemoteBDF); err == nil {
+				pci.Domain = domain
+				pci.Bus = bus
+				pci.Device = device
+				for i := 0; i < len(l.RemoteBDF) && i < 32; i++ {
+					pci.BusId[i] = uint8(l.RemoteBDF[i])
```
ArangoGutierrez: Rebase heads-up: I just merged go-nvml v0.13.1-0 in #400, which changed nvml.PciInfo.BusId to [32]int8. After you rebase onto current main, this cast and the invalidRemotePciBusID one a few lines down (~L1261), plus the existing site up at ~L215, will need int8(...) instead of uint8(...). Bus-ID bytes are ASCII so the conversion is lossless. Just flagging so the rebase CI doesn't surprise you — nothing to change before then.

ArangoGutierrez (commented):
Round two — all five points from the last pass are handled, and handled well. The one I'd call out is the fabric_readiness refactor: carrying stateDir as an injected resolver on the cache (singleton uses the once-cached dir, tests inject the live env) reads much better than the old `c == fabricReadiness` identity check, and it lines up with the existing now/stat seams. gb300 _mbps, the setup.sh hard-fail, the ReadyMarker sync note, and the #255 topology fidelity note all look right.

One heads-up before merge — not a problem with the PR as it stands today. main just took go-nvml v0.13.1-0 (#400), which regenerated the bindings and flipped nvml.PciInfo.BusId from [32]uint8 to [32]int8. When you rebase, the uint8 casts into BusId won't type-check anymore (golang/check will fail). I hit the exact same thing in #400 — trivial fix, uint8 -> int8, bus-ID bytes are ASCII so it's lossless. Marked the spot inline.

giuliocalzo:
Rebased onto current `main` (picks up go-nvml v0.13.1-0 from #400). Thanks for the heads-up @ArangoGutierrez — that saved time.

**The `BusId` `int8` flip** (as you flagged): fixed the three casts in `pkg/gpu/mocknvml/engine/device.go` (`~L215`, `GetNvLinkRemotePciInfo`, and `setInvalidRemotePci` `~L1261`) plus the `busIDString` test helper, all `uint8(...)` -> `int8(...)`. Folded into the commits that introduced each cast so every commit still type-checks.

**One more thing #400 changed** that the type-check alone didn't catch: v0.13.1-0 regenerated the NVML 13.0 bindings and added six new device functions (`GetHostname_v1`, `SetHostname_v1`, `ReadPRMCounters_v1`, `SetRusdSettings_v1`, `GetUnrepairableMemoryFlag_v1`, `WorkloadPowerProfileUpdateProfiles_v1`). Since `cmd/generate-bridge` keys its stub set off the vendored `nvml.h`, the committed `stubs_generated.go` no longer covered them — the `c-shared` `libnvidia-ml.so` still built, but the missing exports left them as undefined symbols, so `nvidia-smi`'s `dlopen(RTLD_NOW)` failed with *"couldn't find libnvidia-ml.so"* and every `validate-nvidia-smi` job failed at the first `nvidia-smi` call.

Fix (`55b341e8`): regenerated `stubs_generated.go` (273 -> 279 stubs) and added the five new pointer-only types to `nvml_types.h` as opaque forward declarations (ABI-safe — the stubs only ever take them by pointer and return `NOT_SUPPORTED`). Verified the rebuilt lib defines all six exports with no undefined NVML symbols (`nm`), gofmt clean, `go test ./pkg/gpu/mocknvml/...` green. Watching CI now.

deployments/nvml-mock/scripts/setup.sh
```diff
@@ -71,6 +71,19 @@ mknod -m 666 "$DEV_ROOT/nvidia-uvm-tools" c 510 1 2>/dev/null || true
...
 mkdir -p "$CDI_DIR"
 
+# Resolve fabricmanager enablement once, here, because it influences both the
+# CDI spec (below) and the daemon launch (step 11). Validate early so a typo
+# fails the pod with a clear message rather than silently disabling the gate.
+MOCK_FM_MODE=$(printf '%s' "${MOCK_FABRICMANAGER:-off}" | tr '[:upper:]' '[:lower:]')
+case "$MOCK_FM_MODE" in
+  off | on | full | "true" | 1) ;;
+  *)
+    echo "ERROR: MOCK_FABRICMANAGER='$MOCK_FABRICMANAGER' is invalid; expected off or on" >&2
```
ArangoGutierrez: The `case` accepts `off | on | full | true | 1`, but the error message only mentions `off or on`. The chart only ever sets `on` so it's harmless — just noting the message understates what's accepted (or the case could narrow to `off`/`on`).

pkg/gpu/mocknvml/engine/config_types.go
```diff
@@ -595,13 +608,65 @@ type TopologyClique struct {
...
-	Links                []NVLinkLinkConfig `json:"links,omitempty"`
+	Version              int `json:"version,omitempty"`
+	LinksPerGPU          int `json:"links_per_gpu,omitempty"`
+	BandwidthPerLinkGBPS int `json:"bandwidth_per_link_gbps,omitempty"`
+	// BandwidthPerLinkMbps, when > 0, sets the per-link speed precisely in
+	// Mbps (what NVML/`nvidia-smi nvlink -s` reports, GB/s = Mbps/1000). It
+	// overrides BandwidthPerLinkGBPS, which can only express whole GB/s — e.g.
+	// NVLink5 is 53.125 GB/s, i.e. 53125 Mbps, not 53.
+	BandwidthPerLinkMbps int  `json:"bandwidth_per_link_mbps,omitempty"`
+	SwitchSupport        bool `json:"switch_support,omitempty"`
```
ArangoGutierrez: Minor: `SwitchSupport` isn't read by the engine — auto-expansion keys off the `switches:` list. A profile that sets `switch_support: true` without a `switches:` entry gets no NV#, which could surprise someone editing a profile. A one-line doc note on the field would cover it.

ArangoGutierrez (commented):
Read through the whole change. The `NodeFabric` model is a clean way to derive every NVLink/topology/affinity surface from one immutable source, so the matrix can't drift between getters, and the switch-link auto-expansion avoids hand-authoring 8x18 link entries per profile. The tests carry their weight — the full NV# matrix per profile, the standalone B200 negative control, the cross-process counter-growth regression, and the fmcoord contract test. `BusId` is correctly `int8` after the go-nvml 0.13.1 bump, and the NVML 13.0 exports look right. Two small non-blocking notes inline. Nice work.

giuliocalzo:
Pushed `3feba471` addressing the latest review comments:

- **Dropped `NVLinkConfig.SwitchSupport`/`SwitchCount`.** They were descriptive-only — the engine derives NVSwitch fan-out solely from the `switches:` list — so `switch_support: true` without a `switches:` entry was misleading. Removed the orphaned keys from all profiles, the `pkg/.../configs` mirror, and docs, and regenerated the Helm snapshots.
- **Narrowed `MOCK_FABRICMANAGER` to `off|on`** in `setup.sh` to match the error message.

`go build`/`go test ./pkg/gpu/mocknvml/engine/` and `helm unittest` (120 tests, 15 snapshots) all pass.

ArangoGutierrez (approved):
Approving. The two inline notes are polish, not blockers — happy to see them in this PR or a follow-up. CI is fully green including the per-profile e2e, and the NodeFabric model plus test coverage hold up well. Thanks Giulio.

giuliocalzo:
@ArangoGutierrez I've polish already based on your suggestions, I'm already working to a proper e2e testing framework to be used locally and by the ci/cd

Issue #373, opened 2026-05-29
feat(mockib): OpenSM / UFM-style fabric manager view


giuliocalzo:
**Priority:** Medium

**Context.** `mock-ib` synthesizes per-node SMP/PMA responses for DR walks, but there is no subnet manager.

**Gap.** No SM/UFM: `sminfo`, full-fabric `ibnetdiscover`, SA query surface beyond the minimal PathRecord used for GUID ping.

**Scope.**
- A mock SM identity + `sminfo`, and `ibnetdiscover` producing a consistent whole-fabric topology from the registry/fabric graph.
- E2E: `ibnetdiscover` lists all peer HCAs/switches; `sminfo` returns a master SM.

**Out of scope (moved to a follow-up).**
- ~~(Stretch) minimal UFM REST telemetry endpoint.~~ → split out to #402.

**Why.** Rounds out the IB diagnostics already mocked in #367/#368.

---
_Implementation in progress in #401 (SMInfo + `ibnetdiscover`). The UFM REST telemetry endpoint is tracked separately in #402._

giuliocalzo:
```
# ibswitches 
Switch  : 0xfc6a1c030082db00 ports 65 "MF0;i28-compleaf-3.pod19.example.com:MKM9600/U1" enhanced port 0 lid 2 lmc 0

# sminfo 
sminfo: sm lid 2 sm guid 0xfc6a1c030082db00, activity count 1671874 priority 0 state 3 SMINFO_MASTER
```

PR #401, opened 2026-06-15
feat(mockib): mock subnet manager for sminfo and ibnetdiscover

giuliocalzo:
## Summary

Adds a mock subnet-manager identity so `sminfo` works, and validates whole-fabric `ibnetdiscover`, rounding out the IB diagnostics mocked in #367/#368. Implements the must-have scope of #373; the UFM REST endpoint and a synthetic switch for >2-node enumeration are deferred to follow-ups.

- **`sminfo`** — synthesize the `SMInfo` SMP attribute (`0x0020`). The daemon answers with a master SM (`SMState=MASTER`) whose GUID is the lowest-`PortGUID` port in the merged local+peer `fabric.Graph` (`(*Graph).MasterSM`), so every pod converges on one consistent SM identity. SMInfo is dispatched early in `subnet.TrySynthesize`, before directed-route target resolution, because it is LID-routed to the advertised SM LID (`1`), which is not a real graph port. GUID/SM_Key are plain big-endian; ActCount/Priority/SMState use the existing libibmad BITSOFFS helpers. SubnGet only.
- **`ibnetdiscover`** — reuses the existing directed-route `NODE_INFO`/`NODE_DESC`/`PORT_INFO` synthesis (the same walk `iblinkinfo` uses); lists the local CA and its cross-pod peer(s).
- **No deployment changes** — both tools ride the existing `libibmockumad` UMAD path; `infiniband-diags` is already installed and the shim already preloaded. Pure Go-side + tests + e2e scripts.

### Known limitation
The mock fabric is point-to-point with no switches: each one-port CA selects a single outbound neighbor and CAs do not forward directed-route SMPs, so a scan from one pod reaches the local CA plus one neighbor — sufficient for the two-node e2e, but it does not enumerate every port of a 3+ node fabric. True whole-fabric enumeration would need a synthetic switch (NodeType=2 + `SwitchInfo 0x0012`), left as a follow-up. This is documented in the README.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./pkg/network/mockib/...`
- [x] `gofmt -l` clean
- [x] `golangci-lint run pkg/network/mockib/...` (0 issues)
- [x] `go test -race ./pkg/network/mockib/...`
  - `TestFillSMInfo_WireBytes` (raw byte-pinned layout), `TestTrySynthesize_SMInfo` + SubnSet/empty-graph negatives, `TestGraph_MasterSM`, `TestServer_SMInfoMaster` (socket round-trip)
- [ ] E2E (CI, multinode job): `validate-sminfo.sh` asserts `SMINFO_MASTER` + non-zero GUID and that both pods agree on a single master SM GUID; `validate-ibnetdiscover.sh` asserts ≥1 cross-pod CA. Both skip IB-disabled profiles (l40s/t4) and `MOCK_IB=sysfs`.

Refs #373

### DIFF
--- a/tests/e2e/validate-ibnetdiscover.sh
+++ b/tests/e2e/validate-ibnetdiscover.sh
@@ -58,15 +58,22 @@ for P in "$LOCAL_POD" "$PEER_POD"; do
   fi
 done
 
-# Local pod's own port GUIDs come from the rendered mock sysfs tree. Any GUID in
+# Local pod's own GUIDs come from the rendered mock sysfs tree. Any GUID in
 # ibnetdiscover's output that is not in this set is, by construction, a peer-pod
-# port (same technique as validate-iblinkinfo.sh).
+# GUID (same technique as validate-iblinkinfo.sh). We collect port_guid AND the
+# CA-level node_guid / sys_image_guid: ibnetdiscover prints node GUIDs (caguid=)
+# in addition to port GUIDs, and the mock derives node_guid from port_guid by
+# clearing the EUI-64 U/L bit (fabric.nodeGUIDFromPortGUID), so node_guid !=
+# port_guid. Excluding only port_guid would let the local CA's own node GUID be
+# misclassified as a cross-pod peer and false-pass the assertion below.
 LOCAL_GUIDS=$(kubectl exec "$LOCAL_POD" -- sh -c '
-  for f in /var/lib/nvml-mock/ib/sys/class/infiniband/*/ports/1/port_guid; do
-    [ -r "$f" ] && cat "$f"
+  for ca in /var/lib/nvml-mock/ib/sys/class/infiniband/*; do
+    for f in "$ca/ports/1/port_guid" "$ca/node_guid" "$ca/sys_image_guid"; do
+      [ -r "$f" ] && cat "$f"
+    done
   done' 2>/dev/null | tr -d ':[:space:]' | tr 'A-F' 'a-f' | sort -u)
 if [ -z "$LOCAL_GUIDS" ]; then
-  echo "FAIL: could not enumerate local port GUIDs from sysfs on $LOCAL_POD"
+  echo "FAIL: could not enumerate local GUIDs (port/node/sys_image) from sysfs on $LOCAL_POD"
   exit 1
 fi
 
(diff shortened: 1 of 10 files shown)


tests/e2e/validate-ibnetdiscover.sh
```diff
@@ -0,0 +1,141 @@
...
+echo "=== Validating ibnetdiscover local=$LOCAL_POD peer=$PEER_POD profile=$PROFILE ==="
+
+if [ "$EXPECTED" -eq 0 ]; then
+  echo "SKIP: IB disabled for profile $PROFILE"
+  exit 0
+fi
+
+# ibnetdiscover needs the mock-ib UMAD daemon; under MOCK_IB=sysfs (no socket)
+# there is nothing to answer the DR walk, so skip rather than hard-fail.
+if ! kubectl exec "$LOCAL_POD" -- test -S "${MOCK_IBPING_SOCKET}" 2>/dev/null; then
```
ArangoGutierrez: Optional nit: validate-sminfo.sh wraps this same wait in a wait_for_socket() helper. Reusing it here would keep both scripts consistent and lets you drop the redundant outer `if !` — the loop's first iteration already returns fast when the socket is present.

ArangoGutierrez (commented):
Clean PR — pure Go-side synthesis over the existing fabric.Graph, no chart or Dockerfile churn, and the README is honest about the point-to-point topology limit. Two calls I think are right: dispatching SMInfo before DR target resolution (SM LID 1 isn't a real graph port), and TestFillSMInfo_WireBytes pinning the raw big-endian offsets rather than round-tripping through GetFieldSpec — that's the correct way to test wire layout. I also checked that the top-level GET-only guard in TrySynthesize is what drops SubnSet(SMInfo), so that case is covered universally rather than per-attr. One small optional nit inline, nothing blocking — CI is green and it's mergeable on current main.

tests/e2e/validate-ibnetdiscover.sh
giuliocalzo: Done in 3f499a99 — added the same `wait_for_socket()` helper here and dropped the redundant outer `if !` block. Thanks!

PR #403, opened 2026-06-15
feat(nvml-mock): RDMA verbs data path for perftest on Kind without IB hardware

### LINKED ISSUE #374: feat(mockib): RDMA verbs data path (perftest / GPUDirect RDMA)
**Priority:** Medium

**Context.** Full `ibv_devinfo` (non-`-l`) is intentionally out of scope today because it requires faking the mlx5 uverbs ioctl surface.

**Gap.** No verbs data path: `ibv_rc_pingpong`, `perftest` (`ib_write_bw`/`ib_read_bw`), GPUDirect RDMA, `nvidia-peermem`.

**Proposed scope.**
- Evaluate feasibility of a verbs-level shim or a loopback/relayed QP path over the existing TCP fabric to support `perftest` bandwidth runs.
- Document explicitly what remains unsupported if full uverbs faking stays out of scope.
- E2E (if feasible): cross-node `ib_write_bw` reports non-zero bandwidth.

**Why.** Enables RDMA bandwidth validation; extends the fabric relay built for ibping.

giuliocalzo:
## Summary

Adds an RDMA verbs **data path** to the `mockib` subsystem so stock `perftest` 4.5 (`ib_write_bw`, `ib_read_bw`) reports **non-zero bandwidth across nodes on Kind** — no GPU or InfiniBand hardware required. Closes #374.

- **In-process libibverbs provider shim** (`pkg/network/mockib/c/rdma_shim.c`): real `verbs_context`/`abi_compat`, `query_port`/`create_qp_ex`, and the extended `ibv_wr_*` op table so stock `perftest` works without flags. MR tracking with no-wrap bounds checks; per-port LID/GID surfaced from the mock sysfs tree so the daemon can resolve routes. Gated by `MOCK_IB_RDMA=1`.
- **Daemon relay** (`protocol/verbs.go`, `daemon/verbs_fabric.go`, `server.go`, `fabric.go`): QP create/connect/destroy, attach, and chunked `verbs_op` framing under the 1 MiB frame size, reusing the existing length-prefixed JSON TCP fabric (port 18515). Egress is fire-and-forget. Real WRITE/READ/SEND bytes are delivered into the responder MR.
- **Packaging**: Dockerfile ships `libibmockrdma.so`; Helm wires `infiniband.rdma.enabled` (**on by default**) to `MOCK_IB_RDMA` + `LD_PRELOAD`.

> The reported bandwidth is a functional artifact of a JSON relay, **not** an InfiniBand measurement.

## Test plan

- [x] `go test ./pkg/network/mockib/...` — green (wire protocol + relay unit tests)
- [x] `helm unittest deployments/nvml-mock/helm/nvml-mock` — 112/112
- [x] `go vet ./pkg/network/mockib/...` — clean
- [x] **Hardware-free loopback** harness (`pkg/network/mockib/c/loopback`) — builds the production shim, moves real bytes through the relay, runs stock `ib_write_bw`, asserts non-zero BW **and** `verbs_op` traversal. Now run as the `rdma-loopback` CI job.
- [x] **Gated cross-pod E2E** (`tests/e2e/validate-rdma.sh`) — asserts `SERVER_IP != CLIENT_IP`, non-zero bandwidth, and real fabric traversal in the daemon logs.


pkg/network/mockib/c/rdma_shim.c
```diff
@@ -0,0 +1,1346 @@
...
+    if (!m) return NULL;
+    m->context = pd->context;
+    m->pd = pd;
+    m->addr = addr;
+    m->length = len;
+    static uint32_t g_key = 0x1000;
+    m->lkey = __sync_fetch_and_add(&g_key, 1);
+    m->rkey = __sync_fetch_and_add(&g_key, 1);
+    pthread_mutex_lock(&mr_mu);
+    for (int i = 0; i < MR_MAX; i++) {
```
ArangoGutierrez, line 1098: When the 256-slot MR table is full this drops the MR but still returns a valid ibv_mr*, so mr_check rejects later ops against it with no diagnostic. Fine for perftest's handful of MRs, but a debug log on table-full would save a confusing session.

ArangoGutierrez (dismissed):
Approving. The shim's bounds logic holds up — mr_check's no-wrap test is correct, and apply_write/apply_read_resp/apply_send all re-bound writes against the process's own registered MRs, so a peer can only touch memory explicitly registered. Two non-blocking notes inline. The one thing I'd like back at some point is fast unit coverage of the shim bounds: right now that rests on review plus the single gated E2E, since the loopback harness was dropped in 16e6122.

PR #405, opened 2026-06-17
refactor(nvml-mock): consolidate NVLink bandwidth to bandwidth_per_link_mbps

giuliocalzo:
## Summary

`NVLinkConfig` exposed two overlapping bandwidth fields — `bandwidth_per_link_gbps` (whole GB/s) and `bandwidth_per_link_mbps` (Mbps). The engine preferred `_mbps` and only fell back to `_gbps`, so the whole-GB/s field was dead weight (it can't express NVLink5's 53.125 GB/s) and just added noise on `gb200`/`gb300`.

This consolidates to a single canonical `bandwidth_per_link_mbps`:

- Remove `BandwidthPerLinkGBPS` from `NVLinkConfig` and simplify the bandwidth selection in `topology.go`.
- Convert the `_gbps`-only profiles to `_mbps` and drop the redundant `_gbps: 53` fallback line from `gb200`/`gb300` (they already set `_mbps: 53125`).
- Update both the Helm `profiles/` and the `pkg/gpu/mocknvml/configs/` mirror, plus `docs/configuration.md`.
- Update unit tests and regenerate Helm snapshots.

## Realistic per-link speeds

While consolidating, the `a100`/`h100`/`b200` profiles were found to use the **bidirectional-per-link marketing figure** (50/50/100 GB/s) instead of the per-link **unidirectional line rate** that `nvidia-smi nvlink -s` actually reports (the semantics the field documents and that `gb200`/`gb300` already follow). Corrected to:

| Profile | NVLink gen | Links | `bandwidth_per_link_mbps` | Aggregate (per-link × 2 × links) |
|---|---|---|---|---|
| a100  | 3.0 | 12 | `25781` (25.781 GB/s) | ~600 GB/s bidir |
| h100  | 4.0 | 18 | `26562` (26.562 GB/s) | ~900 GB/s bidir |
| b200  | 5.0 | 18 | `53125` (53.125 GB/s) | ~1.8 TB/s bidir |
| gb200 | 5.0 | 18 | `53125` (unchanged) | ~1.8 TB/s bidir |
| gb300 | 5.0 | 18 | `53125` (unchanged) | ~1.8 TB/s bidir |

These match the documented `nvidia-smi nvlink -s` values (e.g. GV100/A100 `25.781 GB/s`, H100 `26.562 GB/s`). The docs example was bumped to the NVLink4 rate accordingly.

Closes #404. Deferred from #387 to keep that PR focused.

## Test plan

- [x] `go build ./...`
[...]

PR #409, opened 2026-06-22
fix(e2e): avoid SIGPIPE false-negative in IB validation scripts

giuliocalzo:
## Summary

The InfiniBand e2e validation scripts run under `set -euo pipefail` and check for expected output with `printf '%s\n' "$VAR" | grep -q ...`. When `grep -q` finds its match it exits `0` and closes the read end of the pipe **before** `printf` finishes writing the (multi-HCA) blob, so `printf` dies with `SIGPIPE`. Under `pipefail` the pipeline then inherits `printf`'s non-zero status, flipping `if ! pipeline` into a false-negative failure even though the expected line was present.

This is timing/buffer-dependent, so it shows up as a flaky e2e run. Observed in [run 27936780842 / e2e-device-plugin (h100)](https://github.com/NVIDIA/k8s-test-infra/actions/runs/27936780842):

```
validate-ibv-devinfo.sh: line 108: printf: write error: Broken pipe
FAIL: ibstatus output missing 'phys state: 5: LinkUp'
```

(The `LinkUp` line *was* present — `grep` matched it — the broken pipe just flipped the result.)

## Fix

Replace the `printf '%s\n' "$VAR" | grep -q ...` pattern with a here-string (`grep -q ... <<<"$VAR"`), eliminating the pipe entirely so there is nothing to break. Applied to:

- `tests/e2e/validate-ibv-devinfo.sh` — `mlx5_0` GUID check and `phys state: 5: LinkUp` check
- `tests/e2e/validate-iblinkinfo.sh` — cross-pod GUID check
- `tests/e2e/validate-ibnetdiscover.sh` — cross-pod GUID check

All three scripts are `#!/bin/bash`, so here-strings are supported.

## Test plan

Reproducer confirming the behavior (5000-line input forces the race deterministically):

```
$ bash -c 'set -euo pipefail; FULL=$(printf "phys state: 5: LinkUp\n%.0s" {1..5000})
  grep -qE "phys state:[[:space:]]+5: LinkUp" <<<"$FULL" && echo "NEW rc=0"
  set +e; printf "%s\n" "$FULL" | grep -qE "phys state:[[:space:]]+5: LinkUp"; echo "OLD rc=$?"'
NEW rc=0
OLD rc=141   # SIGPIPE -> false-negative under pipefail
```

- [x] `bash -n` passes on all three scripts
- [x] Reproducer shows old pattern fails (rc=141), new pattern succeeds (rc=0)

### DIFF
--- a/tests/e2e/validate-iblinkinfo.sh
+++ b/tests/e2e/validate-iblinkinfo.sh
@@ -76,7 +76,7 @@ fi
 
 CROSS_POD=""
 for g in $FOUND_GUIDS; do
-  if ! printf '%s\n' "$LOCAL_GUIDS" | grep -qx "$g"; then
+  if ! grep -qx "$g" <<<"$LOCAL_GUIDS"; then
     CROSS_POD="$g"
     break
   fi
--- a/tests/e2e/validate-ibnetdiscover.sh
+++ b/tests/e2e/validate-ibnetdiscover.sh
@@ -111,7 +111,7 @@ attempt_discover() {
 
   cross_pod=""
   for g in $found_guids; do
-    if ! printf '%s\n' "$LOCAL_GUIDS" | grep -qx "$g"; then
+    if ! grep -qx "$g" <<<"$LOCAL_GUIDS"; then
       cross_pod="$g"
       break
     fi
--- a/tests/e2e/validate-ibv-devinfo.sh
+++ b/tests/e2e/validate-ibv-devinfo.sh
@@ -91,7 +91,7 @@ echo "PASS: ibv_devinfo -l listed $ACTUAL devices"
 DEVICES=$(kubectl exec "$POD" -- ibv_devices 2>&1)
 echo "--- ibv_devices ---"
 printf '%s\n' "$DEVICES"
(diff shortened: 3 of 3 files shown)


tests/e2e/validate-iblinkinfo.sh
```diff
@@ -76,7 +76,7 @@ fi
 
 CROSS_POD=""
 for g in $FOUND_GUIDS; do
-  if ! printf '%s\n' "$LOCAL_GUIDS" | grep -qx "$g"; then
+  if ! grep -qx "$g" <<<"$LOCAL_GUIDS"; then
```
ArangoGutierrez, line 79: This is the canonical home for the SIGPIPE fix; #405 carries a duplicate (09caf03) that I've asked to drop.

ArangoGutierrez (approved):
Approving. Correct diagnosis and the right fix — removing the pipe kills the SIGPIPE race and the here-string is equivalent input; all three scripts are bash so the here-string is fine. Heads-up: this same change also rides in #405 as commit 09caf03, so we should land one and drop the other — I've asked on #405 to drop it there and let this PR own it.

tests/e2e/validate-iblinkinfo.sh
giuliocalzo: @ArangoGutierrez can you please merge #409 I will rebase #405

Issue #411, opened 2026-06-22
make check-modules red: tests/mocknvml module has drifted pins (pre-existing)


ArangoGutierrez:
## Problem

`make check-modules` fails on `main`, independent of any recent dependency bump.

The `modules` make target runs `go mod tidy` over every `go.mod` found via `find`, including the second, separate module `tests/mocknvml/go.mod`. That module pins:

- `github.com/NVIDIA/go-nvml` v0.13.0-1
- `github.com/onsi/testify` v1.10.0

`go mod tidy` (go 1.26.x) wants to bump those to current (`go-nvml` v0.13.2-0, `testify` v1.11.1) and create a `tests/mocknvml/vendor/` directory, producing a non-empty diff → `make check-modules` exits non-zero.

## Evidence it is pre-existing (not caused by deps)

Confirmed during the dependency merge train (#410): `make check-modules` fails identically on a clean `main` with no train applied — same two files, same hunks. `git diff main -- tests/mocknvml/` against the train branch was empty; the bumps never touched that module.

## Suggested fix

Retidy the `tests/mocknvml` module: `cd tests/mocknvml && go mod tidy` (and `go mod vendor` if that module should be vendored), align its go-nvml/testify pins with the root module, and verify `make check-modules` goes green. Worth checking whether `tests/mocknvml` actually needs to be vendored or whether the make target should skip it.

ArangoGutierrez:
Closing as completed. The target was renamed `check-modules` to `modules-check`, which is why the command in the description no longer resolves, but the underlying drift is genuinely fixed rather than hidden: PR #458 (08500a09c) committed `tests/mocknvml/vendor/`. Verified by running `make modules-check` against current main on a scratch checkout: exit 0, both modules tidied, vendored and verified, and a zero-line `git status` afterwards. tests/mocknvml is still in scope of the target, not dropped by the rename. Please reopen if the second module drifts again.

PR #413, opened 2026-06-24
fix(nvml-mock): regenerate bridge stubs for go-nvml 0.13.2 + scope e2e device plugin

giuliocalzo:
Two fixes that unblock the nvml-mock e2e matrix on `main`.

## 1. Regenerate bridge stubs for go-nvml 0.13.2 symbols

Every `e2e-device-plugin`, `e2e-dra`, and `e2e-gpu-operator` job was **red on `main`** (all GPU profiles) with:

```
NVIDIA-SMI couldn't find libnvidia-ml.so library in your system.
```

even though the mock `libnvidia-ml.so.1` is present in `lib64`.

**Root cause (reproduced locally):** the go-nvml **0.13.2-0** bump (#410) added new NVML entry points, but the generated bridge stubs were **not regenerated**, leaving 8 symbols undefined in the mock `libnvidia-ml.so`:

```
nvmlDeviceGetRemappedRows_v2
nvmlDeviceGetVgpuSchedulerLog_v2 / State_v2 / SetState_v2
nvmlDeviceVgpuForceGspUnload
nvmlGpuInstanceGetVgpuSchedulerLog_v2 / State_v2 / SetState_v2
```

`nvidia-smi` `dlopen()`s the library with `RTLD_NOW`, which resolves **every** symbol eagerly, so the first undefined symbol made the whole load fatal — surfaced as the misleading *"couldn't find libnvidia-ml.so"*. `LD_PRELOAD` paths kept working (a preloaded library is reused without re-resolution), masking it. Confirmed via `LD_DEBUG`:
```
libnvidia-ml.so.1: error: symbol lookup error: undefined symbol: nvmlDeviceGetRemappedRows_v2 (fatal)
```

**Fix:** regenerate `stubs_generated.go` (`go generate ./pkg/gpu/mocknvml/bridge/`) + add opaque forward declarations for the 4 new `_v2` info structs to `nvml_types.h` (matching the #400 pattern; pointer-only through `NOT_SUPPORTED` stubs, so ABI-safe).

**Verified (local Docker build):** rebuilt lib has no undefined `nvml` symbols; `nvidia-smi -L` succeeds via embedded RUNPATH alone (no `LD_PRELOAD`/`LD_LIBRARY_PATH`), exit 0.

## 2. Scope the mock device-plugin DaemonSet to mock-GPU nodes

`e2e-multi-node` failed at *Deploy NVIDIA device plugin* (rollout stuck at "2 of 3 pods available"). The DaemonSet tolerates all taints and had no `nodeSelector`, so it also scheduled on the **control-plane** node — where the `nvml-mock` chart is never installed (only the labeled workers get it). That pod mounts an empty `/var/lib/nvml-mock` and fails NVML init with `ERROR_LIBRARY_NOT_FOUND`. Looked flaky because the pod can briefly satisfy readiness before crashing.

**Fix:** add `nodeSelector: nvidia.com/gpu.present=true` (the label `setup.sh` applies to every node the chart provisions). Excludes the multi-node control-plane while still covering single-node clusters (whose control-plane is labeled). Race-safe: the label is set during the chart's `--wait` install, which always precedes applying this manifest.

## Related

- #410 — the go-nvml 0.13.2-0 bump that introduced the new symbols.
- #411 — separate pre-existing `main` breakage in the same go-nvml-drift family (`make check-modules` / `tests/mocknvml` go.mod). Not fixed here.

## Test plan

- [ ] `e2e-device-plugin`, `e2e-dra`, `e2e-gpu-operator` (all profiles) green
- [ ] `e2e-multi-node` green (rollout 2/2, control-plane excluded)

### DIFF
--- a/pkg/gpu/mocknvml/bridge/nvml_types.h
+++ b/pkg/gpu/mocknvml/bridge/nvml_types.h
@@ -468,6 +468,17 @@ typedef struct nvmlRusdSettings_v1_st                       nvmlRusdSettings_v1_
 typedef struct nvmlUnrepairableMemoryStatus_v1_st           nvmlUnrepairableMemoryStatus_v1_t;
 typedef struct nvmlWorkloadPowerProfileUpdateProfiles_v1_st nvmlWorkloadPowerProfileUpdateProfiles_v1_t;
 
+/*
+ * NVML additions (go-nvml v0.13.2-0, #410). Versioned _v2 info structs for
+ * remapped rows and the vGPU scheduler APIs. Only ever passed by pointer
+ * through NOT_SUPPORTED stubs, so opaque forward declarations are sufficient
+ * and ABI-safe.
+ */
+typedef struct nvmlRemappedRowsInfo_v2_st                   nvmlRemappedRowsInfo_v2_t;
+typedef struct nvmlVgpuSchedulerLogInfo_v2_st               nvmlVgpuSchedulerLogInfo_v2_t;
+typedef struct nvmlVgpuSchedulerStateInfo_v2_st             nvmlVgpuSchedulerStateInfo_v2_t;
+typedef struct nvmlVgpuSchedulerState_v2_st                 nvmlVgpuSchedulerState_v2_t;
+
 #ifdef __cplusplus
 }
 #endif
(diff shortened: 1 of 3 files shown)


giuliocalzo:
@ArangoGutierrez  — all 28 checks pass on the latest run, including the full `nvml-mock-e2e` matrix (`e2e-device-plugin`, `e2e-dra`, `e2e-gpu-operator` across all profiles) and `e2e-multi-node`.

PR #432, opened 2026-07-06
feat(nvml-mock): mock-driver image and host-driver masquerade

faganihajizada:
## What this adds

Two new ways to test the **unmodified NVIDIA GPU Operator** on clusters with no GPUs, alongside the existing `driver.enabled=false` baseline (functionally unchanged; only a stale version comment in its header was corrected):

| Mode | How | What it unlocks |
|------|-----|-----------------|
| Driver disabled (existing) | nvml-mock symlinks `/run/nvidia/driver` | Operator green, devices in pods (unchanged baseline) |
| **Host driver masquerade** (new) | `--set hostDriver.enabled=true` on nvml-mock | nvidia-smi + mock libs at standard host paths (`/usr/bin`, `/usr/lib64`, ldcache); the validator takes its preinstalled host-driver branch (`IS_HOST_DRIVER=true`); zero driver-root env overrides anywhere; node-level consumers (e.g. slurmd GRES `AutoDetect=nvml`) work unconfigured |
| **Managed mock driver** (new) | `deployments/mock-driver/` image substituted via standard `driver.repository/image/version` with `driver.enabled=true` | The operator's containerized-driver lifecycle, which previously ran nowhere without physical GPUs: DaemonSet rendering and image-tag resolution, the k8s-driver-manager init flow, the startup-probe -> `.driver-ctr-ready` handshake, and the validator's operator-managed branch |

Nothing in the operator is patched or mocked; only the driver image is substituted. The mock entrypoint satisfies the operator's own contract: it fakes `/sys/module/nvidia/refcnt` and `/proc/driver/nvidia` strictly inside its private mount namespace (the host's `/proc` and `/sys` are never touched) and exposes its rootfs at `/run/nvidia/driver` with the same rbind the real entrypoint uses. Readiness sentinels are written by the operator's own probe and validator, never by the mock.

## Design notes

- **Contract vendoring + drift check**: the operator assets that define the driver-container contract (startup-probe ConfigMap, driver DaemonSet template) are vendored under `tests/e2e/contract/v26.3.3/` and re-diffed against upstream in CI (`tests/e2e/check-driver-contract.sh`), so an operator release changing the contract fails loudly instead of silently breaking the mock. The contract has changed before (v25.x inline probe -> v26.x refcnt-checking script).
- **Host mutation is manifest-tracked and guarded**: every path the masquerade writes is recorded before it is created (no crash window leaves untracked files), installs refuse to overwrite anything nvml-mock does not own (foreign nvidia-smi, libs, ldcache entries), mode switches converge stale state in both directions, and uninstall removes exactly the manifest. The e2e asserts zero host residue after `helm uninstall`.
- **Values overlays are deltas**: `gpu-operator-driver-values.yaml` and `gpu-operator-hostdriver-values.yaml` layer on the existing baseline (`-f baseline -f delta`), so mock component policy lives in one file.
- **Shared plumbing**: `lib-driver-root.sh` is the single source of the driver-root layout (used by the mock-driver entrypoint and the hostDriver install); four composite actions (`.github/actions/`) dedupe the two new e2e lanes and are not wired into the existing jobs.
- **mock-driver publishes multi-arch** (amd64 + sbsa arm64) with `-debian12` tag aliases for Kind, mirroring the hardened nvml-mock publish pipeline (SHA-pinned actions, cosign, SBOM).

## Testing

- Two new single-profile CI lanes: `e2e-gpu-operator-driver` (pinned to the vendored operator contract) and `e2e-gpu-operator-hostdriver`. The existing lanes are untouched.
- 136 helm-unittest tests + snapshots pass. Mode-transition behavior (version switches, unclean-teardown recovery, foreign-file refusal, manifest preservation) was additionally verified in privileged containers during development; those checks live in the e2e assertions where CI-appropriate.

## Docs

- `docs/mock-driver.md`: the contract table, usage, when-to-use-which-mode matrix, teardown behavior, non-goals.
- Chart README variant sections, `docs/troubleshooting.md` three-mode matrix, `tests/e2e/VERSION-MATRIX.md` coverage notes.


ArangoGutierrez:
cc @giuliocalzo

ArangoGutierrez (commented):
Found 1 issue.

1. Main bumped every Go pin to 1.26.5 in 4299fcb7 (#445) because 1.26.4 fails govulncheck with GO-2026-5856 (crypto/tls). mock-driver-publish.yaml passes no GOLANG_VERSION build-arg, so published images would build with the vulnerable toolchain; bumping this default to 1.26.5 keeps it in sync with deployments/nvml-mock/Dockerfile. (deployments/mock-driver/Dockerfile:26)

deployments/mock-driver/Dockerfile
```diff
@@ -0,0 +1,83 @@
...
+# deploys it as the driver DaemonSet. See deployments/mock-driver/scripts/
+# nvidia-driver for the contract it satisfies. Built from the same artifacts
+# as nvml-mock (mock NVML/CUDA libraries, the same nvidia-smi extraction,
+# the shared driver-root shell library) but laid out as a driver rootfs
+# (usr/lib64, usr/bin) that the entrypoint exposes at /run/nvidia/driver.
+
+# NOTE: the builder stage mirrors the mocknvml/mockcuda part of the one in
+# deployments/nvml-mock/Dockerfile -- keep them in sync (this image does not
+# need the IB/IMEX/fabricmanager components).
+ARG GOLANG_VERSION=1.26.4
```
ArangoGutierrez: Main bumped every Go pin to 1.26.5 in 4299fcb7 (#445) because 1.26.4 fails govulncheck with GO-2026-5856 (crypto/tls). mock-driver-publish.yaml passes no GOLANG_VERSION build-arg, so published images would build with the vulnerable toolchain; bumping this default to 1.26.5 keeps it in sync with deployments/nvml-mock/Dockerfile.

PR #422, opened 2026-07-01
feat(nvml-mock): node-wide mock GPU injection via mutating admission webhook

giuliocalzo:
## Summary

Adds a **node-wide mock GPU injector** so that *any* ordinary pod can be turned
into a working mock GPU node — with **no `nvidia.com/gpu` request and no mock
mounts of its own**. A mutating admission webhook (shipped with the nvml-mock
chart) overlays the host-staged mock GPU/IB/PCI stack onto admitted pods and
wires up `PATH` / `LD_LIBRARY_PATH` / `LD_PRELOAD` / `MOCK_*` so `nvidia-smi`,
`ibnetdiscover`, and friends work as if real hardware were present.

## Architecture

```mermaid
flowchart TB
  subgraph Node["Kind worker node"]
    DS["nvml-mock DaemonSet<br/>(setup.sh)"]
    Host["Host path: /var/lib/nvml-mock/driver<br/>nvidia-smi · libnvidia-ml.so.1 · libcuda.so.1<br/>IB/PCI LD_PRELOAD shims · config.yaml · topology.yaml"]
    DS -->|stages mock stack| Host
    DS -->|labels node| Lbl["nvidia.com/gpu.present=true"]
  end

  subgraph CP["Control plane"]
    API["kube-apiserver"]
    Web["nvml-mock-injector<br/>MutatingWebhookConfiguration<br/>(self-signed TLS · failurePolicy: Ignore)"]
  end

  Pod0["Plain pod CREATE<br/>no GPU request · no mounts<br/>only NODE_NAME (downward API)"] --> API
  API -->|"AdmissionReview (TLS :8443)"| Web
  Web -->|"JSONPatch: hostPath overlay mount<br/>+ PATH / LD_LIBRARY_PATH / LD_PRELOAD / MOCK_*<br/>+ annotation injected=true"| API
  API --> Pod["Admitted pod (/opt/nvml-mock overlay)"]
  Host -. hostPath overlay .-> Pod
  Pod -->|"nvidia-smi (preloaded libnvidia-ml.so.1)"| Clique["reports per-node NVLink clique<br/>resolved from topology.yaml by NODE_NAME"]
```

**Admission sequence:**

```mermaid
sequenceDiagram
  participant U as kubectl / controller
  participant API as kube-apiserver
  participant W as nvml-mock-injector
  participant K as kubelet
  U->>API: create Pod (plain, no GPU)
  API->>W: AdmissionReview (TLS :8443)
[...]

PR #433, opened 2026-07-06
feat(nvml-mock): add node-wide NRI injection demo

giuliocalzo:
## Summary
- Add `nvml-mock-nri`, an NRI plugin that injects the mock GPU overlay into ordinary containers at create time.
- Wire the NRI DaemonSet into the Helm chart and stage the host overlay, mock IB tools, PCI sysfs shim, and optional device nodes.
- Add a dedicated Kind demo with `gpu-agent` plus self-checking pods for ambient injection, opt-out, and explicit device opt-in.

## References
- Replaces #422, which used a mutating admission webhook for node-wide injection.

## Test plan
- `go test ./pkg/gpu/mocknvml/engine ./pkg/gpu/mocknvml/bridge`
- `go test ./pkg/nri/nvmlmock ./cmd/nvml-mock-nri`
- `bash -n docs/demo/node-wide-injection/run.sh`
- `kubectl apply --dry-run=client -f docs/demo/node-wide-injection/plain-pod.yaml -f docs/demo/node-wide-injection/opt-out-pod.yaml -f docs/demo/node-wide-injection/device-opt-in-pod.yaml`
- `helm template nvml-mock deployments/nvml-mock/helm/nvml-mock --namespace nvml-mock-system --set nri.enabled=true --show-only templates/nri-daemonset.yaml`
- Not completed in this agent session: full demo rerun, because Docker socket access was denied in the sandbox.


pkg/system/mockpcisysfs/c/shim.c
```diff
@@ -0,0 +1,241 @@
...
+#include <unistd.h>
+
+#ifndef PATH_MAX
+#define PATH_MAX 4096
+#endif
+
+static const char *const k_prefixes[] = {
+    "/sys/bus/pci/devices/",
+    "/sys/bus/pci/",
+    "/sys/devices/pci",
```
ArangoGutierrez, line 30: The boundary check at line 65 makes this entry dead: real paths look like /sys/devices/pci0000:00/..., so the byte after the prefix is a hex digit, never NUL or '/', and the loop skips it — those reads silently escape to the real host sysfs. Probably needs a prefix-only match for this entry; worth a shim test too, since the existing one only covers /sys/bus/pci/devices/.

deployments/nvml-mock/helm/nvml-mock/templates/daemonset.yaml
```diff
@@ -120,6 +122,8 @@ spec:
               value: {{ ternary "1" "0" $ibEnabled | quote }}
             - name: MOCK_IB_PING_PORT
               value: "{{ .Values.infiniband.ping.port }}"
+            - name: MOCK_IB_PING_SOCKET
```
ArangoGutierrez, line 125: Setting this unconditionally moves the socket, and validate-sminfo.sh and validate-ibnetdiscover.sh still expand MOCK_IB_PING_SOCKET host-side where it falls back to /run/mock-ib.sock — wait_for_socket times out and both exit 0 as SKIP, so those two e2e steps become silent no-ops. The single-quote fix you applied in validate-ibping.sh is needed in those two as well.

deployments/nvml-mock/helm/nvml-mock/templates/nri-daemonset.yaml
```diff
@@ -0,0 +1,73 @@
...
+      app.kubernetes.io/component: nri
+  updateStrategy:
+    {{- toYaml .Values.updateStrategy | nindent 4 }}
+  template:
+    metadata:
+      labels:
+        {{- include "nvml-mock.selectorLabels" . | nindent 8 }}
+        app.kubernetes.io/component: nri
+    spec:
+      serviceAccountName: {{ include "nvml-mock.fullname" . }}
```
ArangoGutierrez, line 26: This SA is bound to the ClusterRole with get/patch on nodes, but nvml-mock-nri never talks to the API server. Since this pod runs as root and adjusts every container on the node, automountServiceAccountToken: false would be good here — the demo pods already set it.

ArangoGutierrez (commented):
Really like where this is going — injecting at the runtime layer via NRI is the right direction for the mock, and the plugin and shim read well. Found 3 issues worth a look before merge.

1. The boundary check at line 65 makes this entry dead: real paths look like /sys/devices/pci0000:00/..., so the byte after the prefix is a hex digit, never NUL or '/', and the loop skips it — those reads silently escape to the real host sysfs. Probably needs a prefix-only match for this entry; worth a shim test too, since the existing one only covers /sys/bus/pci/devices/. (pkg/system/mockpcisysfs/c/shim.c:30)
2. Setting this unconditionally moves the socket, and validate-sminfo.sh and validate-ibnetdiscover.sh still expand MOCK_IB_PING_SOCKET host-side where it falls back to /run/mock-ib.sock — wait_for_socket times out and both exit 0 as SKIP, so those two e2e steps become silent no-ops. The single-quote fix you applied in validate-ibping.sh is needed in those two as well. (deployments/nvml-mock/helm/nvml-mock/templates/daemonset.yaml:125)
3. This SA is bound to the ClusterRole with get/patch on nodes, but nvml-mock-nri never talks to the API server. Since this pod runs as root and adjusts every container on the node, automountServiceAccountToken: false would be good here — the demo pods already set it. (deployments/nvml-mock/helm/nvml-mock/templates/nri-daemonset.yaml:26)

giuliocalzo:
Pushed the review fixes in `73808648`.

Addressed the three review items:
- Fixed the PCI sysfs shim so `/sys/devices/pci0000:00/...` paths redirect through `MOCK_PCI_ROOT`, and added Linux shim coverage for that path.
- Updated `validate-sminfo.sh` and `validate-ibnetdiscover.sh` so `MOCK_IB_PING_SOCKET` expands inside the pod, matching the existing `validate-ibping.sh` fix.
- Added `automountServiceAccountToken: false` to the NRI DaemonSet and covered it in the Helm unit test.

CI is fully green on the new head, including DCO, Go checks, Helm lint/unit tests, CodeQL, and the full `nvml-mock-e2e` matrix.

ArangoGutierrez (commented):
Re-checked the new commit against the earlier review: the shim prefix fix is correct — the new integration test drives exactly the path shape the old boundary check missed — the socket quoting now matches validate-ibping.sh in both remaining scripts, and automountServiceAccountToken landed with a chart test asserting it. Nothing new to flag; clean follow-up.

pkg/nri/nvmlmock/adjust.go
```diff
@@ -0,0 +1,241 @@
...
+			return true
+		}
+	}
+	return false
+}
+
+func buildEnv(cfg Config, existing []string) []string {
+	env := make(map[string]string, len(existing)+8)
+	order := make([]string, 0, len(existing)+8)
+	for _, item := range existing {
```
ArangoGutierrez, line 215: buildEnv returns the full merged environment and toNRI then AddEnvs every entry, so the plugin claims NRI ownership of env vars it never modifies — in a multi-plugin deployment another plugin touching one of those keys becomes a hard per-key conflict and the container fails to create. Emitting only the vars the plugin adds or changes avoids the over-claim.

cmd/nvml-mock-nri/main.go
```diff
@@ -0,0 +1,180 @@
...
+	for _, env := range adjustment.Env {
+		key, value, ok := strings.Cut(env, "=")
+		if !ok {
+			continue
+		}
+		result.AddEnv(key, value)
+	}
+	for _, device := range adjustment.Devices {
+		nriDevice, err := nriDevice(device)
+		if err != nil {
```
ArangoGutierrez, line 131: A stat failure here — or ReadDir when the device dir isn't staged yet — hard-fails CreateContainer for the pod. Nothing orders this DaemonSet after the main one that stages the device tree, so an opt-in pod on a freshly scheduled node fails creation outright instead of degrading to overlay-only injection.

pkg/system/mockpcisysfs/c/shim.c
```diff
@@ -0,0 +1,239 @@
...
+    if (!root || root[0] == '\0') {
+        disabled_cached = 1;
+        return;
+    }
+    root_cached = root;
+    root_len_cached = strlen(root);
+    disabled_cached = 0;
+}
+
+static int rewrite_path(const char *path, char *out, size_t out_size) {
```
ArangoGutierrez, line 52: rewrite_path returns -1 when the rewritten path would overflow, but every call site only special-cases rc == 1, so an oversized MOCK_PCI_ROOT silently falls through to the real host path — the same silent-escape outcome as the fixed prefix bug, just via a different trigger. Erroring out with ENAMETOOLONG would be safer than escaping.

docs/demo/standalone/demo.sh
```diff
@@ -95,17 +101,23 @@ helm upgrade --install nvml-mock "${REPO_ROOT}/${CHART_PATH}" \
   --set gpu.dynamicMetrics.enabled=true \
   --wait --timeout 120s
 
+# Make the demo namespace the context default so the validate-*.sh helpers
+# (which run `kubectl exec <pod>` without -n) resolve pods in it. Scoped to the
+# kind-${CLUSTER_NAME} context, which is torn down with the cluster.
+info "Setting default namespace to ${NAMESPACE} for the current context"
+kubectl config set-context --current --namespace="${NAMESPACE}"
```
ArangoGutierrez: set-context --current mutates whichever kubectl context is active, and on the cluster-reuse branch nothing has switched to the kind context first — a drifted kubeconfig gets its default namespace silently repointed, which contradicts the comment above it. Pinning the kind context explicitly before this line closes that hole.

ArangoGutierrez (commented):
Took a fuller pass across the whole non-vendor diff at this head; found 4 issues worth a look.

1. buildEnv returns the full merged environment and toNRI then AddEnvs every entry, so the plugin claims NRI ownership of env vars it never modifies — in a multi-plugin deployment another plugin touching one of those keys becomes a hard per-key conflict and the container fails to create. Emitting only the vars the plugin adds or changes avoids the over-claim. (pkg/nri/nvmlmock/adjust.go:156)
2. A stat failure here — or ReadDir when the device dir isn't staged yet — hard-fails CreateContainer for the pod. Nothing orders this DaemonSet after the main one that stages the device tree, so an opt-in pod on a freshly scheduled node fails creation outright instead of degrading to overlay-only injection. (cmd/nvml-mock-nri/main.go:128)
3. rewrite_path returns -1 when the rewritten path would overflow, but every call site only special-cases rc == 1, so an oversized MOCK_PCI_ROOT silently falls through to the real host path — the same silent-escape outcome as the fixed prefix bug, just via a different trigger. Erroring out with ENAMETOOLONG would be safer than escaping. (pkg/system/mockpcisysfs/c/shim.c:52)
4. set-context --current mutates whichever kubectl context is active, and on the cluster-reuse branch nothing has switched to the kind context first — a drifted kubeconfig gets its default namespace silently repointed, which contradicts the comment above it. Pinning the kind context explicitly before this line closes that hole. (docs/demo/standalone/demo.sh:108)

giuliocalzo:
This PR is ready for another review. The earlier review feedback has been addressed:

- **Env ownership**: the NRI plugin now emits only the environment variables it actually adds or changes, so NRI no longer claims ownership of untouched keys and won't hard-conflict with other plugins editing the same variable.
- **Device fail-open**: when the device tree is missing or a device node vanishes mid-flight, injection degrades to overlay-only (with a warning) instead of blocking container creation.
- **Shim fail-closed**: the PCIe sysfs shim now returns `ENAMETOOLONG` when a rewritten path would overflow the buffer, instead of silently escaping to the real host path.
- Added tests for all three (env over-claim, device fail-open, shim overflow) and pinned the demo scripts' `kubectl`/`helm` calls to the kind context.

Local checks are green: `go test`, `golangci-lint`, `helm unittest`, and the shim integration test (built + run in a Linux container).

pkg/gpu/mocknvml/bridge/device.go
```diff
@@ -881,6 +881,18 @@ func nvmlDeviceGetRemappedRows(nvmlDevice C.nvmlDevice_t, corrRows *C.uint, uncR
 	if ret, ok := bridgeVersionCheck("nvmlDeviceGetRemappedRows"); !ok {
 		return ret
 	}
+	return deviceGetRemappedRows(nvmlDevice, corrRows, uncRows, isPending, failureOccurred)
+}
+
+//export nvmlDeviceGetRemappedRows_v2
+func nvmlDeviceGetRemappedRows_v2(nvmlDevice C.nvmlDevice_t, corrRows *C.uint, uncRows *C.uint, isPending *C.uint, failureOccurred *C.uint) C.nvmlReturn_t {
```
ArangoGutierrez: nvmlDeviceGetRemappedRows_v2 is exported with the v1 flat-pointer parameter list, but the real v2 ABI takes (device, nvmlRemappedRowsInfo_v2_t *) — go-nvml calls it with two args, so a real caller gets three garbage pointers read and written through. The old stub used the struct signature; the shared-helper refactor should keep it.

deployments/nvml-mock/helm/nvml-mock/templates/nri-daemonset.yaml
```diff
@@ -0,0 +1,85 @@
...
+kind: DaemonSet
+metadata:
+  name: {{ include "nvml-mock.fullname" . }}-nri
+  labels:
+    {{- include "nvml-mock.labels" . | nindent 4 }}
+    app.kubernetes.io/component: nri
+spec:
+  selector:
+    matchLabels:
+      {{- include "nvml-mock.selectorLabels" . | nindent 6 }}
```
ArangoGutierrez: The main DaemonSet's selector is just name+instance, which also matches these NRI pods — the extra component label here doesn't exclude them from a broader selector. Owner refs keep steady state correct, but on orphaning events the main controller can adopt and then delete NRI pods as excess per-node duplicates. Since the existing selector is immutable, giving the NRI pods their own app.kubernetes.io/name is the clean fix.

pkg/nri/nvmlmock/adjust.go
```diff
@@ -0,0 +1,314 @@
...
+				Source:      cfg.HostOverlayPath,
+				Destination: cfg.ContainerOverlayPath,
+				Type:        "bind",
+				Options:     []string{"rbind", "ro", "nosuid", "nodev"},
+			},
+		},
+		Env: buildEnv(cfg, container.Env, topologyInjectable(cfg)),
+	}
+
+	if strings.EqualFold(container.PodAnnotations[cfg.DeviceAnnotation], "true") {
```
ArangoGutierrez, line 130: Device injection is gated purely by this pod-authored annotation, so pod authors — not just cluster operators — can pull host device nodes into any non-excluded namespace. That fits the ambient design, but a one-line callout in the README or values.yaml threat-model note would make that trust boundary explicit.

ArangoGutierrez (commented):
Found 3 issues.

1. nvmlDeviceGetRemappedRows_v2 is exported with the v1 flat-pointer parameter list, but the real v2 ABI takes (device, nvmlRemappedRowsInfo_v2_t *) — go-nvml calls it with two args, so a real caller gets three garbage pointers read and written through. The old stub used the struct signature; the shared-helper refactor should keep it. (pkg/gpu/mocknvml/bridge/device.go:888)
2. The main DaemonSet's selector is just name+instance, which also matches these NRI pods — the extra component label here doesn't exclude them from a broader selector. Owner refs keep steady state correct, but on orphaning events the main controller can adopt and then delete NRI pods as excess per-node duplicates. Since the existing selector is immutable, giving the NRI pods their own app.kubernetes.io/name is the clean fix. (deployments/nvml-mock/helm/nvml-mock/templates/nri-daemonset.yaml:16)
3. Device injection is gated purely by this pod-authored annotation, so pod authors — not just cluster operators — can pull host device nodes into any non-excluded namespace. That fits the ambient design, but a one-line callout in the README or values.yaml threat-model note would make that trust boundary explicit. (pkg/nri/nvmlmock/adjust.go:130)

giuliocalzo:
@ArangoGutierrez pushed `95604b1e` with the follow-up fixes from your latest review:

- Restored `nvmlDeviceGetRemappedRows_v2` to the real v2 ABI (`device, nvmlRemappedRowsInfo_v2_t *`) and added local layout coverage for the v2 struct.
- Gave the NRI DaemonSet its own `app.kubernetes.io/name: nvml-mock-nri` selector/template labels so it cannot be adopted by the main DaemonSet controller.
- Documented the pod-authored device annotation trust boundary in `values.yaml` and the node-wide injection README.

Verification run for this change:
- `go test ./pkg/gpu/mocknvml/bridge ./pkg/nri/nvmlmock`
- `helm unittest deployments/nvml-mock/helm/nvml-mock`
- `helm template nvml-mock deployments/nvml-mock/helm/nvml-mock --namespace nvml-mock-system --set nri.enabled=true --show-only templates/nri-daemonset.yaml`
- `git diff --check`

ArangoGutierrez:
Thanks Giulio, the follow-ups all look good. I checked the v2 struct layout against the vendored go-nvml on both your branch and current main and it matches, and the full e2e matrix is green.

One last thing before merge: we landed the go-nvml v0.13.3-1 bump (#455) this morning, so your branch now conflicts on go.mod/go.sum/vendor and the bridge files. When you rebase: keep both your nvmlRemappedRowsInfo_v2_t definition and the three new struct defs from #455 in nvml_types.h (AccountingStats_v2 / BBXTimeData_v1 / CPER_v1), regenerate stubs_generated.go with make generate rather than hand-merging it, and run golangci-lint run ./... on top of build/test since the bump deprecated the mock/dgxa100 aliases.

Ping me once it's pushed and I'll approve the CI run.

giuliocalzo:
@ArangoGutierrez Rebased onto the latest `main` to resolve the merge conflicts.

Conflict resolution notes for reviewers:
- `go.mod`: kept `main`'s newer versions (`go-nvml v0.13.3-1`, `Masterminds/semver v3.5.0`, `golang.org/x/time v0.15.0`) alongside this PR's new dependencies (`containerd/nri`, `containerd/log`, `containerd/ttrpc`, and the transitive `grpc`/`genproto` deps).
- `go.sum` and `vendor/` (incl. `vendor/modules.txt`): regenerated via `go mod tidy` + `go mod vendor`.
- `pkg/gpu/mocknvml/bridge/stubs_generated.go`: regenerated with `cmd/generate-bridge` against the bumped `go-nvml` (now 289 stubs).

Verified `go build ./...` passes after the rebase.

ArangoGutierrez (approved):
Rebase verified end to end: every commit is content-identical to the previously reviewed state except the conflict resolution in the feature commit, which correctly inherits main's go-nvml v0.13.3-1, keeps both struct sets in nvml_types.h, and drops exactly the one stub the bridge now implements (290 to 289, generator-consistent). The RemappedRows v2 layout matches the vendored header field for field, all three earlier findings remain fixed, and the full e2e matrix is green.

One non-blocking nit for whenever this file is next touched: the vGPU scheduler _v2 typedefs are now declared twice in nvml_types.h (479-482 and 503-505). Identical redeclarations, harmless, no action needed here.

Thanks for the careful rebase, Giulio. Merging.

PR #442, opened 2026-07-08
Replace nvml-mock E2E with Go harness

giuliocalzo:
## Summary
- Replace the legacy bash nvml-mock E2E workflow with a Go/Ginkgo harness under `tests/e2e/go`, while keeping shell validators and manifests separate under `tests/e2e`.
- Split the Go E2E workflow to mirror the legacy pipeline shape: `e2e`, `e2e-dra`, `e2e-gpu-operator`, and `e2e-multi-node`, with matching `make` targets for local runs.
- Port standalone, DRA, GPU Operator, multi-node, failure-injection, and validator scaffolding into labeled Go scenarios so CI and local runs can select related coverage with Ginkgo label filters.
- Keep nvcr.io-backed GFD/CUDA validator coverage disabled by default and tracked in #446; default `make e2e` excludes dedicated scenarios and runs only standalone checks.
- Cover device-plugin allocatable GPU validation through the dedicated GPU Operator path rather than the default standalone E2E path.

## Test plan
- `go test -tags=e2e ./tests/e2e/go -run 'TestDemoReleaseTargetsDedicatedNamespace|TestUseCaseLabels|TestDRAClusterNameIsDedicated|TestDRAResourceClaimManifest|TestDRADriverHelmRelease|TestValidatorGFDRequiredLabels|TestMaxIntegerLine|TestHasFailureMarker'`
- `go test -tags=e2e ./tests/e2e/go -ginkgo.dry-run -ginkgo.label-filter='!validator && !dra && !gpu-operator && !multi-node'`
- `go test -tags=e2e ./tests/e2e/go -ginkgo.dry-run -ginkgo.label-filter='gpu-operator'`
- `make -n e2e && make -n e2e-dra && make -n e2e-gpu-operator && make -n e2e-multi-node`
- `ruby -e 'require "yaml"; %w[.github/workflows/ci.yaml .github/workflows/nvml-mock-e2e-go.yaml].each { |f| YAML.load_file(f); puts "parsed #{f}" }'`
- `git diff --check`

Full Docker/Kind E2E execution is left to CI because it requires Docker daemon access and cluster lifecycle operations.


Makefile
```diff
@@ -86,3 +86,30 @@ helm-unittest:
...
+# profiles regardless of E2E_PROFILES). Those run in the normal unit-test/CI
+# path; keeping them out of `make e2e` means the output reflects only the
+# E2E_PROFILES-scoped cluster suite.
+# ---------------------------------------------------------------------------
+GINKGO ?= $(GO_CMD) run github.com/onsi/ginkgo/v2/ginkgo
+E2E_TIMEOUT ?= 90m
+E2E_GINKGO_FLAGS ?=
+
+e2e:
+	$(GINKGO) --tags=e2e -v --timeout=$(E2E_TIMEOUT) $(E2E_GINKGO_FLAGS) ./tests/e2e/go | tee e2e.log
```
ArangoGutierrez, line 123: The `| tee e2e.log` makes this recipe exit with tee's status, so a failing ginkgo run still reports success - and `make e2e` is the new workflow's only pass/fail signal. Easiest fix is `SHELL := /usr/bin/env bash` plus `.SHELLFLAGS := -o pipefail -ec` (or drop the tee and use ginkgo's own log output).

tests/e2e/go/assets/device-plugin-mock.yaml
```diff
@@ -0,0 +1,46 @@
...
+spec:
+  selector:
+    matchLabels:
+      name: nvidia-device-plugin-mock
+  template:
+    metadata:
+      labels:
+        name: nvidia-device-plugin-mock
+    spec:
+      tolerations:
```
ArangoGutierrez, line 31: This embedded copy drops the `nodeSelector: nvidia.com/gpu.present: "true"` that tests/e2e/device-plugin-mock.yaml carries. With `operator: Exists` tolerations the DaemonSet also lands on the control-plane, where the mock driver root doesn't exist, so the rollout never completes on the multi-node cluster.

deployments/nvml-mock/helm/nvml-mock/README.md
```diff
@@ -125,7 +125,7 @@ kind delete cluster --name nvml-mock-test
 
 This path uses the NVIDIA DRA (Dynamic Resource Allocation) driver to expose
 mock GPUs as ResourceSlices. DRA requires a cluster with specific feature
-gates. Tested in CI via `.github/workflows/nvml-mock-e2e.yaml` → `e2e-dra` job.
+gates. Tested in CI via `.github/workflows/nvml-mock-e2e-go.yaml`.
```
ArangoGutierrez: The new workflow has no DRA scenario - `KindDRAConfig` and the DRA assertions are embedded but never called - so this claim stops being true once the old workflow is deleted. Either wire up a DRA scenario or soften this to manual-only for now.

deployments/nvml-mock/helm/nvml-mock/README.md
```diff
@@ -46,7 +46,7 @@ use "Option B: Build from source" in the quick start sections below.
 
 This path uses the NVIDIA device plugin to expose mock GPUs as
 `nvidia.com/gpu` allocatable resources. Tested in CI via
-`.github/workflows/nvml-mock-e2e.yaml` → `e2e-device-plugin` job.
+`.github/workflows/nvml-mock-e2e-go.yaml`.
```
ArangoGutierrez: device-plugin-mock.yaml is only applied by the validator scenario, which is gated behind `E2E_RUN_NGC`, and the workflow never sets it - so CI doesn't exercise this path as written. Same suggestion as the DRA note: run it in CI or adjust the claim.

README.md
```diff
@@ -49,7 +49,7 @@ on PRs.
 
 Manual dispatch supports all 7 profiles: `a100`, `h100`, `b200`, `gb200`, `gb300`, `l40s`, `t4`.
 
-See [`.github/workflows/nvml-mock-e2e.yaml`](.github/workflows/nvml-mock-e2e.yaml) for details.
+See [`.github/workflows/nvml-mock-e2e-go.yaml`](.github/workflows/nvml-mock-e2e-go.yaml) for details.
```
ArangoGutierrez, line 55: The table above still advertises four CI-tested suites (Device Plugin, DRA, GPU Operator, Multi-Node Fleet), but the workflow this now links to runs only the standalone and failure-injection scenarios. Worth trimming the table to what actually runs, or marking which rows are pending a port.

.gitignore
```diff
@@ -5,6 +5,10 @@
 *.out
 *.test
 
+# E2E harness failure diagnostics (E2E_ARTIFACTS default)
+e2e.log
+tests/e2e/artifacts/
```
ArangoGutierrez: The harness's default `E2E_ARTIFACTS` is the relative `artifacts/e2e`, which resolves against the suite working dir - local failures write to `tests/e2e/go/artifacts/e2e/`, which this pattern doesn't match. CI is unaffected since it passes an absolute path.

ArangoGutierrez (commented):
Found 6 issues. The harness itself is in good shape - the port builds clean and the unit tests pass locally; these are mostly about the CI gate actually gating, and the docs matching what the new workflow runs.

1. The `| tee e2e.log` makes this recipe exit with tee's status, so a failing ginkgo run still reports success - and `make e2e` is the new workflow's only pass/fail signal. Easiest fix is `SHELL := /usr/bin/env bash` plus `.SHELLFLAGS := -o pipefail -ec` (or drop the tee and use ginkgo's own log output). (Makefile:115)
2. This embedded copy drops the `nodeSelector: nvidia.com/gpu.present: "true"` that tests/e2e/device-plugin-mock.yaml carries. With `operator: Exists` tolerations the DaemonSet also lands on the control-plane, where the mock driver root doesn't exist, so the rollout never completes on the multi-node cluster. (tests/e2e/go/assets/device-plugin-mock.yaml:23)
3. The new workflow has no DRA scenario - `KindDRAConfig` and the DRA assertions are embedded but never called - so this claim stops being true once the old workflow is deleted. Either wire up a DRA scenario or soften this to manual-only for now. (deployments/nvml-mock/helm/nvml-mock/README.md:128)
4. device-plugin-mock.yaml is only applied by the validator scenario, which is gated behind `E2E_RUN_NGC`, and the workflow never sets it - so CI doesn't exercise this path as written. Same suggestion as the DRA note: run it in CI or adjust the claim. (deployments/nvml-mock/helm/nvml-mock/README.md:49)
5. The table above still advertises four CI-tested suites (Device Plugin, DRA, GPU Operator, Multi-Node Fleet), but the workflow this now links to runs only the standalone and failure-injection scenarios. Worth trimming the table to what actually runs, or marking which rows are pending a port. (README.md:52)
6. The harness's default `E2E_ARTIFACTS` is the relative `artifacts/e2e`, which resolves against the suite working dir - local failures write to `tests/e2e/go/artifacts/e2e/`, which this pattern doesn't match. CI is unaffected since it passes an absolute path. (.gitignore:10)

giuliocalzo:
@ArangoGutierrez the Go/Ginkgo E2E harness now has feature parity with the legacy nvml-mock E2E workflow structure:

- standalone `e2e` coverage is the default path
- DRA, GPU Operator, and multi-node scenarios are split into dedicated workflow jobs and matching Make targets
- device-plugin allocatable GPU validation now lives under the GPU Operator path instead of the default standalone run
- artifacts are scoped under `artifacts/e2e/go`

The remaining standalone GFD/CUDA validator enablement is intentionally tracked separately in #446, since those steps depend on resolving the `nvcr.io` image/auth story. That issue will be used for the next steps.

Could you please take another look and approve/merge if this looks good?

deployments/nvml-mock/Dockerfile
```diff
@@ -67,6 +67,7 @@ COPY --from=builder /out/nv-fabricmanager-ctl /usr/bin/nv-fabricmanager-ctl
 # to confirm the topology overlay assigned the expected clique to each
 # node. See NVIDIA/k8s-test-infra#304.
 COPY --from=builder /out/check-fabric    /usr/local/bin/check-fabric
+ADD --chmod=755  https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/${TARGETARCH}/kubectl /usr/local/bin/kubectl
```
ArangoGutierrez: This swaps the old curl + sha256sum --check kubectl download for a plain ADD with no integrity check, so the image now ships an unverified binary. ADD supports --checksum=sha256:<hash> for remote sources — one line restores the control.

.github/workflows/nvml-mock-e2e-go.yaml
```diff
@@ -0,0 +1,269 @@
...
+          tags: ${{ env.E2E_IMAGE }}
+          load: true
+          build-args: |
+            GOLANG_VERSION=${{ env.GO_VERSION }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
+      - name: Install Kind
+        run: |
+          curl -fsSLo ./kind "https://kind.sigs.k8s.io/dl/${{ env.KIND_VERSION }}/kind-linux-amd64"
```
ArangoGutierrez, line 93: kind is curl'd and sudo-moved into PATH with no hash check, here and in the other three jobs; the deleted workflow got kind via the SHA-pinned kind-action. Verifying the published release checksum, or going back to the pinned action, keeps the install auditable.

.github/workflows/nvml-mock-e2e-go.yaml
```diff
@@ -0,0 +1,269 @@
...
+  workflow_dispatch:
+    inputs:
+      gpu_profiles:
+        description: 'GPU profiles to test (JSON array)'
+        required: false
+        default: '["gb200"]'
+        type: string
+
+env:
+  GO_VERSION: "1.26.5"
```
ArangoGutierrez: GO_VERSION is hardcoded and the golang_version input plumbed from hack/golang-version.sh via basic-checks is gone — that wiring existed because this exact workflow has drifted before. Worth restoring the input so go.mod bumps can't silently diverge from CI.

tests/e2e/go/framework/harness/harness.go
```diff
@@ -0,0 +1,62 @@
...
+// the given kind config, wires adapters to its Kind context in the default
+// kubeconfig, and kind-loads image. The Harness is returned even on error (with
+// whatever was created) so callers can DeferCleanup it.
+func Setup(ctx context.Context, name string, kindConfig []byte, image string) (*Harness, error) {
+	h := &Harness{Image: image}
+
+	c, err := cluster.Create(ctx, name, kindConfig)
+	if c != nil {
+		h.Cluster = c
+		if k, kerr := kube.New(c.Context); kerr == nil {
```
ArangoGutierrez: kerr is silently dropped here: h.Kube stays nil and Setup returns no error, so a failing kube.New surfaces later as a nil-pointer panic in the first scenario that touches h.Kube. Propagating kerr keeps the failure at the source.

tests/e2e/go/e2e_suite_unit_test.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	if got := rel.Set["updateStrategy.rollingUpdate.maxUnavailable"]; got != "100%" {
+		t.Fatalf("expected maxUnavailable 100%% for fast scenario rollouts, got %q", got)
+	}
+	if got := rel.Set["terminationGracePeriodSeconds"]; got != "1" {
+		t.Fatalf("expected terminationGracePeriodSeconds=1 for fast scenario rollouts, got %q", got)
+	}
+}
+
+func TestUseCaseLabels(t *testing.T) {
```
ArangoGutierrez: TestUseCaseLabels compares useCaseLabels to a verbatim copy of the same literal, and nothing else consumes the variable, so the test can't catch a real bug. Either assert something behavioral or drop it.

deployments/nvml-mock/helm/nvml-mock/README.md
```diff
@@ -125,7 +125,8 @@ kind delete cluster --name nvml-mock-test
 
 This path uses the NVIDIA DRA (Dynamic Resource Allocation) driver to expose
 mock GPUs as ResourceSlices. DRA requires a cluster with specific feature
-gates. Tested in CI via `.github/workflows/nvml-mock-e2e.yaml` → `e2e-dra` job.
+gates. Use this quick start for local/manual validation; the current Go E2E
+workflow does not yet include a DRA scenario.
```
ArangoGutierrez: This still says the Go workflow has no DRA scenario, but e2e-dra and scenario_dra_test.go landed later in this same PR — the sentence just needs updating.

README.md
```diff
@@ -36,20 +36,19 @@ After install, deploy a consumer to test:
 
 ## E2E Testing
 
-The nvml-mock E2E workflow tests all GPU consumers across multiple profiles
-and node topologies. Run manually via `workflow_dispatch` or automatically
-on PRs.
+The nvml-mock Go E2E workflow currently gates the standalone demo coverage
```
ArangoGutierrez: This section lists only standalone and failure-injection, but the workflow also runs e2e-dra, e2e-gpu-operator, and e2e-multi-node. tests/e2e/README.md has all five right; just this table is out of sync.

ArangoGutierrez (commented):
Found 7 issues.

1. This swaps the old curl + sha256sum --check kubectl download for a plain ADD with no integrity check, so the image now ships an unverified binary. ADD supports --checksum=sha256:<hash> for remote sources — one line restores the control. (deployments/nvml-mock/Dockerfile:70)
2. kind is curl'd and sudo-moved into PATH with no hash check, here and in the other three jobs; the deleted workflow got kind via the SHA-pinned kind-action. Verifying the published release checksum, or going back to the pinned action, keeps the install auditable. (.github/workflows/nvml-mock-e2e-go.yaml:85)
3. GO_VERSION is hardcoded and the golang_version input plumbed from hack/golang-version.sh via basic-checks is gone — that wiring existed because this exact workflow has drifted before. Worth restoring the input so go.mod bumps can't silently diverge from CI. (.github/workflows/nvml-mock-e2e-go.yaml:44)
4. kerr is silently dropped here: h.Kube stays nil and Setup returns no error, so a failing kube.New surfaces later as a nil-pointer panic in the first scenario that touches h.Kube. Propagating kerr keeps the failure at the source. (tests/e2e/go/framework/harness/harness.go:37)
5. TestUseCaseLabels compares useCaseLabels to a verbatim copy of the same literal, and nothing else consumes the variable, so the test can't catch a real bug. Either assert something behavioral or drop it. (tests/e2e/go/e2e_suite_unit_test.go:46)
6. This still says the Go workflow has no DRA scenario, but e2e-dra and scenario_dra_test.go landed later in this same PR — the sentence just needs updating. (deployments/nvml-mock/helm/nvml-mock/README.md:129)
7. This section lists only standalone and failure-injection, but the workflow also runs e2e-dra, e2e-gpu-operator, and e2e-multi-node. tests/e2e/README.md has all five right; just this table is out of sync. (README.md:39)

giuliocalzo:
Pushed follow-up commit `06a4ced4` addressing the latest review items:

- restored Go version plumbing from `basic-checks` into the reusable Go E2E workflow
- added checksum verification for Kind installs in each workflow job
- restored kubectl download verification in the nvml-mock image using the published `kubectl.sha256`
- propagated kube client initialization errors from harness setup instead of dropping them
- removed the self-referential `useCaseLabels` unit test/variable
- updated the stale DRA and root E2E documentation to match the current workflow jobs

Verification run locally:

- `go test -tags=e2e ./tests/e2e/go/... -run 'Test[^E]'`
- Ginkgo dry-runs for default and dedicated label filters
- `make -n e2e`, `make -n e2e-dra`, `make -n e2e-gpu-operator`, `make -n e2e-multi-node`
- YAML parse for changed workflows
- `git diff --check`
- `docker buildx build --check -f deployments/nvml-mock/Dockerfile .`

ArangoGutierrez (commented):
Found 4 issues in the changes since my last review.

1. When `kind create cluster` fails, `Create` still returns the cluster alongside its error, and `kube.New` then almost always fails too — so this early return masks the actionable create error with the generic kube-client one, in exactly the most common CI failure mode. Checking `err` first (or `errors.Join(err, kerr)`) keeps the root cause visible. (tests/e2e/go/framework/harness/harness.go:39)
2. `FirstPodName` has no Running-phase filter, and `deployNRIAgent` deletes and re-applies the DaemonSet per profile, so a terminating pod from the previous round can transiently match the selector and the later exec against it flakes. `nriAgentPodOnNode`'s jsonpath has the same exposure — the standalone scenario already uses the phase-filtered `RunningPodNames` for this. (tests/e2e/go/scenario_nri_test.go:197)
3. This section now covers DRA, GPU Operator, and multi-node — but the NRI scenario this same PR adds (the e2e-nri job, fully documented in tests/e2e/README.md) is missing from both the sentence and the table. (README.md:39)
4. The workflow_call path now derives the Go version from hack/golang-version.sh, but this manual-dispatch default is hardcoded and will silently drift on the next Go bump — that exact drift has failed the whole matrix here before. Worth a comment pointing at go.mod, or dropping the default so dispatch requires an explicit version. (.github/workflows/nvml-mock-e2e-go.yaml:48)

tests/e2e/go/framework/harness/harness.go
```diff
@@ -0,0 +1,73 @@
...
+// the given kind config, wires adapters to its Kind context in the default
+// kubeconfig, and kind-loads image. The Harness is returned even on error (with
+// whatever was created) so callers can DeferCleanup it.
+func Setup(ctx context.Context, name string, kindConfig []byte, image string) (*Harness, error) {
+	h := &Harness{Image: image}
+
+	c, err := cluster.Create(ctx, name, kindConfig)
+	if c != nil {
+		if kerr := h.attachCluster(c, kube.New); kerr != nil {
+			return h, kerr
```
ArangoGutierrez: When `kind create cluster` fails, `Create` still returns the cluster alongside its error, and `kube.New` then almost always fails too — so this early return masks the actionable create error with the generic kube-client one, in exactly the most common CI failure mode. Checking `err` first (or `errors.Join(err, kerr)`) keeps the root cause visible.

tests/e2e/go/scenario_nri_test.go
```diff
@@ -0,0 +1,268 @@
...
+		Expect(strings.ToLower(out)).To(ContainSubstring(strings.ToLower("clusterUuid : "+nriDomainUUID)),
+			"%s: expected clusterUuid %s from check-fabric\n%s", w.Name, nriDomainUUID, strings.TrimSpace(out))
+	}
+}
+
+func firstNRIAgentPod(ctx context.Context, h *harness.Harness) kube.PodRef {
+	GinkgoHelper()
+	var name string
+	Eventually(func() (string, error) {
+		n, err := h.Kube.FirstPodName(ctx, nriWorkloadNS, nriAgentSelector)
```
ArangoGutierrez: `FirstPodName` has no Running-phase filter, and `deployNRIAgent` deletes and re-applies the DaemonSet per profile, so a terminating pod from the previous round can transiently match the selector and the later exec against it flakes. `nriAgentPodOnNode`'s jsonpath has the same exposure — the standalone scenario already uses the phase-filtered `RunningPodNames` for this.

README.md
```diff
@@ -36,20 +36,22 @@ After install, deploy a consumer to test:
 
 ## E2E Testing
 
-The nvml-mock E2E workflow tests all GPU consumers across multiple profiles
-and node topologies. Run manually via `workflow_dispatch` or automatically
-on PRs.
+The nvml-mock Go E2E workflow gates standalone, DRA, GPU Operator, and
```
ArangoGutierrez: This section now covers DRA, GPU Operator, and multi-node — but the NRI scenario this same PR adds (the e2e-nri job, fully documented in tests/e2e/README.md) is missing from both the sentence and the table.

.github/workflows/nvml-mock-e2e-go.yaml
```diff
@@ -0,0 +1,339 @@
...
+    inputs:
+      gpu_profiles:
+        description: 'GPU profiles to test (JSON array)'
+        required: false
+        default: '["gb200"]'
+        type: string
+      golang_version:
+        description: 'Go version to use'
+        required: false
+        default: '1.26.5'
```
ArangoGutierrez: The workflow_call path now derives the Go version from hack/golang-version.sh, but this manual-dispatch default is hardcoded and will silently drift on the next Go bump — that exact drift has failed the whole matrix here before. Worth a comment pointing at go.mod, or dropping the default so dispatch requires an explicit version.

giuliocalzo:
Pushed follow-up commit `90284516` addressing the latest review items:

- preserve the actionable Kind cluster create error before kube client setup
- select only Running NRI gpu-agent pods before exec, including node-specific lookup
- add the NRI scenario to the root E2E workflow summary/table
- remove the hardcoded manual-dispatch Go version default and require an explicit value matching `go.mod` / `hack/golang-version.sh`

Verification:

- `go test -tags=e2e ./tests/e2e/go ./tests/e2e/go/framework/harness`
- `git diff --check`

ArangoGutierrez (approved):
Thanks Giulio, all findings from the earlier rounds check out at the current head: kubectl and kind installs are checksum-verified again, the Go version is wired back to hack/golang-version.sh on both dispatch paths, the harness surfaces the kind-create error properly, and the docs now match what the workflow actually runs. Full e2e matrix green. LGTM.

PR #445, opened 2026-07-09
fix(security): bump Go 1.26.4 -> 1.26.5 to resolve GO-2026-5856

giuliocalzo:
## Summary
- `govulncheck` in the `golang / check` CI job fails on Go **1.26.4** with **GO-2026-5856** — *Encrypted Client Hello privacy leak in `crypto/tls`* ([advisory](https://pkg.go.dev/vuln/GO-2026-5856)). It is a Go standard-library vulnerability, fixed in **go1.26.5**.
- Bumps every Go pin across the build from `1.26.4` -> `1.26.5`. The CI toolchain version is derived from `deployments/devel/Dockerfile` via `hack/golang-version.sh`, so that pin is the load-bearing fix; the remaining pins (image/test Dockerfiles, `mocknvml`/`mockcuda` Makefiles, the e2e workflow dispatch default, and helper scripts) are bumped for consistency so no build path compiles with the affected compiler.
- Adds a CHANGELOG `Security` entry.

```
Vulnerability #1: GO-2026-5856
    Invoking Encrypted Client Hello privacy leak in crypto/tls
  More info: https://pkg.go.dev/vuln/GO-2026-5856
  Standard library
    Found in: crypto/tls@go1.26.4
    Fixed in: crypto/tls@go1.26.5
```

## Test plan
- [ ] `golang / check` CI job (govulncheck) passes.
- [ ] `hack/golang-version.sh` resolves to `1.26.5`.
- [ ] Image builds (`make docker-build`) succeed on the new base image.


ArangoGutierrez (approved):
Verified the bump end to end: hack/golang-version.sh resolves 1.26.5, the only remaining 1.26.4 strings are the two CHANGELOG history lines, and golang:1.26.5 / 1.26.5-bookworm are live on Docker Hub. GO-2026-5856 is confirmed fixed in 1.26.5 on the 1.26 line. Pin coverage is a superset of the #394 refresh, including the newer compute-domain-daemon Dockerfile. CI is green including govulncheck and the full e2e matrix. LGTM.

PR #461, opened 2026-07-15
ci(e2e): build nvml-mock once and share via ttl.sh

giuliocalzo:
## Summary

- Add a `build-image` job that builds the `nvml-mock` image once per E2E workflow run and pushes it to the ephemeral [ttl.sh](https://ttl.sh) registry (`ttl.sh/nvml-mock-<sha>:1h`).
- Update all five E2E jobs to `needs: [build-image]`, pull the shared image, and load it into Kind instead of rebuilding locally (20 builds → 1 build + 19 pulls).
- Document the CI image-sharing pattern in `tests/e2e/README.md`.

Fixes #460

## Test plan

- [ ] CI `nvml-mock E2E` workflow passes on this PR (all matrix profiles and multi-node jobs)
- [ ] `build-image` job pushes successfully to ttl.sh
- [ ] Downstream jobs pull and load the shared image without image pull errors
- [ ] `workflow_dispatch` with custom `gpu_profiles` still works

Issue #436, opened 2026-07-07
Inject CDI device references through the NRI plugin


ArangoGutierrez:
Part of the NRI node-wide injection epic (v0.3.0 flagship), building on PR #433.

Today the NRI plugin injects raw device nodes, mounts, and env directly into the OCI spec. Newer NRI API revisions support injecting CDI device references instead, letting the runtime resolve them — the same delivery path the real stack uses with DRA (the GPU DRA driver hands out CDI devices).

Scope:
- Generate CDI specs for the mock devices as part of overlay staging (setup.sh or the NRI pod).
- Teach adjust.go to inject CDI device references when available, falling back to raw injection otherwise.
- Verify on kind with a containerd version that supports CDI.

This aligns the mock's delivery mechanism with the DRA-era production path and is the stepping stone for ComputeDomain-style testing (see the IMEX issue).

PR #464, opened 2026-07-16
ci(e2e): add dedicated Kind node image with nvidia-container-toolkit + CDI

giuliocalzo:
## Summary

- Adds `deployments/kind-nvidia-cdi/`, a dedicated Kind node image that extends the pinned `kindest/node` base with `nvidia-container-toolkit` installed and CDI + the nvidia runtime baked into `/etc/containerd/config.toml` and `/etc/nvidia-container-runtime/config.toml`.
- Replaces installing/configuring the toolkit inline on every e2e run with a pre-built, reusable image.
- Includes a small `Makefile` with `build`/`push` targets for local use.

This is the self-contained image definition only. Nothing in CI builds or consumes it yet — the workflow wiring lands with #436, which needs a kind cluster whose containerd supports CDI and is the reason this image exists.

An earlier revision of this description pointed at #461 for that wiring. That was wrong: #461 is closed, and it was the nvml-mock ttl.sh sharing PR, superseded by #465. It never contained `node_image` wiring.

The equivalent setup today happens at test runtime in `tests/e2e/go/scenario_gpu_operator_test.go` (toolkit install + `nvidia-ctk runtime configure`), plus `enable_cdi = true` via `containerdConfigPatches` in the `tests/e2e/go/assets/kind-*.yaml` configs. #436 moves that onto this image.

## Test plan

- [x] `make -C deployments/kind-nvidia-cdi build` succeeds
- [x] Extracted `/etc/containerd/config.toml` is valid TOML with `enable_cdi=true`, `default_runtime_name=nvidia`, and the nvidia runtime handler
- [x] `/etc/nvidia-container-runtime/config.toml` is in CDI mode

Verified 2026-07-28. Both arches build (arm64 native, amd64 emulated); the live checks below ran on arm64.

- Configs parsed with a real TOML parser (Python `tomllib`) rather than grep: `enable_cdi=true`, `default_runtime_name=nvidia`, nvidia handler `BinaryName=/usr/bin/nvidia-container-runtime` (binary present in the image), and `mode="cdi"` with `default-kind=nvidia.com/gpu`.
- A kind cluster booted from the image. containerd 2.2.0 migrated the v2 config and reports `enable_cdi=true` and `default_runtime_name=nvidia` in `containerd config dump`. All system pods came up through the nvidia runtime, no errors in the containerd journal.
- CDI injection end to end: a spec placed in `/etc/cdi` was resolved and injected into a container through the CRI `CDIDevices` path.
- The vendored `containerd-config.toml` diffs against the real base image config as exactly the header comment plus the three marked NVIDIA changes — a faithful fork with nothing dropped.

## Maintainer follow-up (82a548b9)

@giuliocalzo is on PTO, so I pushed one commit to land this for v0.3.0-rc1:

- Pin `nvidia-container-toolkit` to 1.19.1-1. The four toolkit packages resolve in lockstep, so pinning only the top-level one still lets apt pick a different release for the dependencies. Exposed as a build arg and a Makefile variable.
- Move the `nvidia-container-runtime.toml` `COPY` below the package install. The toolkit ships its own config at that path and rewrites it during setup, so the earlier `COPY` was not authoritative and the file in the image was not the file in the repo. The extracted config is now byte-identical to it.

Two notes for follow-up work:

- No CI job builds this image, so a later edit to it breaks silently. Worth a `paths`-filtered `make build` job on `deployments/kind-nvidia-cdi/**`.
- containerd in the current base is 2.x, where CDI injection through `cdi.k8s.io/*` annotations no longer works. Only the CRI `CDIDevices` path does, which is what #436 should target.


deployments/kind-nvidia-cdi/containerd-config.toml
```diff
@@ -0,0 +1,61 @@
...
+  type = "snapshot"
+  address = "/run/containerd-fuse-overlayfs.sock"
+
+[plugins."io.containerd.grpc.v1.cri".containerd]
+  # save disk space when using a single snapshotter
+  discard_unpacked_layers = true
+  # explicitly use default snapshotter so we can sed it in entrypoint
+  snapshotter = "overlayfs"
+  # NVIDIA: default to the nvidia runtime so CDI-injected pods work by default
+  default_runtime_name = "nvidia"
```
ArangoGutierrez, line 26: In CDI mode the nvidia runtime just passes through for pods that request no devices, so this cluster-wide default is functionally fine and matches the standard `nvidia-ctk --set-as-default` pattern. The only tradeoff is that a broken toolkit/runtime would take down every pod including the control plane — keeping runc as the default plus an `nvidia` RuntimeClass is the alternative if you'd rather isolate that risk. Non-blocking either way.

ArangoGutierrez (commented):
One non-blocking design note on the default runtime — nothing here blocks merge, and the CDI/nvidia-container-toolkit setup looks right. Thanks for baking this into a dedicated node image.

1. In CDI mode the nvidia runtime just passes through for pods that request no devices, so this cluster-wide default is functionally fine and matches the standard `nvidia-ctk --set-as-default` pattern. The only tradeoff is that a broken toolkit/runtime would take down every pod including the control plane — keeping runc as the default plus an `nvidia` RuntimeClass is the alternative if you'd rather isolate that risk. Non-blocking either way. (deployments/kind-nvidia-cdi/containerd-config.toml:26)

ArangoGutierrez:
Nice — CI is fully green here (CodeQL, DCO, golang, and the whole e2e matrix). Two things before this can go in: it's still marked draft, and nothing references the new kind-nvidia-cdi image yet, so the actual payoff (pointing a GPU-operator e2e job at this node image) isn't wired up. Are you planning to add that integration in this PR, or land the image here and wire it up in a follow-up? Happy to review + approve once it's marked ready.

PR #466, opened 2026-07-16
feat(nri): mock IMEX channel injection over NRI (#437)

### LINKED ISSUE #437: IMEX mock on the NRI foundation (channels, daemon, ComputeDomain demo)
Part of the NRI node-wide injection epic (v0.3.0 flagship), building on PR #433. The main payoff of the NRI direction.

Real IMEX (GB200/NVL72-class multi-node NVLink domains) involves per-node nvidia-imex daemons coordinating across the domain, IMEX channel device nodes (/dev/nvidia-caps-imex-channels/channelN) delivered into containers via the DRA driver's ComputeDomain (CDI), and NVML fabric APIs. Without hardware, the hard part is getting channel devices into arbitrary containers — which the NRI plugin now solves generically.

Scope:
- Inject mock /dev/nvidia-caps-imex-channels/* device nodes per container via the NRI plugin (annotation- or claim-driven).
- Mock nvidia-imex daemon reusing the mockib daemon pattern (per-node daemon + Unix socket + cross-node coordination via Service), including nvidia-imex-ctl status surface.
- Multi-node ComputeDomain-style demo on kind (GB200/GB300 profiles; builds on the fabricmanager/NVSwitch NVML surface from #387).

Acceptance: a workload spanning two kind nodes sees consistent mock IMEX channels and a healthy mock IMEX domain status.

giuliocalzo:
## What This PR Does

Adds mock IMEX channel device-node injection to the existing `nvml-mock` NRI plugin, so opt-in containers see `/dev/nvidia-caps-imex-channels/channel0..N-1` on Kind nodes without real GPUs. Closes #437.

- **NRI plugin** (`pkg/nri/nvmlmock/adjust.go`): annotation-gated, fail-open discovery/injection of channel nodes as `LinuxDevice` entries. Opt-in via pod annotation `nvml-mock.nvidia.com/imex-channels: "true"`; additive to the existing `/dev/nvidia*` device injection; missing/unreadable tree fails open (container still starts).
- **Plugin flags** (`cmd/nvml-mock-nri`): `--imex-channel-annotation`, `--imex-channel-host-path` (+ matching env).
- **Host provisioning** (`deployments/nvml-mock/scripts/setup.sh`): `mknod`s presence-only channel nodes, driven by new env.
- **Helm** (`nvml-mock` chart): `imexChannels` values (`count: 16`, `major: 240`), main-DaemonSet env, NRI args, helm-unittest coverage (+ regenerated snapshot).
- **Demo** (`docs/demo/nri-imex/`): dedicated 2-node ComputeDomain demo proving channels are injected and the real `nvidia-imex` domain is healthy.
- **E2E** (`tests/e2e/go/scenario_imex_test.go`): dedicated `Label("imex")` scenario asserting each worker sees the expected channel count with consistent ComputeDomain identity.

**Limitation:** channels are presence-only — statable/listable but not openable (`open()` → `ENXIO`), consistent with the other mock device nodes. Documented in the demo README and chart values. The real `nvidia-imex` domain-status check is demo-only (kept out of e2e for speed/hermeticity).

## Why

Issue #437 asks for an IMEX mock built on the NRI foundation. This reuses the real `nvidia-imex --nogpu` daemon already in the stack and focuses the mock on channel-node injection, matching existing `nvml-mock` patterns (annotation opt-in, overlay host path, Helm-configured counts).

## Checklist

- [x] Documentation updated (demo + README index)
- [x] CHANGELOG.md updated

## Test plan

- [x] `go test -race ./...` (excl. vendor) green
- [x] `golangci-lint run` — 0 issues
- [x] `helm unittest deployments/nvml-mock/helm/nvml-mock` — 127/127
- [x] `go test -tags e2e -c ./tests/e2e/go/` compiles
- [ ] E2E `Label("imex")` run on a Kind cluster (CI)
- [ ] Manual demo run: `cd docs/demo/nri-imex && ./run.sh`

pod logs
```
[imex-workload] node=nvml-mock-nri-imex-worker verifying NRI channel injection
=== /dev/nvidia-caps-imex-channels ===
channel0
channel1
channel10
channel11
channel12
channel13
channel14
channel15
channel2
channel3
channel4
channel5
channel6
channel7
channel8
channel9
=== check-fabric (ComputeDomain identity via NRI) ===
Discovered 8 GPU(s)
GPU 0 (GPU-b200b200-0000-0000-0000-000000000000)
  clusterUuid : 00000000-0000-0000-0000-0000000000ce
  cliqueId    : 0
  state       : completed (3)
GPU 1 (GPU-b200b200-0000-0000-0000-000000000001)
  clusterUuid : 00000000-0000-0000-0000-0000000000ce
  cliqueId    : 0
  state       : completed (3)
  ...
  ```

### DIFF
--- a/deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml
+++ b/deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml
@@ -736,3 +736,20 @@ tests:
           content:
             name: MOCK_FABRICMANAGER_STATE_DIR
             value: /run/nvidia-fabricmanager
+
+  - it: should export MOCK_IMEX_CHANNELS env from imexChannels values
+    set:
+      imexChannels:
+        count: 16
+        major: 240
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: MOCK_IMEX_CHANNELS
+            value: "16"
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: MOCK_IMEX_CHANNEL_MAJOR
+            value: "240"
--- a/deployments/nvml-mock/helm/nvml-mock/tests/nri_daemonset_test.yaml
+++ b/deployments/nvml-mock/helm/nvml-mock/tests/nri_daemonset_test.yaml
@@ -130,3 +130,15 @@ tests:
       - contains:
           path: spec.template.spec.containers[0].args
           content: --excluded-namespaces=NAMESPACE,kube-system,monitoring,gpu-system
+
+  - it: should pass IMEX channel args to the NRI plugin
+    set:
+      nri:
+        enabled: true
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: --imex-channel-annotation=nvml-mock.nvidia.com/imex-channels
+      - contains:
+          path: spec.template.spec.containers[0].args
+          content: --imex-channel-host-path=/var/lib/nvml-mock/driver/dev/nvidia-caps-imex-channels
--- a/deployments/nvml-mock/helm/nvml-mock/values.yaml
+++ b/deployments/nvml-mock/helm/nvml-mock/values.yaml
@@ -58,6 +58,16 @@ image:
 
 driverVersion: ""                  # Auto-derived from profile. Override: "550.163.01"
 
+# Mock IMEX channel device nodes (issue #437). The main DaemonSet provisions
+# /dev/nvidia-caps-imex-channels/channel0..count-1 on each node via setup.sh;
+# the NRI plugin injects them into pods annotated
+# nvml-mock.nvidia.com/imex-channels: "true". Presence-only: the nodes use a
+# driverless major, so they are statable/listable but not openable (see
+# docs/demo/nri-imex). Set count to 0 to disable host provisioning.
+imexChannels:
+  count: 16
+  major: 240
+
 nodeSelector: {}
 tolerations:
   - operator: Exists
@@ -83,6 +93,9 @@ nri:
   # device-node mounts from the staged mock overlay. Treat that pod-authored
   # control as part of the demo trust boundary and exclude namespaces as needed.
   deviceAnnotation: nvml-mock.nvidia.com/devices
+  # Workloads that set this pod annotation to "true" opt into mock IMEX channel
+  # device nodes (/dev/nvidia-caps-imex-channels/*) from the staged overlay.
+  imexChannelAnnotation: nvml-mock.nvidia.com/imex-channels
   excludedNamespaces: []
   resources: {}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- feat(nri): inject mock IMEX channel device nodes
+  (`/dev/nvidia-caps-imex-channels/*`) into annotated workloads over the
+  NRI foundation, with a dedicated `nri-imex` demo and e2e scenario. (#437)
+
 ### Changed
 - ComputeDomain simulation now runs the REAL `nvidia-imex` daemon in NO
   GPU mode (`--nogpu`) instead of the fake marker-file binaries: the new
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   NRI foundation, with a dedicated `nri-imex` demo and e2e scenario. (#437)
 
 ### Changed
+- The nri-imex demo now exercises the REAL `nvidia-imex-ctl` against a live
+  NO-GPU IMEX domain formed across both workers (`-q` READY, `-N -j` UP with
+  2 nodes READY/NO_GPU), replacing the previous best-effort domain-status
+  check. (#437)
 - ComputeDomain simulation now runs the REAL `nvidia-imex` daemon in NO
   GPU mode (`--nogpu`) instead of the fake marker-file binaries: the new
   `imex-nogpu-shim` injects the flag around upstream's hard-coded argv,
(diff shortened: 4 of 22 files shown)

### CODE REVIEW

File: deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml [hunk 1]
```diff
@@ -736,3 +736,20 @@ tests:
           content:
             name: MOCK_FABRICMANAGER_STATE_DIR
             value: /run/nvidia-fabricmanager
+
+  - it: should export MOCK_IMEX_CHANNELS env from imexChannels values
+    set:
+      imexChannels:
+        count: 16
```
ArangoGutierrez, line 743: count/major here are 16/240, which are the values.yaml defaults, so the assertions hold even if the template drops the .Values references entirely. I replaced both expressions with the literals "16" and "240" and helm unittest still reported 127 passed / 127 total, 15 snapshots passed. The three updated snapshots render with defaults too, so nothing in the suite actually covers the wiring. Overriding with a non-default pair (count 4, major 250) the way the fabricmanager.stateDir test does would make it discriminate.

deployments/nvml-mock/helm/nvml-mock/tests/nri_daemonset_test.yaml
```diff
@@ -130,3 +130,15 @@ tests:
       - contains:
           path: spec.template.spec.containers[0].args
           content: --excluded-namespaces=NAMESPACE,kube-system,monitoring,gpu-system
+
+  - it: should pass IMEX channel args to the NRI plugin
```
ArangoGutierrez, line 134: Same shape: both args are asserted at their default values, so hard-coding them in the template leaves the suite green at 127/127. The custom-overlay case above sets nri.overlay.hostPath to /custom/nvml-mock but never asserts --imex-channel-host-path, so the derived path is never exercised against a non-default overlay root. Adding that assertion there would cover it.

deployments/nvml-mock/helm/nvml-mock/values.yaml
```diff
@@ -58,6 +58,16 @@ image:
...
 
+# Mock IMEX channel device nodes (issue #437). The main DaemonSet provisions
+# /dev/nvidia-caps-imex-channels/channel0..count-1 on each node via setup.sh;
+# the NRI plugin injects them into pods annotated
+# nvml-mock.nvidia.com/imex-channels: "true". Presence-only: the nodes use a
+# driverless major, so they are statable/listable but not openable (see
+# docs/demo/nri-imex). Set count to 0 to disable host provisioning.
+imexChannels:
+  count: 16
+  major: 240
```
ArangoGutierrez, line 69: imexChannels.count/major and nri.imexChannelAnnotation have no entries in values.schema.json, while the comparable knobs all do — gpu.count is bounded 1..64, nri.deviceAnnotation carries minLength 1, fabricmanager.stateDir is typed. major is the one new value that reaches mknod on the host, so a bound there seems worth having.

ArangoGutierrez (commented):
Feature reads well end to end — the fail-open channel discovery mirrors the existing device path, the IsDir guard on discoverDevices is a good catch, and reusing agentPodOnNode across both scenarios is a nice cleanup. Three notes, all on the chart tests and schema. Separately: the branch is conflicting with main right now, so it needs a rebase before it can go anywhere.

1. count/major here are 16/240, which are the values.yaml defaults, so the assertions hold even if the template drops the .Values references entirely. I replaced both expressions with the literals "16" and "240" and helm unittest still reported 127 passed / 127 total, 15 snapshots passed. The three updated snapshots render with defaults too, so nothing in the suite actually covers the wiring. Overriding with a non-default pair (count 4, major 250) the way the fabricmanager.stateDir test does would make it discriminate. (deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml:743)
2. Same shape: both args are asserted at their default values, so hard-coding them in the template leaves the suite green at 127/127. The custom-overlay case above sets nri.overlay.hostPath to /custom/nvml-mock but never asserts --imex-channel-host-path, so the derived path is never exercised against a non-default overlay root. Adding that assertion there would cover it. (deployments/nvml-mock/helm/nvml-mock/tests/nri_daemonset_test.yaml:134)
3. imexChannels.count/major and nri.imexChannelAnnotation have no entries in values.schema.json, while the comparable knobs all do — gpu.count is bounded 1..64, nri.deviceAnnotation carries minLength 1, fabricmanager.stateDir is typed. major is the one new value that reaches mknod on the host, so a bound there seems worth having. (deployments/nvml-mock/helm/nvml-mock/values.yaml:69)

ArangoGutierrez:
Giulio — closing this in favour of #549 while you're out. Sorry to do it without you;
happy to reopen and talk it through when you're back.

The reason isn't the work, it's timing. #541 landed a mock IMEX surface on main on the
29th, after you wrote this, and it provisions the same device nodes: same directory
(`$DRIVER_ROOT/dev` is `$DEV_ROOT` in both), different major (235 vs your 240), different
count, plus a `proc-devices` file that advertises 235 to the DRA driver. And `mknod`
doesn't overwrite — it returns EEXIST and `setup.sh` swallows that with `|| true` — so
landing both would have left channels 0-15 on one major and the rest on another, in a
single directory, with proc-devices naming only one of them. Nothing would have failed
loudly.

Your base is also 29 commits behind now and 14 of the files you touch have moved,
including adjust.go, scenario_nri_test.go and both DaemonSet templates. Rebasing that
onto a foundation that now has #539, #540 and #548 looked more likely to produce a clean
merge that was quietly wrong than to produce something we could trust.

So #549 keeps your injection half and drops the provisioning half — it consumes
`imex.mockChannels` instead of standing up a second mechanism. The annotation, both
flags, the discovery shape and the directory-skip fix are all yours, and the commits
carry your `Co-authored-by`.

Two things from your branch worth recording:

- Your `IsDir` skip was a real bug fix, not cleanup. `discoverDevices` on main has no
  directory check, and #541 creates `nvidia-caps-imex-channels/` inside the device root
  it scans, so any pod with the `devices` annotation on a channel-enabled node was
  already being handed a directory as a device node. #549 carries that fix with a test.
- `e2e-nri (gb300)` wasn't a gb300 problem. The job died on `No space left on device`
  during `kind load docker-image` — `scenario_imex_test.go` stands up a second five-node
  cluster next to the main one. #549's specs ride the existing NRI cluster instead.

#437 stays open: neither this nor #549 delivers the domain-status half of its acceptance.

PR #469, opened 2026-07-16
test(e2e): add Ginkgo skipped-spec summary report

giuliocalzo:
## Summary

- Adds a `ReportAfterSuite("skipped specs summary", ...)` node to the Go/Ginkgo e2e suite that prints one consolidated, easy-to-find block at the end of a run listing every skipped spec and its reason.
- Programmatic `Skip("...")` reasons are read from `Failure.Message`; specs excluded by `--label-filter`/focus get a distinct `(filtered: did not match label filter)` marker so the two buckets are visually separable.
- Always prints the block (shows `No specs skipped.` when none), emits only on parallel process 1, and is purely informational — it never fails the suite.

Motivation: under `-v` multi-profile output the per-spec `S` markers scatter through a long log, making it hard to see at a glance what was skipped and why (IB disabled per profile, no `device_defaults.fabric` block, the `E2E_RUN_NGC` validator gate, label-filter exclusions).

Example output:

```
========== SKIPPED SPECS (2) ==========
- ... nvidia-smi ... [profile a100]
    reason: IB disabled for profile a100
- ... validator scenario
    reason: set E2E_RUN_NGC=true to run nvcr.io-backed validator scenario; see #446
=======================================
```

## Test plan

- [x] `go vet -tags e2e ./tests/e2e/go`
- [x] `ginkgo --tags e2e --dry-run ./tests/e2e/go` (suite compiles and outlines, 26/26 specs)
- [ ] Full `make e2e E2E_PROFILES=a100` run to confirm the block renders with real skip reasons (requires Kind/Docker)

### DIFF
--- a/tests/e2e/go/report_skips_test.go
+++ b/tests/e2e/go/report_skips_test.go
@@ -1,0 +1,50 @@
+//go:build e2e
+
+// Copyright 2026 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+package e2e
+
+import (
+	"fmt"
+	"strings"
+
+	. "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/ginkgo/v2/types"
+)
+
+// ReportAfterSuite prints a single, easy-to-find summary of every skipped spec
+// and its reason. Ginkgo scatters per-spec `S` markers through the verbose
+// multi-profile log; this consolidates them so it is obvious at a glance what
+// was skipped and why. It is purely informational and never fails the suite.
+var _ = ReportAfterSuite("skipped specs summary", func(report Report) {
+	// ReportAfterSuite runs once, but guard anyway so parallel runners never
+	// emit the block twice.
+	if GinkgoParallelProcess() != 1 {
+		return
+	}
+
+	var lines []string
+	for _, spec := range report.SpecReports {
+		if spec.State != types.SpecStateSkipped {
+			continue
+		}
+		reason := strings.TrimSpace(spec.Failure.Message)
+		if reason == "" {
+			reason = "(filtered: did not match label filter)"
+		}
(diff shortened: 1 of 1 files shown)


tests/e2e/go/report_skips_test.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	var lines []string
+	for _, spec := range report.SpecReports {
+		if spec.State != types.SpecStateSkipped {
+			continue
+		}
+		reason := strings.TrimSpace(spec.Failure.Message)
+		if reason == "" {
+			reason = "(filtered: did not match label filter)"
```
ArangoGutierrez: An empty Failure.Message does not identify a label-filter miss: Ginkgo also produces empty failures for focus/skip filters, suite-wide skips, interrupts, and deadlines. Use a cause-neutral fallback unless the configured filter can be identified.

ArangoGutierrez (commented):
Found 1 issue.

1. An empty Failure.Message does not identify a label-filter miss: Ginkgo also produces empty failures for focus/skip filters, suite-wide skips, interrupts, and deadlines. Use a cause-neutral fallback unless the configured filter can be identified. (tests/e2e/go/report_skips_test.go:34)

tests/e2e/go/report_skips_test.go
giuliocalzo: Fixed in 9ca28744. Replaced the hardcoded "(filtered: did not match label filter)" with a cause-neutral fallback `emptySkipReason()`: for empty-message skips it now reports "(no skip reason reported)", and only appends "possibly excluded by ..." naming the filters actually present in report.SuiteConfig (label-filter/focus/skip). It no longer asserts label filtering as the cause for interrupts, deadlines, or suite-wide skips.

PR #474, opened 2026-07-17
ci: add self-managing repository automation

ArangoGutierrez:
## Summary

- add open-source policy, templates, labels, ownership, dependency updates, stale handling, and hardened workflow governance
- automate PR size, type, area labels, and deterministic reviewer routing
- add Kubernetes-style slash commands, current-head LGTM and approval provenance, and native auto-merge reconciliation
- make Release Please the version authority and publish digest-bound multi-platform images and OCI Helm charts with SBOMs, signatures, provenance, and resumable fail-closed checks
- retire the legacy artifact publishers atomically so release.yml is the sole publisher

## Validation

- Go test suite passed
- repository automation tests: 760 of 760
- npm audit: zero vulnerabilities; ESLint and reproducible package passed
- Helm lint and unit tests: 129 of 129 with 15 snapshots
- all-workflow actionlint, JSON, diff, secret, action-pin, permission, signature, and DCO checks passed
- local release image build and Helm package inspection passed
- independent final QA approved

## Rollout notes

- no PAT or static signing key is introduced; workflows use scoped GITHUB_TOKEN permissions and keyless OIDC
- manual publish false is read-only; manual publication is restricted to the default branch
- legacy publishers are removed in the same change to avoid races
- activation, required checks, first-run verification, rollback, and release operations are documented in docs/maintainers/repository-automation.md
- no release, tag, workflow dispatch, registry publication, or repository-setting mutation was performed while preparing this PR


.github/actions/repo-automation/dist/index.js
roma-glushko, line 1: Do we have to commit compiled JS?

.github/repo-automation/labels.yml
```diff
@@ -0,0 +1,107 @@
+schemaVersion: 1
+labels:
```
roma-glushko, line 2: It's a great idea to define labels declaratively!

.github/actions/repo-automation/src/dco.js
```diff
@@ -0,0 +1,256 @@
+"use strict";
+
+const TRAILER_LINE = /^([A-Za-z0-9][A-Za-z0-9-]*):[ \t]*([\s\S]*)$/;
```
roma-glushko, line 3: Would be a bit better if we used TypeScript instead of plain JS here. Types help to save our backs from errors.

roma-glushko (dismissed):
Great changes 👏 
Excited to see the new policy system in action!

.github/actions/repo-automation/dist/index.js
ArangoGutierrez: Yes. action.yml executes dist/index.js with the Node 24 action runtime, and a checked-out local JavaScript action does not install its npm dependency tree before execution. The committed bundle is generated with the pinned ncc version; the package verification regenerates it and fails on any dist drift.

.github/actions/repo-automation/src/dco.js
ArangoGutierrez: The custom action is currently JavaScript end to end across its source modules and tests. Converting only dco.js would add a mixed compilation path without providing types across its callers and GitHub boundary. I am keeping this PR internally consistent; a TypeScript conversion should migrate the whole action as a separate focused change.

.github/repo-automation/labels.yml
ArangoGutierrez: The label-sync workflow treats this file as the authoritative managed label set and applies create or metadata-update operations idempotently while preserving unmanaged labels.

.superpowers/sdd/plan3-task-6-report.md
```diff
@@ -0,0 +1,146 @@
+# Plan 3 Task 6 report
```
giuliocalzo: @ArangoGutierrez maybe we can remove these files

ArangoGutierrez: Removed in e9bbb554, along with the rest of the internal working notes (`.superpowers/` task reports and the `docs/superpowers/` plan/spec files — 11 files, ~3.9k lines). Both paths were already in `.gitignore`; the tracked copies predated the ignore rules. Good catch.

PR #476, opened 2026-07-17
feat(nvml-mock): runtime control via nvml-mock-ctl

giuliocalzo:
## Summary

Adds runtime control of `nvml-mock` device configuration through a new `nvml-mock-ctl` CLI. Overrides are applied to a node-local overlay file that both **already-running** and **newly-started** consumer processes pick up within one TTL — no consumer restart required. Overrides are ephemeral to the `nvml-mock` DaemonSet pod lifecycle: restarting the DaemonSet pod wipes them back to the pristine YAML profile; consumer restarts do not.

Any field in the `DeviceConfig` schema can be overridden generically (`set`), and the most-tweaked readings have first-class convenience commands (`temp`/`power`/`fan`/`util`/`clocks`/`throttle`/`pstate`).

## What's included

**Engine (hot-reload)**
- Overlay document parsing + generic deep-merge over the existing `DeviceConfig` schema (any field is overridable).
- TTL-gated overlay store with a generation counter and lock-free fast path.
- `ConfigurableDevice` now holds config/injector in `atomic.Pointer`s with lazy, TTL-gated refresh (consistent, lock-free hot reads).
- Dynamic-metrics simulator is hot-reloadable (rebuilt on overlay refresh), so runtime overrides of temperature/power/utilization surface on running consumers instead of staying frozen at construction-time config.
- `failureInjector.Reset()` so a device can recover to healthy at runtime.
- `parsePstate` extended to cover P9–P15.

**CLI (`nvml-mock-ctl`)**
- `pkg/gpu/mockctl` mutation library: load/mutate/save overlay, `key.path=value` parsing, `--gpu` targeting by index / `all` / UUID.
- `cmd/nvml-mock-ctl`: `fail`, `set`, `status`, `reset`; serialized via `flock`; atomic, world-readable writes. (The earlier `apply -f patch.yaml` command was dropped — multiple fields can be changed in a single `set` call.)
- Convenience commands that pin a common reading and are observable via `nvidia-smi`:
  - `temp <celsius>`, `power <watts>`, `fan <percent>` — write both the static block and a zero-variation dynamic block (and force `fan.count >= 1`) so the value is deterministic regardless of whether the profile drives the metric dynamically.
  - `util <percent>` — pins GPU + memory utilization; disables the dynamic utilization sub-simulator (rather than zeroing variation) so it is deterministic even at `0%`.
  - `clocks <mhz>` — pins SM + graphics clocks (static, hot-reloads directly).
  - `throttle <reason>[ reason ...]` — sets active clock-throttle reasons; authoritative (requested on, rest off), `none` clears, with friendly aliases (`thermal`/`power`/`power_brake`/`idle`/…).
  - `pstate <0-15>` — pins the reported performance state.

**NVML getter fidelity (so overrides are observable via `nvidia-smi`)**
- Implemented previously-stubbed getters that modern `nvidia-smi` calls, which had reported `[N/A]`: `nvmlDeviceGetTemperatureV` (versioned `temperature.gpu`), `nvmlDeviceGetPciInfoExt` (detailed PCI block), `nvmlDeviceGetMarginTemperature` ("GPU T.Limit Temp"), and the `NVML_FI_DEV_POWER_*` field values (`power.draw.instant` / DCGM). Added the matching C struct bodies and an engine `GetMarginTemperature` + power field-value resolver.

**Delivery & docs**
- Dockerfile ships the binary; DaemonSet mounts the overlay dir; `setup.sh` clears overrides on pod start; CDI mounts the config **directory** so atomic renames are visible to consumers.
- The DaemonSet no longer sets a redundant `MOCK_NVML_OVERRIDES` env var — the CLI's default overlay path already resolves to the host driver config dir.
- E2E scenarios proving inject-then-reset on a running consumer (standalone via nvidia-smi, and GPU Operator via dcgm-exporter without restart), including temp/power/fan/util/clocks/throttle/pstate coverage.
- User docs `docs/nvml-mock-ctl.md`, cross-linked from the READMEs.

## Test plan

- [x] `go test -race` across the module (incl. `mockctl`, `nvml-mock-ctl`, engine) — green
- [x] `golangci-lint` on changed packages — 0 issues
- [x] `helm unittest deployments/nvml-mock/helm/nvml-mock` — passed
- [x] `go vet -tags e2e ./tests/e2e/go/...` — clean
- [x] Standalone runtime-control E2E against a live Kind cluster — specs passing (fail/lost/ECC, `set`, `temp`, `power`, `fan`, UUID, `status`, healthy recovery)
- [ ] Standalone E2E for the new `util`/`clocks`/`throttle`/`pstate` specs against a live Kind cluster (compiled + vetted; live run pending)
- [x] Verified live via `nvidia-smi`: `temp`/`power`/`fan` pin the target GPU only and `reset` reverts; `temperature.gpu`, detailed PCI, `power.draw.instant`, and `GPU T.Limit Temp` now report real values


cmd/nvml-mock-ctl/main.go
```diff
@@ -0,0 +1,477 @@
...
+		if code := applyPatch(doc, target, base, mockctl.UtilizationPatch(pct), stderr); code != 0 {
+			return code
+		}
+	case "clocks":
+		mhz, perr := singleIntArg(positional, cmd, 0, 100000)
+		if perr != nil {
+			fprintf(stderr, "%v\n", perr)
+			return 2
+		}
+		if code := applyPatch(doc, target, base, mockctl.ClocksPatch(uint32(mhz)), stderr); code != 0 {
```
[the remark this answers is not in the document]
giuliocalzo: fixed

giuliocalzo:
@ArangoGutierrez this is ready for review.

Since the initial review request, a few things changed:

- **New convenience commands** on `nvml-mock-ctl`: `util`, `clocks`, `throttle`, and `pstate` (in addition to the existing `temp`/`power`/`fan`), each with builder + CLI unit tests, docs, and standalone e2e assertions (mutate → assert via `nvidia-smi` → scope-check → reset). `parsePstate` was extended to cover P9–P15.
- **Removed** the `apply -f patch.yaml` command — multiple fields can be set in a single `set` call, so the file-based path was redundant.
- **Removed** the redundant `MOCK_NVML_OVERRIDES` env var from the DaemonSet; the CLI's default overlay path already resolves to the host driver config dir.
- **CodeQL fix**: reworked `ClocksPatch` to take an `int` so there's no unchecked `int→uint32` narrowing at the call site (the flagged alert).

The PR description's test plan is up to date. One caveat: the new `util`/`clocks`/`throttle`/`pstate` e2e specs are compiled + `go vet -tags e2e` clean but the live Kind run for those four is still pending (called out as unchecked in the test plan). Everything else — unit tests (`-race`), `golangci-lint`, `helm unittest`, and the existing runtime-control e2e — is green.

Happy to split anything out if you'd prefer a smaller first PR.

cmd/nvml-mock-ctl/main_test.go
```diff
@@ -0,0 +1,250 @@
...
+
+import (
+	"bytes"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func runCLI(t *testing.T, overlay string, args ...string) (string, string, int) {
```
roma-glushko: Does overlay mean a configuration to apply? I would rather call it config override than overlay that also may convey network overlay meaning.

pkg/gpu/mockctl/overlay_test.go
```diff
@@ -0,0 +1,289 @@
...
+
+import (
+	"testing"
+
+	"github.com/NVIDIA/k8s-test-infra/pkg/gpu/mocknvml/engine"
+)
+
+func TestParseSet_TypesAndNesting(t *testing.T) {
+	m, err := ParseSet([]string{"ecc.mode_current=disabled", "failure.after_calls=1", "failure.mode=lost"})
+	if err != nil {
```
roma-glushko: We should use testify/require instead of bare t.Fatal() in tests.

pkg/gpu/mockctl/config_override.go
```diff
@@ -0,0 +1,367 @@
...
+		return nil
+	}
+	failure := map[string]any{"mode": mode}
+	if afterCalls > 0 {
+		failure["after_calls"] = afterCalls
+	}
+	if xidCode > 0 {
+		failure["xid"] = map[string]any{"code": xidCode}
+	}
+	d.SetFields(t, map[string]any{"failure": failure})
```
ArangoGutierrez: Fail routes a partial `{mode: ...}` block through SetFields, which deep-merges rather than replaces, so `fail --mode lost` after `fail --mode ecc_uncorrectable --xid 79` leaves a stale `xid: {code: 79}` in the block and the device keeps reporting Xid 79. Unlike the other patch builders (temp/power/fan/util) this one isn't authoritative; clearing the failure block before setting it would make it consistent.

pkg/gpu/mocknvml/engine/config_override_store.go
```diff
@@ -0,0 +1,163 @@
...
+
+func (s *configOverrideStore) snapshot() (uint64, *ConfigOverrideDoc) {
+	now := s.now()
+
+	// Lock-free fast path: within the TTL window return the last published
+	// generation/doc without touching the mutex. checkedNanos == 0 means
+	// "never checked" and always falls through to the slow path so the first
+	// call performs a real stat/read under the mutex.
+	if checked := s.checkedNanos.Load(); checked != 0 && now.UnixNano()-checked < int64(s.ttl) {
+		return s.genAtomic.Load(), s.docAtomic.Load()
```
ArangoGutierrez: The fast path loads gen then doc, but the slow-path publish stores them in the opposite order (doc, then gen, then checkedNanos). A reader racing a concurrent publish can observe a torn pair — gen from generation N with doc from N+1 — so the documented invariant (a reader that sees a fresh timestamp also sees the matching gen+doc) isn't strictly held for the pair. It's benign today since refresh() keys off gen alone (worst case a one-cycle-late or redundant merge), but bundling gen+doc into a single atomic.Pointer would make the pair truly atomic and match the comment.

pkg/gpu/mocknvml/engine/device.go
```diff
@@ -112,35 +128,134 @@ func NewConfigurableDevice(index int, baseDevice *mockserver.Device, config *Dev
...
+// reconcileFailure aligns the injector with the effective failure config.
+// Unchanged mode keeps the existing injector (preserving accumulated ECC
+// counters); a changed mode installs a fresh injector; healthy clears it.
+func (d *ConfigurableDevice) reconcileFailure(cfg *FailureInjectionConfig) {
+	cur := d.failure.Load()
+	newMode := FailureModeHealthy
+	if cfg != nil {
+		newMode = normalizedMode(cfg.Mode)
+	}
+	if cur.Mode() == newMode {
```
ArangoGutierrez: reconcileFailure keeps the existing injector whenever the mode is unchanged, so same-mode parameter edits (after_calls, xid, probability) are silently dropped at runtime: `fail lost --after-calls 100` then `fail lost --after-calls 5` stays at 100 until the mode changes or the pod restarts. Since the feature advertises overriding any field at runtime, consider replacing the injector when those params change, or documenting the reset workaround.

pkg/gpu/mocknvml/engine/failure_injector.go
```diff
@@ -227,6 +228,20 @@ func (f *failureInjector) ClaimXid() (uint64, bool) {
 	return xid, true
 }
 
+// Reset returns the injector to its untripped state. It is used by runtime
```
ArangoGutierrez: Reset() is unused outside its own unit test — the runtime healthy path in reconcileFailure clears the injector with failure.Store(nil) instead. This comment and the struct comment (line 28) still say Reset is called by runtime control, which is no longer accurate; either wire it up or drop it along with the stale comments.

pkg/gpu/mockctl/config_override.go
```diff
@@ -0,0 +1,367 @@
...
+	_, err := engine.MergeDeviceConfig(base, patch)
+	return err
+}
+
+// ResolveTarget interprets a --gpu spec: "all", a device index, or a UUID.
+func ResolveTarget(spec string, cfg *engine.Config) (Target, error) {
+	if spec == "all" {
+		return Target{All: true}, nil
+	}
+	if idx, err := strconv.Atoi(spec); err == nil {
```
ArangoGutierrez, line 348: ResolveTarget accepts any integer index without a bounds check, so a typo like `--gpu 8` on an 8-GPU node writes to a bucket no device reads and still prints `ok: ... applied`, leaving the user wondering why nothing changed. When cfg is available, validating 0 <= idx < NumDevices (and erroring otherwise) would catch this.

ArangoGutierrez (commented):
Nice work — the lock-free TTL overlay store is a clean design and the e2e coverage is thorough. A few things worth a look before merge, mostly in the failure-injection path.

1. Fail routes a partial `{mode: ...}` block through SetFields, which deep-merges rather than replaces, so `fail --mode lost` after `fail --mode ecc_uncorrectable --xid 79` leaves a stale `xid: {code: 79}` in the block and the device keeps reporting Xid 79. Unlike the other patch builders (temp/power/fan/util) this one isn't authoritative; clearing the failure block before setting it would make it consistent. (pkg/gpu/mockctl/config_override.go:120)
2. The fast path loads gen then doc, but the slow-path publish stores them in the opposite order (doc, then gen, then checkedNanos). A reader racing a concurrent publish can observe a torn pair — gen from generation N with doc from N+1 — so the documented invariant (a reader that sees a fresh timestamp also sees the matching gen+doc) isn't strictly held for the pair. It's benign today since refresh() keys off gen alone (worst case a one-cycle-late or redundant merge), but bundling gen+doc into a single atomic.Pointer would make the pair truly atomic and match the comment. (pkg/gpu/mocknvml/engine/config_override_store.go:90)
3. reconcileFailure keeps the existing injector whenever the mode is unchanged, so same-mode parameter edits (after_calls, xid, probability) are silently dropped at runtime: `fail lost --after-calls 100` then `fail lost --after-calls 5` stays at 100 until the mode changes or the pod restarts. Since the feature advertises overriding any field at runtime, consider replacing the injector when those params change, or documenting the reset workaround. (pkg/gpu/mocknvml/engine/device.go:239)
4. Reset() is unused outside its own unit test — the runtime healthy path in reconcileFailure clears the injector with failure.Store(nil) instead. This comment and the struct comment (line 28) still say Reset is called by runtime control, which is no longer accurate; either wire it up or drop it along with the stale comments. (pkg/gpu/mocknvml/engine/failure_injector.go:231)
5. ResolveTarget accepts any integer index without a bounds check, so a typo like `--gpu 8` on an 8-GPU node writes to a bucket no device reads and still prints `ok: ... applied`, leaving the user wondering why nothing changed. When cfg is available, validating 0 <= idx < NumDevices (and erroring otherwise) would catch this. (pkg/gpu/mockctl/config_override.go:342)

pkg/gpu/mockctl/config_override.go
giuliocalzo: Good catch — `Fail` now replaces the whole `failure` block instead of deep-merging it through `SetFields`, so a stale `xid.code` from a previous fail no longer lingers (matching the temp/power/fan/util builders). Added `TestDocFail_IsAuthoritative`. Fixed in 797dd8c0.

pkg/gpu/mocknvml/engine/config_override_store.go
giuliocalzo: Bundled the `(gen, doc)` pair into a single immutable `configOverrideSnapshot` published via one `atomic.Pointer`, with `checkedNanos` still stored last. The lock-free fast path now returns a matching gen+doc and can never observe a torn pair. Fixed in 797dd8c0.

pkg/gpu/mocknvml/engine/device.go
giuliocalzo: `reconcileFailure` now installs a fresh injector on a same-mode parameter change (after_calls/xid/probability/seed) via a new nil-safe `sameConfig` check; identical configs still keep the injector so accumulated ECC counters survive unrelated edits. Added `TestRefresh_SameModeParamChangeReinstallsInjector`. Fixed in 797dd8c0.

pkg/gpu/mocknvml/engine/failure_injector.go
giuliocalzo: Dropped the unused `Reset()` and corrected the stale struct/method comments (the runtime healthy path stores nil, as you noted). Its comparison logic is repurposed into the new `sameConfig` helper used by `reconcileFailure`. Fixed in 797dd8c0.

pkg/gpu/mockctl/config_override.go
giuliocalzo: `ResolveTarget` now bounds-checks a numeric index against `cfg.NumDevices` when a resolved config is available (skipped when unknown, e.g. no YAML), erroring on out-of-range like `--gpu 8` on an 8-GPU node. Added `TestResolveTarget_IndexBounds`. Fixed in 797dd8c0.

cmd/nvml-mock-ctl/main_test.go
giuliocalzo: Agreed — renamed the concept from "overlay" to "config override" throughout the feature (files, identifiers, CLI help/flags, docs and comments), leaving the unrelated topology/NRI overlay untouched. Done in 1feb0e8.

pkg/gpu/mockctl/overlay_test.go
giuliocalzo: Done — converted the nvml-mock-ctl CLI and mockctl tests to testify/require. Done in 97b9400.

cmd/nvml-mock-ctl/main.go
```diff
@@ -0,0 +1,477 @@
...
+		if code := applyPatch(doc, target, base, mockctl.TemperaturePatch(celsius), stderr); code != 0 {
+			return code
+		}
+	case "power":
+		watts, perr := singleFloatArg(positional, cmd)
+		if perr != nil {
+			fprintf(stderr, "%v\n", perr)
+			return 2
+		}
+		if watts < 0 {
```
ArangoGutierrez: power is the only convenience command without a finite-range guard. strconv.ParseFloat accepts NaN and Inf with a nil error, and the watts < 0 check misses both, so uint32(watts*1000 + 0.5) silently overflows: power ... Inf writes 4294967295 mW and power ... 5000000 wraps to 705032704 mW, both printing ok. Worth rejecting NaN/Inf and bounding the value the way the sibling int commands do with singleIntArg.

tests/e2e/go/scenario_runtime_control.go
```diff
@@ -0,0 +1,684 @@
...
+func assertRuntimeUUIDTargeting(ctx SpecContext, h *harness.Harness, consumer kube.PodRef) {
+	GinkgoHelper()
+	resetRuntimeOverrides(ctx, h)
+
+	uuid := smiGPUValue(ctx, h, consumer, 0, "uuid")
+	Expect(uuid).NotTo(BeEmpty(), "nvidia-smi should report a UUID for GPU 0")
+
+	By("target GPU 0 by UUID with fallen_off_bus via nvml-mock-ctl")
+	out, err := nvmlMockCtlTry(ctx, h, "fail", "--gpu", uuid, "--mode", "fallen_off_bus")
+	if err != nil {
```
ArangoGutierrez, line 605: This skips on any exec error, not just an unresolvable-UUID one, so a real regression that makes fail --gpu <uuid> exit non-zero would be masked as a green skip on the very path this spec covers. Matching the cannot-resolve signal in the output before skipping, and failing otherwise, keeps the coverage honest.

ArangoGutierrez (commented):
Nice work on this. The control surface is clean and the e2e genuinely exercises every command against a running pod without a restart, with discriminating assertions; CI is green across all profiles. Two small robustness nits, neither blocking.

1. power is the only convenience command without a finite-range guard. strconv.ParseFloat accepts NaN and Inf with a nil error, and the watts < 0 check misses both, so uint32(watts*1000 + 0.5) silently overflows: power ... Inf writes 4294967295 mW and power ... 5000000 wraps to 705032704 mW, both printing ok. Worth rejecting NaN/Inf and bounding the value the way the sibling int commands do with singleIntArg. (cmd/nvml-mock-ctl/main.go:191)
2. This skips on any exec error, not just an unresolvable-UUID one, so a real regression that makes fail --gpu <uuid> exit non-zero would be masked as a green skip on the very path this spec covers. Matching the cannot-resolve signal in the output before skipping, and failing otherwise, keeps the coverage honest. (tests/e2e/go/scenario_runtime_control.go:605)

cmd/nvml-mock-ctl/main.go
```diff
@@ -0,0 +1,477 @@
...
+	}
+	return def
+}
+
+func loadConfig(path string) *engine.Config {
+	yc, err := engine.LoadYAMLConfig(path)
+	if err != nil {
+		return nil
+	}
+	return &engine.Config{YAMLConfig: yc, NumDevices: len(yc.Devices), DriverVersion: yc.System.DriverVersion}
```
ArangoGutierrez: loadConfig builds NumDevices from len(yc.Devices), but the engine's LoadConfig overrides that with system.num_devices when set - which setup.sh injects from GPU_COUNT - so the new ResolveTarget bounds check tests against the wrong count whenever gpu.count differs from the profile's device list: on an a100 profile with count: 4, --gpu 5 still prints ok while writing a bucket no device reads, and with num_devices > len(devices) valid indices get rejected. loadConfig also swallows the load error silently, so any config-read failure disables the check with no hint; mirroring the engine's num_devices handling (and warning on load failure like the engine does) would make the guard authoritative.

cmd/nvml-mock-ctl/main_test.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func TestCLI_ConvenienceArgValidation(t *testing.T) {
+	dir := t.TempDir()
+	configOverride := filepath.Join(dir, "overrides.yaml")
+	cases := [][]string{
+		{"temp", "--gpu", "0"},                        // missing value
+		{"temp", "--gpu", "0", "hot"},                 // non-integer
+		{"fan", "--gpu", "0", "150"},                  // out of range
+		{"power", "--gpu", "0", "-5"},                 // negative watts
```
ArangoGutierrez: The negative-watts case never reaches the watts < 0 guard: flag.Parse rejects the bare -5 as an unknown flag, so run() exits from argument parsing and the case still passes with the guard deleted (mutation-checked in a scratch worktree). Passing the value after a -- terminator, or asserting the guard's specific error message, would make the case cover what its comment claims.

ArangoGutierrez (commented):
Found 2 issues.

1. loadConfig builds NumDevices from len(yc.Devices), but the engine's LoadConfig overrides that with system.num_devices when set - which setup.sh injects from GPU_COUNT - so the new ResolveTarget bounds check tests against the wrong count whenever gpu.count differs from the profile's device list: on an a100 profile with count: 4, --gpu 5 still prints ok while writing a bucket no device reads, and with num_devices > len(devices) valid indices get rejected. loadConfig also swallows the load error silently, so any config-read failure disables the check with no hint; mirroring the engine's num_devices handling (and warning on load failure like the engine does) would make the guard authoritative. (cmd/nvml-mock-ctl/main.go:345)
2. The negative-watts case never reaches the watts < 0 guard: flag.Parse rejects the bare -5 as an unknown flag, so run() exits from argument parsing and the case still passes with the guard deleted (mutation-checked in a scratch worktree). Passing the value after a -- terminator, or asserting the guard's specific error message, would make the case cover what its comment claims. (cmd/nvml-mock-ctl/main_test.go:132)

cmd/nvml-mock-ctl/main.go
giuliocalzo: Fixed in 0ecf4978. `power` now goes through `singleFloatArg(positional, cmd, 0, maxPowerWatts)`, which rejects NaN/±Inf explicitly and bounds watts to a finite [0, 100000] range — well under the uint32 milliwatt ceiling — so `uint32(watts*1000 + 0.5)` can no longer overflow. Added NaN/Inf/overflow test cases too.

tests/e2e/go/scenario_runtime_control.go
giuliocalzo: Fixed in 0ecf4978. The spec now only skips when the CLI output contains the "cannot resolve" signal (the documented v1 UUID-resolution limitation). Any other exec failure now fails the spec via `Expect(err).NotTo(HaveOccurred())` instead of being masked as a green skip.

cmd/nvml-mock-ctl/main.go
giuliocalzo: Fixed in 0ecf4978. `loadConfig` now resolves `NumDevices` the same way the engine's `LoadConfig` does: device-list length unless `system.num_devices` (which setup.sh injects from GPU_COUNT) overrides it, so the `--gpu` bounds check matches the count the running engine actually serves. Config-load errors are now surfaced as a warning instead of being swallowed silently.

cmd/nvml-mock-ctl/main_test.go
giuliocalzo: Fixed in 0ecf4978. The negative-watts arg is now passed after a `--` terminator so it reaches the guard instead of being consumed by `flag.Parse` as an unknown flag; with the guard deleted the case now fails (mutation-tested). Also added NaN/Inf and overflow cases for good measure.

giuliocalzo:
@ArangoGutierrez thanks for the second review pass — all four items are addressed in `0ecf4978` (single commit, rebased onto the latest branch tip):

1. **`power` range guard** (`cmd/nvml-mock-ctl/main.go`) — `singleFloatArg` now rejects `NaN`/`±Inf` and bounds watts to a finite `[0, 100000]` range, well under the uint32 milliwatt ceiling, so `uint32(watts*1000 + 0.5)` can't overflow. Dropped the redundant `watts < 0` check.
2. **UUID e2e skip** (`tests/e2e/go/scenario_runtime_control.go`) — the spec now only skips on the documented `cannot resolve` signal; any other `exec` failure fails the spec via `Expect(err).NotTo(HaveOccurred())` instead of being masked as a green skip.
3. **`loadConfig` NumDevices** (`cmd/nvml-mock-ctl/main.go`) — resolves `NumDevices` the same way the engine's `LoadConfig` does (`system.num_devices` wins over the device-list length), so the `--gpu` bounds check matches what the engine serves. Load errors are now surfaced as a warning rather than swallowed silently.
4. **Negative-watts test** (`cmd/nvml-mock-ctl/main_test.go`) — the `-5` arg is now passed after `--` so it reaches the guard instead of being consumed by `flag.Parse`; added `NaN`/`Inf`/overflow cases.

Verified locally: `gofmt` clean, `go build`/`go vet` clean (CLI + e2e with `-tags e2e`), and `go test ./cmd/nvml-mock-ctl/... ./pkg/gpu/mockctl/...` pass. Individual review threads have in-thread replies as well.

ArangoGutierrez (approved):
Re-reviewed at 7b16d4c98. The power guard is correct - NaN/Inf/negative/overflow all rejected and watts*1000 stays well under the uint32 ceiling - and loadConfig now resolves the device count exactly like the engine (system.num_devices override), so the --gpu bounds check lines up. The e2e skip is nicely tightened to only the documented "cannot resolve" case. Failure-injection feedback all addressed with tests. Full e2e matrix green. LGTM.

PR #477, opened 2026-07-20
fix(mockpcisysfs): make lspci enumerate mock GPUs

giuliocalzo:
## Summary

`lspci` inside the nvml-mock pod failed fatally (`lspci: Cannot open /sys/bus/pci/devices/<bdf>/vendor`) because the fake PCI sysfs tree only carried topology symlinks + `numa_node`. `libpci` reads the device identity attribute files with die-on-error, so the missing `vendor`/`device`/`class`/... made it abort even though `libpcimocksys.so` was correctly redirecting `/sys/bus/pci` (via `LD_PRELOAD` + `MOCK_PCI_ROOT`).

- **Render PCI identity attribute files** — `render-pci-sysfs` now writes `vendor`, `device`, `class` (`0x030200`, 3D controller), `revision`, `irq`, `subsystem_vendor`, `subsystem_device`, a synthetic 256-byte binary `config` space, and a `resource` BAR table for each device. Values are unpacked from the profile's existing `pci.device_id` / `subsystem_id` (`device<<16 | vendor`), inherited from `device_defaults` with per-device overrides. Falls back to the NVIDIA vendor (`0x10de`) when a profile omits IDs, so `lspci` never fatals.
- **Close two shim redirection gaps this uncovered:**
  - Hook the `_FORTIFY_SOURCE` open variants (`__open_2`/`__open64_2`/`__openat_2`/`__openat64_2`). `libpci` opens `config` with a non-constant flags argument, which glibc rewrites to `__open_2` — bypassing the `open`/`openat` hooks and hitting the real host path (this was the residual `pcilib: Cannot open .../config` warning).
  - Hook `fopen`/`fopen64`, which `libpci` uses to read `resource` for `lspci -v`.

Only PCI sysfs paths are rewritten; every other `open`/`fopen` passes through untouched.

## Verification

Built the real shim + renderer in a `golang:bookworm` + `pciutils` container, rendered the H100 profile, and ran actual `lspci`:

```
$ lspci
1a:00.0 3D controller: NVIDIA Corporation GH100 [H100 SXM5 80GB]
...(all 8 GPUs)...
$ lspci -nn
1a:00.0 3D controller [0302]: NVIDIA Corporation GH100 [H100 SXM5 80GB] [10de:2330]
$ lspci -x -s 0000:1a:00.0
00: de 10 30 23 00 00 00 00 00 00 02 03 00 00 00 00
20: 00 00 00 00 00 00 00 00 00 00 00 00 de 10 58 16
```

Plain `lspci`, `-nn`, `-v`, and `-x` all render cleanly with no `pcilib`/`lspci` warnings.

## Test plan

- [x] `go test ./pkg/system/mockpcisysfs/... ./cmd/render-pci-sysfs/...`
- [x] `golangci-lint run` on changed packages (0 issues)
- [x] Integration tests incl. new `TestFortifiedOpenPCIRedirect` (compiles a fortified-open C helper) pass in-container
- [x] e2e assertion extended to check the NVIDIA identity attributes
- [ ] CI: lint, unit, Helm unittest, E2E


pkg/system/mockpcisysfs/shim/shim_test.go
```diff
@@ -80,6 +80,76 @@ func TestOpenSysDevicesPCIRedirect(t *testing.T) {
...
+
+	// Helper: open(argv[2], flags=atoi(argv[1])) — the runtime flags value
+	// forces the compiler to emit __open_2 under _FORTIFY_SOURCE. It prints
+	// the first byte of config as a decimal so the test can assert 0xde.
+	src := filepath.Join(t.TempDir(), "fortopen.c")
+	const prog = `#include <fcntl.h>
+#include <unistd.h>
+#include <stdio.h>
+#include <stdlib.h>
+int main(int argc, char **argv) {
```
roma-glushko, line 126: Why this check can't be done directly from Golang?

roma-glushko (dismissed):

pkg/system/mockpcisysfs/shim/shim_test.go
giuliocalzo: Good question — this can't be done from Go. Go's `os.Open`/`syscall.Openat` issue `openat(2)` directly and never call glibc, so an `LD_PRELOAD` libc shim never sees them. And `__open_2` is compiler-generated: glibc only rewrites `open(path, flags)` → `__open_2` under `-D_FORTIFY_SOURCE` + `-O` with non-constant `flags` (cgo doesn't enable that). The C helper reproduces exactly what `libpci`/`lspci` do at runtime, and the test skips when `cc` is absent.

pkg/system/mockpcisysfs/shim/shim_test.go
```diff
@@ -80,6 +80,76 @@ func TestOpenSysDevicesPCIRedirect(t *testing.T) {
...
 }
 
+// TestFortifiedOpenPCIRedirect guards the _FORTIFY_SOURCE open path. When a
+// caller opens a file with a non-constant flags argument (as libpci does for a
+// PCI `config` file), glibc's fortify headers emit __open_2 instead of open.
+// If the shim only interposes open/openat, that call escapes redirection and
+// hits the real host path — the exact bug behind lspci's "Cannot open
+// .../config" warning. This compiles a tiny helper that forces __open_2 and
+// asserts it reads the redirected mock config.
+func TestFortifiedOpenPCIRedirect(t *testing.T) {
```
ArangoGutierrez, line 90: Nice regression guard, but it only runs under `-tags integration` with the shim prebuilt, and CI's `go test $(go list ./...)` passes no such tag. `shim/` has no untagged `.go` file, so the package is dropped from the CI run entirely, which means the fortified-open fix (and `fopen`/`fopen64`, which this test does not exercise) currently have no automated enforcement. A small linux CI job would close it: `make -C pkg/system/mockpcisysfs && go test -tags integration ./pkg/system/mockpcisysfs/...`.

ArangoGutierrez (commented):
Clean fix. The fortified-open interposition is the right call, and the identity unpacking, config-space header, and NVIDIA-vendor fallback all check out against the NVML packing and h100.yaml. One non-blocking note on CI coverage below.

1. Nice regression guard, but it only runs under `-tags integration` with the shim prebuilt, and CI's `go test $(go list ./...)` passes no such tag. `shim/` has no untagged `.go` file, so the package is dropped from the CI run entirely, which means the fortified-open fix (and `fopen`/`fopen64`, which this test does not exercise) currently have no automated enforcement. A small linux CI job would close it: `make -C pkg/system/mockpcisysfs && go test -tags integration ./pkg/system/mockpcisysfs/...`. (pkg/system/mockpcisysfs/shim/shim_test.go:90)

pkg/system/mockpcisysfs/shim/shim_test.go
giuliocalzo: Good call — added a linux CI step in `golang.yaml` that builds the shim and runs `go test -tags integration ./pkg/system/mockpcisysfs/...`, so the fortified-open and fopen paths are now enforced. Also added `TestFopenPCIRedirect` to cover the `fopen`/`fopen64` hooks (small C consumer, since Go never calls fopen). Kept it scoped to `mockpcisysfs`; the same integration-tag gap exists for `mockib` and is best handled as a follow-up. (8ff58e8)

Issue #438, opened 2026-07-07
Stage shared libs + RPATH for relocated IB CLI tools in the overlay


ArangoGutierrez:
Surfaced during PR #433 review triage (below inline-comment threshold, but real).

setup.sh copies dynamically-linked IB CLI tools (ibstat, iblinkinfo, ibv_devinfo, ...) into the overlay for injected containers, but does not stage their shared-library dependencies (libibverbs, libibumad, ...) or /etc/libibverbs.d, and the injected LD_LIBRARY_PATH only covers driver/usr/lib64 (NVML/CUDA). The tools work in ubuntu:22.04 demo images only because the image happens to ship the libs; on minimal images (alpine, distroless) they fail with missing-.so errors, and nothing in CI exercises them in injected containers.

Apply the nvidia-smi treatment: stage the required libs into the overlay and patchelf the tools with RPATH `$ORIGIN/../lib64` (plus /etc/libibverbs.d staging), so the binaries are self-locating across mount contexts.

Acceptance: IB CLI tools run inside an NRI-injected minimal-image container; covered by the NRI e2e job.

PR #479, opened 2026-07-20
feat(nvml-mock): node-level mock InfiniBand/NIC device surface (#478)

### LINKED ISSUE #478: [Feature]: Node-level mock InfiniBand/NIC device surface for host-level consumers (Tier 2)
## Problem statement

`nvml-mock` renders its mock InfiniBand/RDMA device surface **only inside pods**
(the per-container `MOCK_IB_ROOT` sysfs tree plus `LD_PRELOAD` shims injected by
the `nvml-mock-nri` plugin). This is enough for libc-based userspace tools
running in an injected pod (e.g. `ibstat`, `ibv_devinfo`), but it does **not**
present any device surface at the **node level**.

As a result, host-level consumers cannot see the mock NICs:

- **Node Feature Discovery (NFD)** scans the node's real `/sys/bus/pci`, which
  contains no `15b3` (Mellanox) devices — so it never derives the NIC feature
  labels the NVIDIA Network Operator keys on. Today the
  `docs/demo/network-operator` demo works around this by stamping
  `feature.node.kubernetes.io/pci-15b3.present=true` onto the nodes directly.
- **RDMA tooling / device plugins** (e.g. the operator's
  `rdma-shared-device-plugin`) open real device nodes under `/dev/infiniband/*`
  and read `/sys/class/infiniband` from the host / their own (NRI-excluded)
  containers — none of which exist on a mock node.
- Anything **scanning PCI** on the node (rather than trusting a stamped label)
  cannot resolve a NIC.

This is the InfiniBand/NIC analogue of the existing GPU node-level PCI rendering
(`render-pci-sysfs`) and the `pci-10de.present` labeling in `setup.sh`: the GPU
path already has a node-level surface, but the NIC/IB path does not.

This gap blocks the mock from exercising the *real* node-level detection and
device-plugin code paths of the NVIDIA Network Operator (and similar RDMA
consumers) without hardware.

## Proposed solution

Extend `nvml-mock` to render a **node-level** mock InfiniBand/NIC device surface
(gated on a profile having InfiniBand enabled), mirroring the existing GPU
node-level PCI rendering. Three sub-capabilities:

1. **Node-level `/sys/class/infiniband` tree** — render the mock IB sysfs on the
   node (e.g. via a `hostPath` the mock DaemonSet writes and operator
   DaemonSets can read), not just inside injected pods.
2. **Node-level `/dev/infiniband/*` character devices** — provide the char
   device nodes RDMA tooling opens (`uverbsN`, `rdma_cm`, `umadN`, …) so plugins
   that `open()` real device nodes can proceed.
3. **PCI NIC entries in the node's PCI view** — render `15b3` devices into the
   node's mock PCI sysfs (analogous to GPU `render-pci-sysfs`), so anything
   scanning PCI — not only consumers of the stamped `pci-15b3.present` label —
   resolves the NICs.

Alternatives considered:

- *Keep pod-only rendering + stamped labels* (status quo): cheap, but cannot
  exercise real host-level detection/device-plugin paths.
- *Real OFED/SR-IOV on Kind*: not viable — needs kernel modules Kind's shared
  kernel cannot provide.

## Scope / non-goals

- **In scope:** node-level rendering of the IB sysfs tree, `/dev/infiniband/*`
  char devices, and `15b3` PCI entries, gated behind an InfiniBand-enabled
  profile; wiring through the Helm chart / `setup.sh` following existing GPU
  patterns.
- **Non-goals:** a real RDMA kernel subsystem (rdma netlink) — Kind's kernel
  does not expose it, so the `rdma-shared-device-plugin` will still crash-loop;
  building OFED/DOCA kernel modules; functional RDMA data-path traffic.

## Acceptance criteria

- [ ] On a node running an InfiniBand-enabled profile, `/sys/class/infiniband`
      exists at the **node level** and lists the mock HCAs (matching the count
      the profile renders inside pods).
- [ ] The mock `/dev/infiniband/` directory exists on the node with the expected
      char device nodes (`uverbsN`, `rdma_cm`, `umadN`) for each rendered HCA.
- [ ] The node's mock PCI view contains `15b3` (Mellanox) device entries, such
      that a PCI scan of the node resolves the NIC(s) **without** relying on the
      pre-stamped `feature.node.kubernetes.io/pci-15b3.present` label.
- [ ] All of the above is gated on the profile having InfiniBand enabled and is
      absent for GPU-only profiles (no `15b3`/IB surface leaks onto non-IB
      profiles).
- [ ] Rendering is idempotent across DaemonSet restarts, and a corresponding
      cleanup removes the node-level IB sysfs, `/dev/infiniband/*`, and `15b3`
      PCI entries.
- [ ] The `docs/demo/network-operator` demo no longer needs to pre-stamp
      `pci-15b3.present` in `kind.yaml`: NFD derives the NIC label from the
      node-level PCI surface on its own. (Documented in the demo README.)
- [ ] Docs updated (`nvml-mock` chart README + relevant `docs/`) describing the
      node-level IB/NIC surface and its limits (still no rdma netlink / kernel
      driver).

## Notes / references

- GPU precedent: node-level PCI rendering (`render-pci-sysfs`) and
  `pci-10de.present` labeling in `deployments/nvml-mock/scripts/setup.sh`.
- Current pod-only IB rendering: `MOCK_IB_ROOT` + `LD_PRELOAD` shims injected by
  `nvml-mock-nri` (`pkg/nri/nvmlmock/adjust.go`).
- Boundary reminder: even with a full node-level surface, the
  `rdma-shared-device-plugin` needs a real RDMA kernel subsystem (rdma netlink)
  that Kind's kernel does not expose — see `docs/demo/network-operator`.

This is **Tier 2** of a larger gap analysis (Tier 1, the node-level NIC label,
is already addressed in the network-operator demo).

giuliocalzo:
## Summary

Adds the **Tier 2 node-level mock InfiniBand/NIC device surface** (#478) to
`nvml-mock`, so host-level consumers (NFD, RDMA tooling, anything scanning PCI)
can resolve the mock Mellanox NICs — not just libc tools inside injected pods.
This is the InfiniBand/NIC analogue of the existing GPU node-level PCI rendering
and `pci-10de.present` labeling.

Delivered on an InfiniBand-enabled profile (gated on `$ibEnabled`):

- **`15b3` NIC PCI entries.** `render-pci-sysfs` now emits attribute-bearing PCI
  devices (`vendor`/`device`/`class`/`subsystem_*`) and synthesizes one Mellanox
  `15b3` NIC per mock HCA from the profile's `infiniband:` block, into the node's
  mock PCI tree (`/var/lib/nvml-mock/sys/...`). It gains `--gpu-count` and parses
  the IB block.
- **Real `/dev/infiniband/*` char devices.** `setup.sh` upgrades the renderer's
  placeholder files to real character devices via `mknod` (`uverbsN`, `umadN`,
  `rdma_cm`, …); the NRI plugin injects them into pods annotated
  `nvml-mock.nvidia.com/devices=true`.
- **NFD `pci-15b3.present` label.** The kernel `/sys` can't be faked, so the
  chart writes an NFD **`local`** source feature file into the node's
  `features.d` directory; NFD reads it and publishes
  `feature.node.kubernetes.io/pci-15b3.present=true` — durable across operator
  reconciles of NFD's DaemonSet. Controlled by `infiniband.nfd.publishNicLabel`
  (**default `true`** on IB profiles) / `infiniband.nfd.featuresDir`.
- **Docs + tests.** Chart README node-level surface + limits; Helm unit tests and
  render snapshots for the NIC entries, NFD wiring, and the disabled path.

## Non-goals (physical constraints on Kind)
- Populating the **real host kernel `/sys`** — that needs a kernel module; the
  mock surface lives in the node-local mock tree, visible to consumers that read
  it (IB `LD_PRELOAD` shims / NRI-injected pods), not the kernel's native `/sys`.
- `rdma-shared-device-plugin` advertising `rdma/*` — it needs a real kernel
  RDMA-netlink subsystem that Kind's shared kernel does not expose.

## Test plan
- [x] `go test -race $(go list ./... | grep -v vendor)` — pass
- [x] `golangci-lint run` — 0 issues
- [x] `helm unittest deployments/nvml-mock/helm/nvml-mock` — 134/134 (15 snapshots)
- [x] Live on Kind: mock PCI tree shows `15b3` NIC entries; NFD publishes
      `pci-15b3.present=true` from the `local` feature file; `/dev/infiniband/*`
      are real char devices.

Closes #478 (achievable subset; non-goals recorded on the issue).


ArangoGutierrez:
Heads up Giulio — #557 lands in the same part of `setup.sh` you touch here, so this will want a rebase once it merges.

It fixes #438: the IB CLI tools were staged into the overlay without their shared libraries, so they failed with `cannot open shared object file` in any NRI-injected pod. Not just minimal images, incidentally — debian does it too, since it doesn't ship the IB stack either. The tools now come pre-staged from the image build with their library closure and an RPATH, and `setup.sh` copies that tree in rather than looping over `command -v`.

Two things worth knowing when you rebase:

The conflict is textual and sits in section 4b. The tool loop is still there but its meaning changed — it now only copies a tool if one isn't already staged, so `ibstatus` (a shell script, no RPATH to give it) still gets picked up while the six ELF tools keep their patched copies. Easier to take main's version of that block and re-apply your `mknod` hunk on top than to merge the loop by hand.

Mechanically your work doesn't overlap: you add the `/dev/infiniband` char devices and the NFD label, #557 adds the libraries the tools need to open them. The `adjust.go` device injection doesn't collide either.

No rush — this is v0.4.0 and #557 isn't merged yet.

Issue #486, opened 2026-07-22
refactor(e2e): Use Tilt to deploy E2E test


roma-glushko:
Use Tilt and the newly added make commands to setup environments for E2E test runs.

Acceptance criteria:

- [ ] make cluster-create is used to provision a new in-runner cluster for E2E tests
- [ ]  Tilt is used to deploy the needed runtime and wait till it's ready
- [ ]  go tests are only responsible for functional cluster validation & brief information collection (decouple environment setup from actual test)

PR #465, opened 2026-07-16
ci(e2e): build nvml-mock image once and share via ttl.sh

giuliocalzo:
## Summary

Closes #460.

Builds the `nvml-mock` E2E image **once** instead of rebuilding it in every matrix leg (`e2e`, `e2e-dra`, `e2e-gpu-operator`, `e2e-multi-node`, `e2e-nri`), and hardens the hand-off between the build and the legs:

- **Build once:** a new `build-image` job builds `nvml-mock` a single time (buildx + GHA cache) and pushes it to the ephemeral, auth-free **ttl.sh** registry as `E2E_IMAGE`. Every leg `needs: build-image` and `docker pull`s it (`E2E_SKIP_BUILD=true`, then kind-loads) instead of rebuilding. ttl.sh is auth-free, so this also works on fork PRs.
- **Digest-pinned pull (integrity):** `build-image` exposes the pushed image digest as a job output, and each leg pulls `ttl.sh/nvml-mock-<sha>@sha256:<digest>` (content-addressed), then re-tags it to `E2E_IMAGE`. This closes the window where a third party could overwrite the sha-derived tag on the anonymous registry between push and pull.
- **24h TTL:** the ttl.sh tag uses a `24h` TTL (ttl.sh max) so re-running a single failed leg within the day still finds the image, instead of hitting an expired `1h` tag and being forced into a full rebuild.
- `tests/e2e/README.md` "CI Behavior" updated to match.

No Go harness or `Makefile` changes are needed: the leg re-tags the digest-pinned image back to `E2E_IMAGE`, which already flows through both `kind load docker-image` and the Helm `image.repository`/`image.tag` split.

_The digest-pinning and 24h-TTL items above address review feedback from @ArangoGutierrez._

## Timing comparison

Comparing a full old per-leg-build run against this PR's build-once + digest-pull run (both `success`, 25 profile legs each):

| Metric | Old — per-leg build ([#458](https://github.com/NVIDIA/k8s-test-infra/actions/runs/29486704680)) | New — build-once + digest ([#465](https://github.com/NVIDIA/k8s-test-infra/actions/runs/29487761864)) | Δ |
|---|---:|---:|---:|
| **Total runner-minutes** (legs + build-image) | 186.3 min | 58.1 min | **−128 min (−69%)** |
| Per-leg avg duration | 447 s | 138 s | −309 s (−69%) |
| Per-leg min / max | 334 s / 592 s | 104 s / 187 s | much faster |
[...]

PR #492, opened 2026-07-22
refactor(e2e): Move cluster setup to tilt in E2E tests (p1)

roma-glushko:
## What This PR Does

Exploring a way to leverage Tilt to do cluster setup in E2E tests.

The PR touches only one the `e2e` stage right now. We will propagate the changes to other stages & simplify ginkgo code in the followup PRs.

Also, rename the namespace to `mokka` to reflect the project new name.

Related to #486 

As a nice side effect, the `e2e` tests seem to start finishing quicker (generally shed ~1m from almost each `e2e` run):

Screenshot 2026-07-24 at 11 49 01

Before:

Screenshot 2026-07-24 at 11 49 33

## Why

This helps to have one, unified configuration for envs in CI and locally which means less configurations to think about, easier way to reproduce any CI related failures or simply avoid them because the env is close to the local setup.

## Checklist


tests/e2e/ci/nvml-mock.values.yaml
```diff
@@ -0,0 +1,44 @@
+# Copyright 2026 NVIDIA CORPORATION
```
roma-glushko, line 1: Extracted this override from ginkgo code.

.github/workflows/nvml-mock-e2e-go.yaml
```diff
@@ -88,11 +99,45 @@ jobs:
...
+          cache-to: type=gha,mode=max,scope=nvml-mock
+
+  # Build the custom kind-node image (kindest/node + nvidia-container-toolkit)
+  # once per commit for the standalone `e2e` pilot. Same ttl.sh distribution
+  # pattern as `build-image` above — the pilot leg pulls by digest and retags
+  # locally to kind-node-nv:latest, then `make cluster-create` sees it and
+  # skips the rebuild (KIND_NODE_IMAGE_PREBUILT=1 in the env). Runs in
+  # parallel with `build-image`; independent scope on the GHA cache so the
+  # two jobs don't contend for the same key.
+  build-kind-node-image:
```
roma-glushko, line 112: Build a kind node image with CDE and containerd setup for all other tests (built in parallel with nvml-mock).

.github/workflows/nvml-mock-e2e-go.yaml
```diff
@@ -128,11 +163,58 @@ jobs:
...
+      - name: Roll out nvml-mock via Tilt (${{ matrix.gpu-profile }})
+        # `tilt ci` runs the Tiltfile headless: waits for every resource to
+        # become ready, then exits 0/1. --nvmlmock-image tells Tilt to skip
+        # docker_build and pin image.repository/tag directly to the ttl.sh
+        # ref the build-image job published. Kubelet inside kind pulls it
+        # over the docker-bridge egress; ttl.sh is auth-free so no
+        # `docker pull` / `kind load` dance is needed. The sha-scoped URL
+        # (`ttl.sh/nvml-mock-<sha>:24h`) makes tag collisions effectively
+        # impossible for this commit.
+        run: tilt ci -- --gpu-profile="${{ matrix.gpu-profile }}" --nvmlmock-image="$E2E_IMAGE"
```
roma-glushko, line 209: The main heavylifting - tilt deploy the needed configuration of the cluster and waits until the system is ready. One single command rules them all.

ArangoGutierrez (commented):
Correcting this review: my previous body was posted in error (wrong content from another PR). The intended review of this change is below.

Two things worth a look on an otherwise clean refactor — the namespace rename is consistent and attach mode reads correctly.

1. `.github/workflows/nvml-mock-e2e-go.yaml:168` — The Install Tilt step pipes a curl download straight into `sudo tar` with no integrity check, while every Install Kind step in this same file verifies KIND_AMD64_SHA256 with `sha256sum --check`. A swapped or MITM'd release asset would land in /usr/local/bin as root unverified. Worth pinning a TILT_SHA256 and checking it before extraction — the same control the kind download already uses (and that PR #442 restored).

2. `Tiltfile:59` — This comment says the ref may be `repo:tag` or `repo@digest`, but `_split_image` in local/nvml_mock.tiltfile hard-fails on any `@` ref since the chart only supports repo:tag. A user who follows this and passes a digest hits an immediate fail(). Drop the `repo@digest` mention here (the matching comment at local/nvml_mock.tiltfile:16-17 has the same wording).

.github/workflows/nvml-mock-e2e-go.yaml
ArangoGutierrez, line 209: This is now the only leg that doesn't pin the nvml-mock image by digest. Tilt gets the `ttl.sh/nvml-mock-<sha>:24h` tag, which renders into the DaemonSet as-is with `IfNotPresent`, and since nothing kind-loads it any more the kubelet pulls that tag from ttl.sh at pod start. #465 added the digest pull specifically to close that window on an anonymous registry, and the kind-node step just above still does it. Keeping the digest pull plus a `kind load` before `tilt ci` would restore it without needing `image.digest` in the chart.

Makefile
```diff
@@ -111,8 +111,18 @@ KIND_CLUSTER_NAME   ?= $(if $(filter compute-domain,$(PROFILE)),nvml-mock-comput
...
 .PHONY: image-kind-node cluster-create cluster-delete
+# KIND_NODE_IMAGE_PREBUILT (env, any non-empty value): skip the local docker
+# build and trust that $(KIND_NODE_IMAGE) is already present in the local
+# daemon. CI sets this after `docker pull` + `docker tag` of the pre-built
+# image from ttl.sh (see .github/workflows/nvml-mock-e2e-go.yaml build-kind-node
+# job); local devs leave it unset and get the rebuild-when-Dockerfile-changes
+# behavior.
 image-kind-node:
-	@docker build -t $(KIND_NODE_IMAGE) ./local/kind
+	@if [ -n "$$KIND_NODE_IMAGE_PREBUILT" ]; then \
```
ArangoGutierrez, line 123: The skip branch trusts that `$(KIND_NODE_IMAGE)` is already in the local daemon but never checks, so a staging step that failed or tagged something else shows up later as an opaque `kind create cluster` pull error. A `docker image inspect $(KIND_NODE_IMAGE) >/dev/null` in that branch would fail where the assumption actually broke.

local/README.md
```diff
@@ -17,7 +17,7 @@ This directory contains Tiltfiles, Helm value overrides, and Kind cluster config
 make cluster-create
 
 # 2. Start the dev stack (default: a100 profile, homogeneous fleet)
-tilt up
+tilt up -- --gpu-operator
```
ArangoGutierrez, line 20: The quick start now runs the GPU Operator variant, but the line above still describes it as the default a100 homogeneous fleet and `--gpu-operator` has its own section further down. Reads like a stray local change.

ArangoGutierrez (dismissed):
Approving. The attach-mode split reads well — Tilt owning the rollout and Ginkgo only observing is the right seam, and the runtime win is real. Three non-blocking notes:

1. This is now the only leg that doesn't pin the nvml-mock image by digest. Tilt gets the `ttl.sh/nvml-mock-<sha>:24h` tag, which renders into the DaemonSet as-is with `IfNotPresent`, and since nothing kind-loads it any more the kubelet pulls that tag from ttl.sh at pod start. #465 added the digest pull specifically to close that window on an anonymous registry, and the kind-node step just above still does it. Keeping the digest pull plus a `kind load` before `tilt ci` would restore it without needing `image.digest` in the chart. (.github/workflows/nvml-mock-e2e-go.yaml:207)
2. The skip branch trusts that `$(KIND_NODE_IMAGE)` is already in the local daemon but never checks, so a staging step that failed or tagged something else shows up later as an opaque `kind create cluster` pull error. A `docker image inspect $(KIND_NODE_IMAGE) >/dev/null` in that branch would fail where the assumption actually broke. (Makefile:121)
3. The quick start now runs the GPU Operator variant, but the line above still describes it as the default a100 homogeneous fleet and `--gpu-operator` has its own section further down. Reads like a stray local change. (local/README.md:20)

.github/workflows/nvml-mock-e2e-go.yaml
roma-glushko: Added the kind load step back in https://github.com/NVIDIA/k8s-test-infra/pull/492/commits/d0f99623aa7ee91a3f82df44f61ad40719a44db6 ✅ 

Thank you!

Makefile
roma-glushko: Added the image existence check in https://github.com/NVIDIA/k8s-test-infra/pull/492/commits/d0f99623aa7ee91a3f82df44f61ad40719a44db6 ✅ 

Thank you!

local/README.md
roma-glushko: Updated the comment in https://github.com/NVIDIA/k8s-test-infra/pull/492/commits/d0f99623aa7ee91a3f82df44f61ad40719a44db6 ✅ 

Thank you!

ArangoGutierrez (commented):
The e2e matrix isn't running on the current head, and it's failing in a way the PR UI hides.

The merge of main in 5b606eb7 brought in the new `e2e-nfd` job, which depends on `build-image` — the job this PR renames to `build-nvmlmock-image`. Git merged both changes without a textual conflict, but the resulting job graph points at a job that no longer exists, so Actions rejects the workflow before it schedules anything.

The reason it still looks green: a workflow that fails validation never creates check-runs, so it leaves no red mark anywhere. The PR currently reads 5/5 passing, MERGEABLE, CLEAN — while the 26 e2e jobs that were green on d0f99623 simply aren't running. The two failed runs show up only in the Actions tab under the `pull-request/492` branch.

The fix is the two references inside `e2e-nfd` (lines 445 and 475). Worth confirming the full matrix comes back green afterwards, since that matrix is what this PR is changing.

Everything else reads well — the three earlier comments are addressed, the CI values file is wired correctly through the Tiltfile overlay stack, and no chart templates or values.yaml are touched.

.github/workflows/nvml-mock-e2e-go.yaml
```diff
@@ -64,7 +75,7 @@ env:
 jobs:
   # Build nvml-mock once and publish to ttl.sh; all legs consume this artifact
   # instead of rebuilding the image per job (see issue #460).
-  build-image:
+  build-nvmlmock-image:
```
ArangoGutierrez, line 78: Renaming this job orphans the `e2e-nfd` leg that arrived with the main merge. `e2e-nfd` still declares `needs: build-image` (line 445) and reads `needs.build-image.outputs.digest` (line 475), and there is no `build-image` job any more.

Actions rejects the whole workflow at validation, so nothing schedules. And because `ci.yaml` calls this file via `uses:`, that run dies with it — both completed with zero jobs:

- https://github.com/NVIDIA/k8s-test-infra/actions/runs/30353747066
- https://github.com/NVIDIA/k8s-test-infra/actions/runs/30353752841

Pointing both references at `build-nvmlmock-image` should be all it needs. (GitHub won't let me anchor a comment on 445 itself — those lines aren't in the diff, since they match main.)

roma-glushko: Should be addressed in https://github.com/NVIDIA/k8s-test-infra/pull/492/commits/6e91c86380c2412a79582861215a83ebb3b5c2ba ✅

ArangoGutierrez: Almost there — 445 is fixed, but the same job still has the old reference one step further down, at line 475:

```
pinned="ttl.sh/nvml-mock-${{ github.sha }}@${{ needs.build-image.outputs.digest }}"
```

`build-image` isn't in `e2e-nfd`'s `needs:` any more, so that interpolates to an empty string rather than erroring. `e2e-nfd (a100)` just failed on it:

```
pinned="ttl.sh/nvml-mock-6e91c86380c2412a79582861215a83ebb3b5c2ba@"
docker pull "$pinned"
invalid reference format
```

https://github.com/NVIDIA/k8s-test-infra/actions/runs/30369893461/job/90312838241

Changing it to `needs.build-nvmlmock-image.outputs.digest` should be the last of it.

Good news is the graph fix worked — the workflow schedules again, which is why the rest of the matrix is running now.

roma-glushko: Yes! Always forget about that. 
Fixed in https://github.com/NVIDIA/k8s-test-infra/pull/492/commits/8a78ae739585fc57a83ba37ee9cb70aa763da4e2 ✅ 

Waiting for the CI to get green.

roma-glushko: @ArangoGutierrez NFD test suite is green now ✅

ArangoGutierrez (approved):
Green across the board now — 36/36 including the nfd leg. Thanks for the quick turnaround on both job-reference fixes.

One trivial thing for whenever you're next in this file: the line-20 header comment picked up a doubled token from the rename sweep (`build-nvmlmock-imagebuild-nvmlmock-image`). Comment-only, nothing references it.

Issue #441, opened 2026-07-07
tracking: NRI node-wide injection — harden and extend (v0.3.0 flagship)


ArangoGutierrez:
PR #433 moves mock delivery from per-pod-spec wiring to the container runtime layer: an NRI plugin intercepts CreateContainer and injects the overlay mount, env, and device nodes into every container on the node — the same layer the real stack (nvidia-container-toolkit / CDI) operates at. Workloads become testable unmodified, and the injection contract lives in one place (pkg/nri/nvmlmock/adjust.go) instead of being copied into every manifest and template.

This epic tracks productionizing that foundation and building on it. It is the flagship item for the v0.3.0 release.

Hardening:
- [ ] #434 NRI plugin failure-mode hardening (fail-closed / fail-open)
- [ ] #435 e2e: NRI injection CI job with injection self-check
- [ ] #438 Stage shared libs + RPATH for relocated IB CLI tools in the overlay
- [ ] #439 nvml-mock-nri: unit tests, unix.Major/Minor, stale LD_PRELOAD comment

Extending:
- [ ] #436 Inject CDI device references through the NRI plugin
- [ ] #437 IMEX mock on the NRI foundation (channels, daemon, ComputeDomain demo)
- [ ] #440 Compose mock device plugin advertisement with NRI delivery

Known trade-offs of the direction, accepted for test infrastructure: node-level blast radius (root DaemonSet rewriting container specs — never deploy outside test clusters), containerd/CRI-O coupling with NRI enablement as a prerequisite, and the vendored nri/ttrpc surface. Scheduling advertisement is intentionally out of scope for the plugin itself (#440 covers composition with the device plugin).

ArangoGutierrez:
v0.3.0-rc1 status for this epic.

Three of the seven children are done and ship in the rc:

- #439 — nvml-mock-nri has unit coverage, the full Linux `dev_t` decode is fixed, and the stale LD_PRELOAD comment is gone (PR #539).
- #434 — the plugin now detects a wedge and a silent unregistration and recovers from both. The probes are mutation-checked: revert them and the spec goes red (PR #540).
- #435 — the e2e NRI job runs across all six GPU profiles with an injection self-check.

Four are still open and move to v0.3.0 final:

- #436 CDI device references through the plugin. Unblocked now that #464 landed the dedicated Kind node image with nvidia-container-toolkit and CDI.
[...]

PR #493, opened 2026-07-23
docs(demo): NVSentinel thermal-margin detect/remediate/recover demo on mock GPUs

giuliocalzo:
## Summary
- Adds a self-contained NVSentinel demo on Kind that drives a **GPU thermal-margin violation** on mock GPUs and shows the full detect -> cordon -> drain -> cooldown -> auto-uncordon loop.
- Extends `nvml-mock` to expose the signed GPU **T.Limit threshold fields** (NVML 193-196) and a **signed margin temperature**, so DCGM's `GpuThermalMarginWatch` arms and alarms like real Ada+ hardware.
- Recovery is automatic on cooldown — no DCGM restart needed (unlike a latched XID/ECC fault).

## What's included
- `docs/demo/nv-sentinel/`: `run.sh`, `README.md`, `kind.yaml`, GPU Operator / NVSentinel / MongoDB values, and a sample GPU workload.
- `pkg/gpu/mocknvml`: `FieldValueInt` (signed-int) support through the cgo bridge, T.Limit threshold field resolution (fields 193-196), and unclamped (signed) margin temperature.

## Test plan
- [x] `go test ./pkg/gpu/mocknvml/engine/...`
- [x] `go build ./pkg/gpu/mocknvml/bridge/...` (cgo)
- [x] `golangci-lint run` on the changed packages (0 issues)
- [x] `bash -n docs/demo/nv-sentinel/run.sh`
- [x] Manual end-to-end on Kind: heat GPU 0 -> negative margin -> `GPU_TEMP_HW_SLOWDOWN_VIOLATION` -> node cordoned + workload drained; clear override -> margin recovers -> node auto-uncordoned.


lalitadithya:
this is great stuff!! Thank you for putting this together!

ArangoGutierrez (commented):
No issues found — checked for bugs and CLAUDE.md compliance.

avasilevskii (approved):

docs/demo/nv-sentinel/README.md
roma-glushko, line 1: If NVSentinel is a legitimate consumer of mokka like GPU operator or Topograph, it makes sense to support it as a tilt-deployable component in `./local` directory.

giuliocalzo: I can create the tiltfile as well but I'm not sure we can simulate the exec command to validate the thermal output and the related nvsentinel actions...

roma-glushko: @giuliocalzo the compute domain setup goes with two jobs that do fabric check and reconfiguration. Do we need something like that here?

docs/demo/nv-sentinel/run.sh
```diff
@@ -0,0 +1,360 @@
...
+
+  Inspect further:
+    kubectl --context ${KUBE_CONTEXT} -n ${NVSENTINEL_NAMESPACE} get pods
+    kubectl --context ${KUBE_CONTEXT} get nodes
+    kubectl --context ${KUBE_CONTEXT} -n ${NVSENTINEL_NAMESPACE} logs -l app.kubernetes.io/instance=nvsentinel --prefix | grep -iE 'cordon'
+
+  Re-run the fault manually:
+    MOCK=\$(kubectl --context ${KUBE_CONTEXT} -n ${NVML_MOCK_NAMESPACE} get pod -l app.kubernetes.io/name=nvml-mock --field-selector spec.nodeName=${TARGET_NODE} -o jsonpath='{.items[0].metadata.name}')
+    kubectl --context ${KUBE_CONTEXT} -n ${NVML_MOCK_NAMESPACE} exec \$MOCK -- nvml-mock-ctl temp --gpu ${TARGET_GPU} ${HOT_TEMP_C}   # heat -> cordon
+    kubectl --context ${KUBE_CONTEXT} -n ${NVML_MOCK_NAMESPACE} exec \$MOCK -- nvml-mock-ctl reset --gpu ${TARGET_GPU}               # cool -> auto-uncordon
```
roma-glushko, line 356: @giuliocalzo I'm generally not sure how this is going to work outside of the local kind cluster with a handful of nodes. Are we proposing that test suites like one that @avasilevskii is working on to ssh into a every single node to do the runtime information modification of simulated gpu? Would this be convenient to do for even 5 thousands nodes?

I don't think this approach is going to scale. I understand that this is the only way to do it with the current architecture, but maybe we should zoom out a bit and actually do the proper adjustment to the architecture in order to provide the best UX for our users?

ArangoGutierrez: Fair concern, but I think it lands on the next artifact rather than this one, and two different mechanisms are getting conflated here.

Demo mechanism: one kubectl exec into the nvml-mock DaemonSet pod writes overrides.yaml into that node's hostPath config dir, and every consumer on the node picks it up within a second, so there is no ssh anywhere. The part of run.sh that genuinely does touch every node is the container-toolkit and CDI setup loop up at the top, and that is exactly what the NRI node-wide injection work in #441 removes.

Scale mechanism: you are right that we do not have one. The override file is already the correct reconciliation target though, so what is missing is delivery, not an engine change. The shape I want is a watch in the DaemonSet driven by a small CR, so one operator action fans out to N nodes instead of N execs.

I would rather design that against the suite @avasilevskii is building than block a demo doc on it, so I opened #533 to scope it properly: CRD plus controller, agent-applied, with campaign semantics and real status so we can answer "did this land on all 4,997 nodes". Broken into 11 issues, and your point is why it exists, so please tear into the design one (#522).

This PR is the demo mechanism, and as a demo it works and is fully green.

docs/demo/nv-sentinel/README.md
ArangoGutierrez: On compute domain: not needed here. This demo never creates one. It is h100 thermal margin on a single GPU with cordon, drain and uncordon, and nothing in that loop touches the fabric, so the fabric check and reconfiguration jobs would add setup time and failure surface for signal we do not assert.

ArangoGutierrez (approved):
Approving this as a demo doc.

Roma's scaling point is a real one, but it applies to the test suite rather than to this PR, so I have split it out into #533 instead of holding this up. Short version: the exec here is into the nvml-mock DaemonSet pod, not ssh, and the part of run.sh that genuinely touches every node is the container-toolkit and CDI setup loop that #441 is already removing. The scale answer is a CRD plus an agent that reconciles overrides.yaml, which is additive since the engine already reads that file declaratively.

Worth saying that this is not only docs. The signed thermal margin, the T.Limit field values 193-196, and populating busIdLegacy are what let NVSentinel's metadata-collector and GpuThermalMarginWatch work against the mock at all, and the tests that come with them are good ones. Full matrix is green.

One small thing for whenever you are back, not blocking and I have not filed it yet: writeBusID only writes the NUL when the string is shorter than the destination, so a 16-char bus ID into BusIdLegacy [16]int8 comes out unterminated. You clearly saw this coming since you added the 12-char legacy sentinel for exactly that reason, it is just the config path that has no guard. Every shipped profile uses 12-char IDs so nothing is broken today.

Thanks for this one, the detect and recover loop on mock GPUs is a nice result.

docs/demo/nv-sentinel/README.md
roma-glushko: I meant that compute domain example shares the same problem (and has a solution) as Giulio was curious how to approach with tilt:

> but I'm not sure we can simulate the exec command to validate the thermal output and the related nvsentinel actions...

docs/demo/nv-sentinel/run.sh
roma-glushko: For posterity, I've submitted [MEP0001](https://github.com/NVIDIA/k8s-test-infra/pull/536) that proposes the architectural changes that would help us to solve this problem and a couple of others mentioned in the document. Ultimately, the goal here is to make cluster admins live as easy as possible while they deploy and operate Mokka.

Issue #502, opened 2026-07-25
Mokka pre-silicon coverage gaps (AICR preflight POC)


ArangoGutierrez:
## Summary

Tracking issue for every Mokka capability gap found by the Mokka + AICR pre-silicon integration preflight POC (2026-07-25). One surface for the roadmap this POC produced.

Context: the POC ran the AICR `validate` suite against an `nvml-mock` `gb200` cluster to answer whether Mokka + AICR materially reduces GB200 bring-up time, or merely proves the stack deploys against mocked APIs.

## Headline coverage

Of **21** AICR checks: **A 14**, **B 0**, **C 4**, **G 3**.

- **Today: 66.7%** carries real pre-silicon signal (A / total).
- **Reachable: 81.0%** once the gaps below close ((A + G) / total). Roadmap, not a current claim.
- **Mokka specifically unlocks 42.9%** (9 / 21). The other 5 bucket-A checks are control-plane checks any cluster would run.
- **9 of 21 executed** in this run: 5 pass, 4 fail, 12 not run.

## The gaps

- [ ] **#498, nvml-mock does not ship the mock IMEX surface the DRA compute-domain plugin needs.** Size S (corrected down from M). Blocks `dra-support` and `slinky-slurm-imex-channel`, so it unlocks 2 checks from G to A. **Highest leverage, and the only gap on the GB200 critical path**, since ComputeDomains is how IMEX is expressed in DRA. Corrected 2026-07-25: the DRA driver already ships the consumption mechanism (`altProcDevices` / `ALT_PROC_DEVICES_PATH`) and its CI builds this surface against nvml-mock, so this is a packaging gap rather than a capability gap. See #498 for the full correction.
- [ ] **#499 — no simulated node provisioner.** Size M. Blocks `cluster-autoscaling`, unlocking 1 check. This is the one gap explicitly about composing Mokka's vertical depth with KWOK's horizontal breadth. Note it would prove the autoscaling control loop only, **not** GPU-stack fidelity on the provisioned KWOK nodes.
- [x] **#500 — GFD e2e assertion is warning-only and can never fail.** Size S. **Closed in the POC PR.** Does not move a bucket: it was a defect in our own e2e rather than in Mokka's capability, so it is a test-quality fix, not coverage.

Related, not a capability gap: **#501 — source real GB200 NVL72 bring-up failure history**. Blocks the back-test, needs a human rather than engineering, and is the single highest-value unblock for a defensible time-saved number.

[...]

Issue #501, opened 2026-07-25
Source real GB200 NVL72 bring-up failure history for the AICR preflight back-test


ArangoGutierrez:
## Summary

The Mokka + AICR pre-silicon preflight POC (2026-07-25) could not complete its back-test because **no captured DSX-OS GB200 NVL72 bring-up failure set exists in any searchable location**. This issue tracks sourcing it.

The back-test is the POC's central metric: for each real failure observed during an actual bring-up, determine whether an AICR check under Mokka would have caught it pre-silicon. Without real failure data there is no defensible time-saved number, and the bring-up-time claim stays qualitative.

**No failure set was synthesized to fill the gap.** That was deliberate.

## Where we looked, and why it came up empty

Searched: `NVIDIA/aicr` (docs, roadmap, changelog, git log, issues), `NVIDIA/k8s-test-infra` (docs, issues), and internal planning notes. All negative for observed NVL72 bring-up records.

The structural reason: DSX-OS configs live on **internal GitLab**, not GitHub, so real bring-up artifacts are not reachable from any repository in scope. Jira was the highest-value system not searched.

One important distinction that came out of the search: the "what breaks at scale" material in internal planning notes is **anticipated**, not observed. It is the specification of a report from a 5K-node run that has not happened yet, and one note says outright that multi-rack IMEX breakage is predicted rather than observed. It must not be used as a failure set.

## What is needed

A record of what actually broke during a specific real GB200 NVL72 bring-up, with enough specificity to back-test: what failed, at which stage, and ideally when and where. A runbook alone is not sufficient; the failure log is the artifact.

## Leads, ranked

1. **Andre Prado (aprado@) and the DSX Sw Bringup org.** Strongest lead. An exec thread on repaves noted runbooks that take manual effort and days, with Andrew Morse already on the thread, so a warm intro exists.
2. **The DGXC Capacity Bring-up thread.** An org measured on bring-up hours tracks what consumes them. Owner not yet identified.
3. **Aleksei Vasilevskii.** Named internally as the fastest unblock; an internal install runbook request from 2026-07-22 is still outstanding. Caveat: may hold the runbook rather than a failure log, and is on other work for ~2 weeks.

## Interim state

[...]

Issue #500, opened 2026-07-25
gap: GPU Feature Discovery e2e assertion is warning-only and can never fail


ArangoGutierrez:
## Summary

The GPU Operator e2e scenario's GPU Feature Discovery check could never fail. GFD could have been removed entirely and the scenario would still have gone green. Found during the Mokka + AICR pre-silicon integration preflight POC (2026-07-25).

**Status: fixed in the preflight POC PR.** Filed for the record and as the tracking surface, since it is a test-quality regression class worth naming.

## The problem

`tests/e2e/go/scenario_gpu_operator_test.go` asserted three GFD labels through `assertions.NodeLabelSoft`:

```go
for _, label := range []string{"nvidia.com/gpu.product", "nvidia.com/gpu.memory", "nvidia.com/gpu.count"} {
    assertions.NodeLabelSoft(ctx, h.Kube, node, label)
}
```

`NodeLabelSoft` (`tests/e2e/go/assertions/resources.go:116-126`) only prints:

```go
case !ok || v == "":
    _, _ = fmt.Fprintf(ginkgo.GinkgoWriter, "WARNING: label %s not set on %s (soft)\n", key, node)
```

There is no `gomega` assertion in any branch. The check cannot fail.

## Why it matters for coverage

GFD is the component that turns mock NVML device data into node labels, which is one of the "resource advertisement" surfaces a pre-silicon preflight is supposed to cover. A check that cannot fail contributes no signal, which makes it a textbook bucket-B row: green regardless of correctness.

This also inflates confidence in the AICR checks that sit downstream of node labelling.

## The fix

Assert the labels **hard**, with values derived from the profile rather than read back off the node, so the comparison actually discriminates:

- `profile.GFDProductName()` renders `device_defaults.name` the way GFD publishes it (spaces become dashes: `NVIDIA GB200` becomes `NVIDIA-GB200`).
- `profile.MemoryMiB()` converts `device_defaults.memory.total_bytes` to the MiB value GFD publishes.
- `assertions.ExpectedGFDLabels(product, memoryMiB, gpuCount)` builds the expected set.
- `assertions.WaitGFDLabels(...)` polls until the node matches and fails the spec otherwise, reporting every mismatching label rather than stopping at the first.

Deriving the expected values independently is what makes this real. Reading the labels off the node and asserting they equal themselves would be the same theater in a different shape.

## Verified against a live cluster

[...]

Issue #499, opened 2026-07-25
gap: no simulated node provisioner, so AICR cluster-autoscaling has nothing to exercise


ArangoGutierrez:
## Summary

Mokka composes with no simulated node provisioner, so AICR's `cluster-autoscaling` conformance check has nothing to exercise and skips. This is the one gap in the preflight catalog that is explicitly about composing Mokka's vertical depth with KWOK's horizontal breadth.

Found during the Mokka + AICR pre-silicon integration preflight POC (2026-07-25).

## Which AICR check this blocks

| AICR check | What it would validate | Current bucket | Bucket if closed |
|---|---|---|---|
| `cluster-autoscaling` | Karpenter behavioural scale-out driven by an HPA and a GPU NodePool, or an EKS node group, or the GKE cluster autoscaler | G | A |

## Why this is bucket G and not bucket C

It is not hardware-dependent. AICR's own CI already exercises the Karpenter path against **KWOK-faked** nodes: `.github/actions/gpu-validate-conformance/action.yml:38-43` runs `install-karpenter-kwok` before `aicr validate --phase conformance`. So the capability is known to work without real cloud scale-out; this harness simply does not stand it up.

`validators/conformance/cluster_autoscaling_check.go:61-67` skips gracefully when no mechanism is detectable, which is what happens on a plain Kind cluster.

## What Mokka does today vs what it would need

**Today.** Mokka runs on real nodes with a real kubelet, because it replaces `libnvidia-ml.so.1` on the host. Node count is bounded by how many real nodes or VMs you have.

**Needed.** A composed topology where KWOK provides cheap hollow nodes for the provisioner to scale into, while real nodes run Mokka for GPU-stack depth. Concretely: install `karpenter-kwok` alongside the Mokka release, expose a GPU NodePool whose nodes carry `nvidia.com/gpu.present=true` and the `kwok.x-k8s.io/node=fake:NoSchedule` toleration contract, and let the check's behavioural pods drive scale-out.

## Important caveat on what this would and would not prove

Closing this gap proves the **autoscaling control loop** reacts correctly to GPU-shaped pending pods. It does **not** prove GPU-stack fidelity on the nodes that get provisioned, because those are KWOK nodes with no kubelet and no driver. Deep GPU-stack fidelity at KWOK-scale node counts remains unproven and must not be claimed on the strength of this work.

That boundary is the whole reason this is filed separately rather than folded into a scale story.

## Implementation sketch

[...]

PR #503, opened 2026-07-25
poc: Mokka + AICR pre-silicon integration preflight

ArangoGutierrez:
## The question this answers

Agreed with @mchmarny on 2026-07-24:

> Does Mokka + AICR **materially reduce GB200 bring-up time**, or does it **merely prove the stack deploys against mocked APIs**?

An honest negative result was an acceptable outcome. Nothing here was tuned toward a favourable answer.

## Answer

It does more than prove deployment against mocked APIs, but by less than the headline number suggests, and the strongest results are **two real integration defects found pre-silicon** rather than a green suite. **Verdict: go with scope**, on a narrower claim than the original pitch.

The single most persuasive evidence is not the coverage percentage. It is that coverage improved monotonically as components were added, and each fix moved a check to its *next* real requirement rather than flipping it green:

| Cluster state | pass / fail | What moved |
|---|---|---|
| GPU Operator + DRA driver | 5 / 4 | baseline |
| + cert-manager, NFD, kube-prometheus-stack | 5 / 4 | `platform-health` 8 missing namespaces to 5; `ai-service-metrics` moves to "no `DCGM_FI_DEV_GPU_UTIL` time series" |
| + dcgm-exporter ServiceMonitor | 5 / 4 | moves to "custom metrics API not available" |
| + prometheus-adapter | **6 / 3** | **`ai-service-metrics` passes** |

A check that passed trivially against a mock could not have failed four different ways for four different real reasons.

## Headline numbers

Real run, 2026-07-25, provenance `sim`. 1-node Kind cluster, Kubernetes v1.36.1, linux/arm64 (correct for GB200, since Grace is ARM), `nvml-mock` `gb200` profile, GPU Operator with ClusterPolicy `ready`, `nvidia-dra-driver-gpu` v25.12.0, AICR at `06256cc8`.

Of **21** AICR checks: **A 14 · B 0 · C 4 · G 3**

| Metric | Value |
|---|---|
| **Meaningful pre-silicon TODAY** | **66.7%** (A / total) |
| **Reachable once tracked gaps close** | **81.0%** ((A + G) / total). Roadmap, not a current claim. |
| **What Mokka specifically unlocks** | **42.9%** (9 / 21) |
| **Actually executed in this run** | **9 of 21** (6 pass, 3 fail). 12 not run. |

Two qualifications that cut against the headline, stated up front because they matter more than the percentage:

1. **Only 9 of the 14 bucket-A checks depend on the GPU stack.** The other 5 are control-plane checks any cluster would run; `gang-scheduling` is deliberately CPU-only upstream. Counting them as coverage Mokka bought would inflate the number, so the harness tracks `gpuDependent` separately.
2. **12 of 21 checks did not execute.** Their buckets are analysis, not observation.

## The most useful result was a failure

`dra-support` failed. The real DRA driver installs and publishes all 8 simulated GB200 devices as ResourceSlices, but the check also wants a `nvidia-dra-driver-gpu-controller` Deployment, which chart v25.12.0 templates only when `resources.computeDomains.enabled=true`. Enabling that crashes the kubelet plugin:

```
error getting nvcap for IMEX channel '0': error getting device major:
error parsing '/proc/devices': unexpected regex match: []
```

**This took two corrections to get right, and the second one matters.**

The first reading was that AICR carried a stale expectation. Checking the chart templates disproved it: the controller is gated on `resources.computeDomains.enabled`.

The second reading was that Mokka lacks an `nvidia-caps-imex-channels` device class and that closing it meant inventing a way to fake `/proc/devices`, sized M. That was also wrong. The DRA driver **already** supports mock NVML: its chart exposes `altProcDevices`, whose documented example path is literally `/var/lib/nvml-mock/imex/proc-devices`, and its CI builds the whole surface against `nvml-mock` in `hack/ci/mock-nvml/setup-mock-gpu.sh`.

So the real gap is **packaging, not capability**, and it is smaller (S, not M): `nvml-mock` does not ship that surface, and the released NGC chart v25.12.0 this repo's e2e uses has no `altProcDevices` key. Both halves exist; neither is reachable from the `nvml-mock` chart. Reproducing the surface by hand and mounting it over `/proc/devices` is rejected by runc, which is exactly why the env-var indirection exists.

Buckets and coverage are unchanged, because a user of the `nvml-mock` chart alone still cannot run the check. Tracked as #498, rewritten.

On the positive side, `check-nvidia-smi` passed, exercising device-plugin allocation, CDI injection, container-toolkit wiring, and `dlopen` resolution of `libnvidia-ml.so.1`. That is exactly the chain that broke in #455, where three missing `//export` symbols made `RTLD_NOW` abort while build, unit tests, and lint all stayed green.

## A defect in AICR's own recipe

@mchmarny this one is for you specifically, and it is not filed anywhere.

The generated `kind` + `gb200` recipe **selects `ai-service-metrics` as a conformance check and declares both `kube-prometheus-stack` and `prometheus-adapter` as components, but never wires the GPU Operator's dcgm-exporter into Prometheus.** No GPU time series is ever produced, so the check cannot pass from a clean deploy of its own bundle.

Verified three ways against the generated recipe, since one negative grep is not evidence:

1. The `gpu-operator` component's `dcgmExporter` overrides contain only `config: {create: false, data: '', name: ''}`. No `serviceMonitor` block, and the chart default is disabled. On the live cluster no `nvidia-dcgm-exporter` ServiceMonitor existed until enabled by hand.
2. The `kube-prometheus-stack` overrides touch only alertmanager, defaultRules, grafana, prometheus (resources, retention, storage) and prometheusOperator. No `additionalScrapeConfigs`, no widened `serviceMonitorSelector`.
3. DCGM is mentioned exactly twice in the recipe, both inside that gpu-operator block. No scrape configuration anywhere.

Prometheus selects ServiceMonitors labelled `release: kube-prometheus-stack`; with none created at all, no selector helps. Setting `dcgmExporter.serviceMonitor.enabled=true` with that label made `DCGM_FI_DEV_GPU_UTIL` appear (carrying `DCGM_FI_DRIVER_VERSION: 580.65.06` and the GB200 model name off the mock), and adding prometheus-adapter made the check pass.

**I have not opened an issue on NVIDIA/aicr.** It is your repo, and you may know a deployment path that configures this which the generated recipe simply does not express. If it is real, say so and I will file it properly.

## Back-test: caught 2 of 4, on a proxy set

**The primary back-test is pending.** No captured DSX-OS GB200 NVL72 bring-up failure set exists in any searchable location, because DSX-OS configs live on internal GitLab. **No failure set was synthesized.** Tracked as #501, which needs a human rather than engineering.

A clearly-relabelled **secondary proxy** ran against real hardware-encountered failures on **cloud** GB200 (`p6e-gb200.36xlarge`) from NVIDIA/aicr issues:

| Failure | Caught? | By what |
|---|---|---|
| aicr#859 prereq ConfigMap ordering, deploy hangs | **caught** | `operator-health`, `expected-resources` |
| aicr#1255 K8s floor below the DRA `resource.k8s.io/v1` requirement | **caught** | readiness version constraint, `dra-support` |
| aicr#1553 `nvidia_uvm` wedged on driver-managed nodes | **missed** | kernel module state, bucket C permanently |
| aicr#1861 validate phase outlives the AWS token | **missed** | cloud IAM lifetime, out of simulation scope |

Both misses are explained by category, not by a fixable gap. **Four caveats apply**: cloud GB200 rather than an NVL72 rack, n=4, two of the four are failures of AICR's own tooling caught by AICR's own validators (partly circular), and a devil's-advocate review recommended against running any proxy at all. That objection is recorded in the findings note because you may share it.

## Scale caveat

The mock replaces `libnvidia-ml.so.1` on the host and needs a real kubelet, so **it cannot run on KWOK nodes**. This POC ran on **one node**. It says nothing about GPU-stack fidelity at KWOK-scale node counts. That combination is a claim to prove, not to assert, and nothing here asserts it.

## Claims supported, and not

| Claim | Supported? | Basis |
|---|---|---|
| Pre-silicon **integration preflight** for the A-bucket checks | **Yes, scoped** | 6 passed and 3 failed for real, diagnosable reasons; coverage improved monotonically as components were added |
| **Reduces** time to first token | **Directionally, not quantified** | no bring-up timeline data exists to attach hours to |
| **Hit** first token on day one | **No** | simulation cannot prove it |
| AICR-on-silicon remains the final readiness gate | **Yes** | bucket C is non-empty by construction |
| Deep GPU-stack fidelity at KWOK scale | **No** | not tested; the mock cannot run on KWOK nodes |
| Mokka validates CUDA execution | **No** | 1 CUDA driver-API symbol; cuda-validation runs `WITH_WORKLOAD=false` |

## Gap backlog

Summary issue: **#502**. Gaps filed: **3**. Closed in this PR: **1**.

| Gap | Issue | Unlocks | Size | Status |
|---|---|---|---|---|
| `nvml-mock` does not ship the mock IMEX surface the DRA compute-domain plugin needs | #498 | 2 checks, G to A | S | open |
| No simulated node provisioner | #499 | 1 check, G to A | M | open |
| GFD e2e assertion warning-only, can never fail | #500 | test quality | S | **closed here** |

#500 is worth a look on its own: `NodeLabelSoft` only printed a warning, so GFD could have been deleted entirely and the GPU Operator scenario would still have gone green. It now asserts hard against values derived from the profile. It did **not** move a bucket, because it was a defect in our e2e rather than in Mokka, and it is reported that way.

**Deliberately not filed: MIG.** Coverage is zero, but no AICR check requires it, so closing it would unlock nothing. Recorded so nobody refiles it as a blocker.

## Diagrams

<details>
<summary>Stack layering: AICR is the workload, Mokka is the substrate</summary>

```mermaid
flowchart TB
  L5["AICR: recipes, validate and conformance suite (runs unmodified)"]
  L4["GPU stack: GPU Operator, DRA driver, device plugin, GFD, DCGM"]
  L3["kubelet + containerd, CDI injection"]
  L2["Mokka Sim: libnvidia-ml.so.1 replacement, device nodes, CDI spec -- SIMULATION BOUNDARY"]
  L1["Real CPU nodes: Kubernetes, real kubelet, real container runtime"]
  L1 --> L2 --> L3 --> L4 --> L5
```

Direction matters: the common misreading is that simulation lives inside AICR. It does not.
</details>

<details>
<summary>Preflight timeline: where integration failures get found</summary>

```mermaid
flowchart LR
  subgraph WITHOUT["WITHOUT preflight"]
    W1["~3 months: engineers idle"] --> W2["silicon arrives"] --> W3["integration failures found HERE"] --> W4["hardware-dependent validation"]
  end
  subgraph WITH["WITH preflight"]
    P1["~3 months: AICR on Mokka, integration failures found and fixed"] --> P2["silicon arrives"] --> P3["hardware-dependent validation only. AICR on silicon = FINAL GATE"]
  end
```

No day-one first-token claim is drawn, because simulation does not prove it.
</details>

Five diagrams live in [`docs/aicr-preflight/diagrams/`](docs/aicr-preflight/diagrams/), each as mermaid plus a hand-authored SVG for email and slides.

## Review ask

@mchmarny, the part I most want your eye on is **[`docs/aicr-preflight/`](docs/aicr-preflight/)**, specifically the AICR description and where I have drawn the final-gate boundary. Those are yours to validate and I would rather be corrected now than in front of a customer.

Two specific calls I would like you to check:

1. **`slinky-slurm-imex-channel` is bucketed G, not C.** My reasoning: whether the *fabric carries traffic* is hardware-dependent and stays C (that is what `nccl-all-reduce-bw-nvls` is for), but whether *channel allocation hands out distinct channels* is driver-surface plumbing and is simulatable. This is the closest call in the catalog and sits nearest your boundary.
2. **Whether the cloud-GB200 proxy back-test should exist at all.** It is labelled and caveated, and the objection to it is recorded, but you may still want it removed.

## How to re-run

```bash
make e2e-gpu-operator E2E_PROFILES=gb200 E2E_KEEP_CLUSTER=true
```

```bash
aicr recipe --service kind --accelerator gb200 --intent inference -o recipe.yaml && aicr validate --recipe recipe.yaml --phase deployment --phase conformance --fail-on-error=false --output ctrf/validate.json
```

```bash
go run ./tests/aicr-preflight -ctrf ctrf -markdown coverage.md -json report.json -provenance sim
```

Full prerequisites in [`tests/aicr-preflight/README.md`](tests/aicr-preflight/README.md). Raw run output is committed under [`results/2026-07-25-gb200-kind/`](tests/aicr-preflight/results/2026-07-25-gb200-kind/).

## Design note

The harness is built so an incomplete run degrades to "not run" rather than to a favourable number: a check the run never reported can never be recorded as passing, CTRF `pending` and `other` map to not-run, a bucket-C check reporting green under `sim` is flagged suspect, and every record carries provenance. Both invariants are mutation-tested. The catalog carries the analytical judgment with source citations; the harness supplies only the observed outcome and never edits a bucket.

## Verification

`go build ./...`, `go test ./...`, `go test -tags=e2e` on the touched packages, and `golangci-lint run ./...` all pass (0 issues). All commits are DCO signed-off and GPG signed.

Not for merge as-is: this is a POC for review, and the back-test section is openly pending.

### DIFF
--- a/cmd/render-imex-procdevices/main.go
+++ b/cmd/render-imex-procdevices/main.go
@@ -1,0 +1,73 @@
+// Copyright 2026 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+// render-imex-procdevices writes a substitute /proc/devices carrying NVIDIA
+// caps character-device entries, for consumption by the NVIDIA DRA driver's
+// compute-domain kubelet plugin via ALT_PROC_DEVICES_PATH.
+//
+// On a Mokka node there is no NVIDIA kernel module, so /proc/devices has no
+// nvidia-caps-imex-channels entry and the plugin aborts at startup. The DRA
+// driver's chart exposes an altProcDevices value for exactly this case; this
+// command produces the file it points at.
+//
+// Usage:
+//
+//	render-imex-procdevices \
+//	    --source /proc/devices \
+//	    --output /var/lib/nvml-mock/imex/proc-devices \
+//	    --imex-major 235 --caps-major 236
+//
+// Rendering is idempotent, so the DaemonSet may re-run it on restart.
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/NVIDIA/k8s-test-infra/pkg/system/mockimex/render"
+)
+
+func main() {
+	if err := run(os.Args[1:]); err != nil {
+		fmt.Fprintf(os.Stderr, "render-imex-procdevices: %v\n", err)
+		os.Exit(1)
+	}
+}
+
+func run(args []string) error {
+	fs := flag.NewFlagSet("render-imex-procdevices", flag.ContinueOnError)
+	source := fs.String("source", "/proc/devices", "path to the real /proc/devices to derive from")
+	output := fs.String("output", "", "path to write the rendered file (required)")
+	imexMajor := fs.Int("imex-major", 235, "device major to advertise for "+render.IMEXChannelsDeviceName)
+	capsMajor := fs.Int("caps-major", 236, "device major to advertise for "+render.CapsDeviceName)
+
+	if err := fs.Parse(args); err != nil {
+		return err
+	}
+	if *output == "" {
+		return fmt.Errorf("--output is required")
+	}
+
+	data, err := os.ReadFile(*source) //nolint:gosec // operator-supplied source path
+	if err != nil {
+		return fmt.Errorf("read %s: %w", *source, err)
+	}
+
+	rendered, err := render.ProcDevices(string(data), *imexMajor, *capsMajor)
+	if err != nil {
+		return fmt.Errorf("render from %s: %w", *source, err)
+	}
+
+	if err := os.MkdirAll(filepath.Dir(*output), 0o755); err != nil {
+		return fmt.Errorf("create output directory for %s: %w", *output, err)
+	}
+	if err := os.WriteFile(*output, []byte(rendered), 0o644); err != nil { //nolint:gosec // world-readable by design: consumed by another container
+		return fmt.Errorf("write %s: %w", *output, err)
+	}
+
+	fmt.Printf("wrote %s (%s major=%d, %s major=%d)\n",
+		*output, render.IMEXChannelsDeviceName, *imexMajor, render.CapsDeviceName, *capsMajor)
+	return nil
+}
--- a/deployments/nvml-mock/Dockerfile
+++ b/deployments/nvml-mock/Dockerfile
@@ -23,6 +23,7 @@ COPY pkg/gpu/mockcuda/ pkg/gpu/mockcuda/
 COPY pkg/network/mockib/ pkg/network/mockib/
 COPY pkg/nri/ pkg/nri/
 COPY pkg/system/mockpcisysfs/ pkg/system/mockpcisysfs/
+COPY pkg/system/mockimex/ pkg/system/mockimex/
 COPY pkg/imexcoord/ pkg/imexcoord/
 COPY pkg/fmcoord/ pkg/fmcoord/
 COPY cmd/mock-ib/ cmd/mock-ib/
--- a/deployments/nvml-mock/Dockerfile
+++ b/deployments/nvml-mock/Dockerfile
@@ -29,6 +29,7 @@ COPY cmd/mock-ib/ cmd/mock-ib/
 COPY cmd/nvml-mock-nri/ cmd/nvml-mock-nri/
 COPY cmd/nvml-mock-ctl/ cmd/nvml-mock-ctl/
 COPY cmd/render-pci-sysfs/ cmd/render-pci-sysfs/
+COPY cmd/render-imex-procdevices/ cmd/render-imex-procdevices/
 COPY cmd/fake-imex/ cmd/fake-imex/
 COPY cmd/fake-fabricmanager/ cmd/fake-fabricmanager/
 COPY cmd/check-fabric/ cmd/check-fabric/
@@ -41,6 +42,7 @@ RUN mkdir -p /out \
     && go build -mod=vendor -o /out/nvml-mock-nri ./cmd/nvml-mock-nri \
     && go build -mod=vendor -o /out/nvml-mock-ctl ./cmd/nvml-mock-ctl \
     && go build -mod=vendor -o /out/render-pci-sysfs ./cmd/render-pci-sysfs \
+    && go build -mod=vendor -o /out/render-imex-procdevices ./cmd/render-imex-procdevices \
     && go build -mod=vendor -o /out/nvidia-imex     ./cmd/fake-imex/daemon \
     && go build -mod=vendor -o /out/nvidia-imex-ctl ./cmd/fake-imex/ctl \
     && go build -mod=vendor -o /out/nv-fabricmanager     ./cmd/fake-fabricmanager/daemon \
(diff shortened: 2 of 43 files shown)


ArangoGutierrez:
Superseded by the re-run on `poc/mokka-aicr-kind-sweep`; a replacement PR will follow.

Closing rather than rebasing, and the reasoning for the record:

- The code from this PR is already on `main`. #541 and #542 landed, and the IMEX renderer this PR introduced at `pkg/system/mockimex/render/` now lives at `internal/imex/` after #729 relocated it.
- The two branches write 11 of the same paths under `docs/aicr-preflight/`, so leaving both open guarantees a conflict for whoever merges second.
- This PR's results are from 2026-07-25, which predates the `v0.3.0` tag. The new branch carries two dated runs (`v0.3.0` on 2026-08-03, and `upstream/main` @ `57ef016` on 2026-08-26). Adding a third, non-comparable set of numbers would make the coverage figures harder to read rather than easier.

Carried forward: the `preflight-workflow` diagram and the diagrams index, which existed only here.

Left behind deliberately: `tests/aicr-preflight/` and its 2026-07-25 results. That harness will not be in the tree, so those numbers could not be regenerated or checked against anything. They stay reachable at `pull/503/head`.

Issue #487, opened 2026-07-22
[Feature]: Deploy Topograph in the local kind clusters


roma-glushko:
### Problem statement

We want to be able to have a quick way to test whether we are simulating cross-rack network topology without issues

### Proposed solution

- deploy Topograph chart as another optional component via Tilt

### Audience

Operators / cluster admins

### Pre-submission checks

- [x] I searched existing issues / Discussions and this is not a duplicate.
- [x] I understand triage cadence is weekly and Unprioritized SLA is 7 calendric days.

roma-glushko:
While thinking about I've realized that:
- there is no way to realistically simulate rack-scale GPU distribution. [MEP0001](https://github.com/NVIDIA/k8s-test-infra/pull/536) is a proposal to solve this.
- [Topograph](https://github.com/giuliocalzo/topograph) uses cloud-specific API to resolve topology (eg. AWS, GCP, OCI, Nebius, Lambda, etc.) and two flavours of IB (Bare Metal & K8s). There is a question whether we need to be able to fullfil all those profiles.

A separate questions:
- how would cluster admins know what should be their cross-rack topology to simulate?
- whether it makes sense in the first place to do any simulation of cross-rack networking? Seems like it may be helpful for testing of topology-aware workload scheduling at least.

roma-glushko:
Closed in https://github.com/NVIDIA/k8s-test-infra/pull/518 ✅

PR #518, opened 2026-07-27
docs(demo): #487 Support Topograph deployment

roma-glushko:
## What This PR Does

Deploy Topograph as an optional component inside the existing local kind clusters. Topograph is configred to use the `inifinibandk8s` topology provider.

## Why

Topograph is an important part of the AI infrastructure and does K8s-native cross rack topology discovery. Since we emulate  parts of that, we need to be able to deploy it locally for testing purposes.

## Testing

Topograph deployments in the local cluster:
Screenshot 2026-07-24 at 17 05 00

Topograph nodes:
Screenshot 2026-07-24 at 17 04 57

Node-level labels added by Topograph:
Screenshot 2026-07-24 at 17 25 06

## Checklist


local/compute-domain/scenarios/check-fabric.sh
```diff
@@ -12,7 +12,7 @@
 
 set -euo pipefail
 
-CLUSTER_NAME="nvml-mock-compute-domain"
+CLUSTER_NAME="mokka-compute-domain"
```
ArangoGutierrez: Renaming CLUSTER_NAME is not enough here: the asserts still build ${CLUSTER_NAME}-worker[2-4], but the kind config now registers the nodes as worker-0..worker-3, and pod_on_node filters on spec.nodeName - so nothing matches and each of the four assertions burns a 30s poll before failing. The expected values are stale too: the new topology.yaml puts worker-1 in clique 1 and moves worker-2/worker-3 to domain ...cd, which a single EXPECTED_UUID cannot express. topology-rebind.sh carries the same four node names and needs the same fix.

local/compute-domain/rebind.topology.yaml
```diff
@@ -15,7 +15,7 @@ topology:
       cliques:
         - id: 99
           nodes:
-            - nvml-mock-compute-domain-worker
-            - nvml-mock-compute-domain-worker2
-            - nvml-mock-compute-domain-worker3
-            - nvml-mock-compute-domain-worker4
+            - mokka-compute-domain-worker
```
ArangoGutierrez: These four names were cluster-renamed but not node-renamed, so they match nothing now that the nodes register as worker-0..worker-3. applyTopologyOverlay keys on NODE_NAME and no-ops on a miss, so the helm upgrade succeeds, nothing actually rebinds, and topology-rebind.sh fails its clique-99 assertions with a misleading message. The header comment that node names match topology.yaml is no longer true either.

local/kind/compute-domain.kind.yaml
```diff
@@ -47,15 +47,43 @@ nodes:
...
+      nvidia.com/gpu.clique: "00000000-0000-0000-0000-0000000000ab.0"
+    kubeadmConfigPatches:
+      - |
+        kind: JoinConfiguration
+        nodeRegistration:
+          name: worker-0
   - role: worker
     labels:
       nvml-mock/clique: "0"
+      nvidia.com/gpu.clique: "00000000-0000-0000-0000-0000000000ab.0"
```
ArangoGutierrez: The static nvidia.com/gpu.clique labels disagree with the new topology.yaml on three of four workers: topology gives worker-1 clique 1, worker-2 domain ...cd clique 0 and worker-3 domain ...cd clique 1, while the labels say ab.0, ab.1, ab.1. Since these labels are exactly what topograph reads, the demo would emit the old two-clique layout rather than the 2-spine/4-leaf one the new diagram advertises - and topology.yaml:15 asks for this alignment explicitly. The nvml-mock/clique values drift the same way.

local/topograph/dra-driver.values.yaml
```diff
@@ -0,0 +1,21 @@
...
+#
+# Flips gpuCliqueLabelEnabled=true so the compute-domain-kubelet-plugin
+# writes the nvidia.com/gpu.clique node label — the sole surface
+# topograph's dra provider reads. Chart default is false (see
+# tmp/dra-driver-nvidia-gpu/deployments/helm/dra-driver-nvidia-gpu/
+# values.yaml:307). Label writer implementation:
+# tmp/dra-driver-nvidia-gpu/cmd/compute-domain-kubelet-plugin/
+# computedomain.go:429-453 (SetGPUCliqueLabel — MergePatch on nodes,
+# 10-minute refresh loop).
+kubeletPlugin:
```
ArangoGutierrez: Nothing loads this file - the DRA block only appends the compute-domain overlay, with no with_topograph branch. Given useGpuCliqueLabel reads the static label instead of the device-plugin path, deleting it looks like the right call rather than wiring it up; it also cites tmp/dra-driver-nvidia-gpu/... paths that are not in the repo, and it describes topograph's 'dra' provider while this setup uses infiniband-k8s.

Tiltfile
```diff
@@ -53,6 +54,8 @@ config.define_bool('dra', args=False,
     usage='Also deploy NVIDIA DRA driver on top of nvml-mock')
 config.define_bool('fgo', args=False,
     usage='Also deploy Run:ai Fake GPU Operator (combine with --multi-gpu-profile to exercise both integration and scale pools)')
+config.define_bool('topograph', args=False,
+    usage='Also deploy NVIDIA topograph. Implies --compute-domain --dra (needed to produce the nvidia.com/gpu.clique labels topograph reads). Still requires the compute-domain Kind cluster: make cluster-create PROFILE=compute-domain.')
```
ArangoGutierrez: The usage string promises 'Implies --compute-domain --dra', but the block below only sets with_compute_domain, and the README added in this PR says --dra is explicitly not required. This is the string tilt up -- --help prints, so it is the one users see. topograph.values.yaml repeats the claim on line 8, and line 19 credits the label to the DRA compute-domain-kubelet-plugin, contradicting its own prerequisites block twelve lines up.

local/kind/compute-domain.kind.yaml
```diff
@@ -28,7 +28,7 @@
 #   tilt up -- --compute-domain --k8s-context=kind-nvml-mock-compute-domain
```
ArangoGutierrez, line 28: A few references still carry the old names. This usage line passes --k8s-context=kind-nvml-mock-compute-domain and line 9 describes name: nvml-mock-compute-domain, three lines above the new name: mokka-compute-domain - and allow_k8s_contexts now only permits kind-mokka-compute-domain, so copy-pasting this fails. The local/README.md profile table still lists kind-gpu-test and kind-nvml-mock-compute-domain, and the Makefile comment above KIND_CLUSTER_NAME still justifies the separate cluster name by saying topology.yaml hardcodes <cluster-name>-worker[N], which this PR just stopped doing.

local/compute-domain/compute_domain.tiltfile
```diff
@@ -105,7 +105,7 @@ def build_images(with_dra):
 def install(active_consumers):
     """Install the nvml-mock helm release with the compute-domain topology.
 
-    Returns ['nvml-mock'] so downstream consumers (dra, gpu-operator)
+    Returns ['mokka'] so downstream consumers (dra, gpu-operator)
```
ArangoGutierrez: The docstring now says the function returns ['mokka'], but the last line still returns ['nvml-mock'] and the module header at line 22 also still says ['nvml-mock']. The return value is the right one - it has to match the helm_resource name, and labels=['mokka'] below is only a Tilt UI grouping - so the docstring is the part to revert.

local/compute-domain/topology.yaml
```diff
@@ -1,29 +1,56 @@
...
-#   domain "demo-domain"
-#     clique 0: -worker, -worker2
-#     clique 1: -worker3, -worker4
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Keep this aligned with the nvidia.com/gpu.clique labels in kind-cluster.yaml.
```
ArangoGutierrez: The old header said this file is a copy of docs/demo/compute-domain/topology.yaml kept in lockstep, and dropping that note is what let the two silently diverge - the demo copy still has one domain and the nvml-mock-compute-domain-* names. This line also points at kind-cluster.yaml, which is not a file here; the labels live in local/kind/compute-domain.kind.yaml.

ArangoGutierrez (commented):
The topograph wiring itself checks out - I pulled chart 0.5.0 and global.provider.params.useGpuCliqueLabel is real and does read the static node label, so the README is right that --dra is not needed. What needs another pass is the node rename: worker-0..worker-3 did not reach the two scenario scripts, the rebind fixture, or the clique labels, so both compute-domain scenarios fail as-is.

1. [must-fix] Renaming CLUSTER_NAME is not enough here: the asserts still build ${CLUSTER_NAME}-worker[2-4], but the kind config now registers the nodes as worker-0..worker-3, and pod_on_node filters on spec.nodeName - so nothing matches and each of the four assertions burns a 30s poll before failing. The expected values are stale too: the new topology.yaml puts worker-1 in clique 1 and moves worker-2/worker-3 to domain ...cd, which a single EXPECTED_UUID cannot express. topology-rebind.sh carries the same four node names and needs the same fix. (local/compute-domain/scenarios/check-fabric.sh:15; also local/compute-domain/scenarios/topology-rebind.sh:15)
2. [must-fix] These four names were cluster-renamed but not node-renamed, so they match nothing now that the nodes register as worker-0..worker-3. applyTopologyOverlay keys on NODE_NAME and no-ops on a miss, so the helm upgrade succeeds, nothing actually rebinds, and topology-rebind.sh fails its clique-99 assertions with a misleading message. The header comment that node names match topology.yaml is no longer true either. (local/compute-domain/rebind.topology.yaml:18)
3. [must-fix] The static nvidia.com/gpu.clique labels disagree with the new topology.yaml on three of four workers: topology gives worker-1 clique 1, worker-2 domain ...cd clique 0 and worker-3 domain ...cd clique 1, while the labels say ab.0, ab.1, ab.1. Since these labels are exactly what topograph reads, the demo would emit the old two-clique layout rather than the 2-spine/4-leaf one the new diagram advertises - and topology.yaml:15 asks for this alignment explicitly. The nvml-mock/clique values drift the same way. (local/kind/compute-domain.kind.yaml:66)
4. [should-fix] Nothing loads this file - the DRA block only appends the compute-domain overlay, with no with_topograph branch. Given useGpuCliqueLabel reads the static label instead of the device-plugin path, deleting it looks like the right call rather than wiring it up; it also cites tmp/dra-driver-nvidia-gpu/... paths that are not in the repo, and it describes topograph's 'dra' provider while this setup uses infiniband-k8s. (local/topograph/dra-driver.values.yaml:18)
5. [should-fix] The usage string promises 'Implies --compute-domain --dra', but the block below only sets with_compute_domain, and the README added in this PR says --dra is explicitly not required. This is the string tilt up -- --help prints, so it is the one users see. topograph.values.yaml repeats the claim on line 8, and line 19 credits the label to the DRA compute-domain-kubelet-plugin, contradicting its own prerequisites block twelve lines up. (Tiltfile:58; also local/topograph/topograph.values.yaml:19)
6. [should-fix] A few references still carry the old names. This usage line passes --k8s-context=kind-nvml-mock-compute-domain and line 9 describes name: nvml-mock-compute-domain, three lines above the new name: mokka-compute-domain - and allow_k8s_contexts now only permits kind-mokka-compute-domain, so copy-pasting this fails. The local/README.md profile table still lists kind-gpu-test and kind-nvml-mock-compute-domain, and the Makefile comment above KIND_CLUSTER_NAME still justifies the separate cluster name by saying topology.yaml hardcodes <cluster-name>-worker[N], which this PR just stopped doing. (local/kind/compute-domain.kind.yaml:28; also local/README.md:31, Makefile:110)
7. [should-fix] The docstring now says the function returns ['mokka'], but the last line still returns ['nvml-mock'] and the module header at line 22 also still says ['nvml-mock']. The return value is the right one - it has to match the helm_resource name, and labels=['mokka'] below is only a Tilt UI grouping - so the docstring is the part to revert. (local/compute-domain/compute_domain.tiltfile:108)
8. [should-fix] The old header said this file is a copy of docs/demo/compute-domain/topology.yaml kept in lockstep, and dropping that note is what let the two silently diverge - the demo copy still has one domain and the nvml-mock-compute-domain-* names. This line also points at kind-cluster.yaml, which is not a file here; the labels live in local/kind/compute-domain.kind.yaml. (local/compute-domain/topology.yaml:15)
9. [consider] kubeadm reads InitConfiguration.nodeRegistration.name for the first control plane, so this JoinConfiguration patch is inert. I ran a kind v0.32.0 cluster with these exact patches to check: the worker came up as worker-0, but the control plane stayed <cluster>-control-plane. Nothing references control-0 so it is cosmetic, but switching the kind to InitConfiguration would make it real. (local/kind/compute-domain.kind.yaml:53)

local/compute-domain/scenarios/check-fabric.sh
roma-glushko: The renaming should be fully applied in https://github.com/NVIDIA/k8s-test-infra/pull/518/commits/67518533438abf7be7f4473e617f6cf3b8cfda26 ✅ 

Thanks!

local/compute-domain/rebind.topology.yaml
roma-glushko: The renaming should be fully applied in https://github.com/NVIDIA/k8s-test-infra/pull/518/commits/67518533438abf7be7f4473e617f6cf3b8cfda26 ✅ 

Thanks!

local/kind/compute-domain.kind.yaml
roma-glushko: Updated in https://github.com/NVIDIA/k8s-test-infra/pull/518/commits/b07a90f9dc7306fb4433e4cee6038319cb61a580 ✅ 

Thanks!

local/topograph/dra-driver.values.yaml
roma-glushko: Updated in https://github.com/NVIDIA/k8s-test-infra/pull/518/commits/b07a90f9dc7306fb4433e4cee6038319cb61a580 ✅ 

Thanks!

Tiltfile
roma-glushko: Updated in https://github.com/NVIDIA/k8s-test-infra/pull/518/commits/b07a90f9dc7306fb4433e4cee6038319cb61a580 ✅ 

Thanks!

Issue #456, opened 2026-07-13
Guard the mock library build against undefined NVML symbols


ArangoGutierrez:
## Problem

The nvml-mock shared library is built with `-buildmode=c-shared`, which links without `--no-undefined`. When a go-nvml bump adds new CGo call sites (as v0.13.3-1 did with nvmlDeviceGetAccountingStats_v2, nvmlDeviceGetBBXTimeData_v1, and nvmlSystemGetCPER_v1), the build still succeeds with those symbols left undefined — and the failure only surfaces at e2e, where nvidia-smi's RTLD_NOW dlopen aborts with the misleading message "couldn't find libnvidia-ml.so".

This class of bug is invisible to go build, go vet, go test, and golangci-lint. On #455 it cost a full e2e matrix round (18 failed jobs) to discover; the fix landed in 922df4d6.

## Proposal

Fail at build time instead. Either (or both):

1. After building the library, assert no undefined nvml exports remain:
   ```sh
   test -z "$(nm -D libnvidia-ml.so.* | awk '$1=="U" && $2 ~ /^nvml/')"
   ```
2. Add `-Wl,--no-undefined` to the mock's CGO_LDFLAGS so the link itself fails. Needs validation that the flag plays well with the cgo runtime's expected undefined symbols — option 1 is the narrow, safe check.

A tiny dlopen(RTLD_NOW) smoke probe in the image-build stage would also catch anything symbol-level checks miss.

## References

- #455 (the incident: go-nvml v0.13.3-1 bump)
- Fix commit: 922df4d6 (regenerated stubs + nvml_types.h additions)

PR #535, opened 2026-07-27
fix(mocknvml): honor timeout in nvmlEventSetWait instead of returning immediately

avasilevskii:
## What This PR Does

- `nvmlEventSetWait_v1/_v2` now block for up to the caller's `timeoutms` (re-checking every 100 ms) instead of returning `NVML_ERROR_TIMEOUT` immediately. An already-pending or mid-wait injected Xid is still delivered promptly, `timeoutms=0` stays a non-blocking poll, and a `NULL` payload is now rejected with `NVML_ERROR_INVALID_ARGUMENT`.
- Adds unit tests for the blocking contract (including a poll-count cap so the spin can't be reintroduced) and integration checks that exercise the wait through the built library.
- Builds the integration image via the mock's Makefile instead of a bare `go build`, which had been dropping the `--defsym` aliases — so the unversioned `nvmlEventSetWait` symbol was undefined in the test image although production ships it.
- Fixes `NO_PADDING=1`, which skipped only the two-pass sizing and still compiled in the 14 MiB padding blob, contradicting its documented behavior.
- Updates the READMEs and code comments that described only `_v2` or claimed the wait simply returns `TIMEOUT`.

## Why

Real NVML blocks for the requested timeout, and clients rely on that as their only inter-poll delay — the `nvidia-device-plugin` health monitor and `dcgm-exporter` both run `for { eventSet.Wait(timeout) }` with no sleep of their own. Returning instantly turned that into a busy spin: on GB300 mock cluster, `nvidia-device-plugin` held one of CPU cores saturated on every GPU node.

Fixing it in the mock rather than setting `DP_DISABLE_HEALTHCHECKS=all` keeps the health-check path genuinely exercised, which matters for failure-injection and remediation testing.

## Fix
After deploying the fix, the nvidia-device-plugin CPU utilization went down to almost idle:
image

## Checklist

- [X] CHANGELOG.md updated (if user-facing change)


ArangoGutierrez (commented):
Good fix, and the busy-spin diagnosis is convincing. I mutation-checked the new unit tests locally (reverting waitForDelivery to an immediate return turns three of them red), so those are real guards. Seven things worth a look, mostly around the build and the docs rather than the wait logic itself.

1. `make` is incremental and the `$(LIB_REALNAME)` rule declares no prerequisites, so if a `libnvidia-ml.so.550.54.15` is already in the build context it prints "Nothing to be done for 'all'" and never recompiles — I reproduced that with a stale artifact in place. `.gitignore` hides those files but `.dockerignore` has no `.so` exclusion, so `COPY pkg/ ./pkg/` can carry one in; the sibling production build guards this with `make clean && make` in deployments/nvml-mock/Dockerfile, and the same guard would fit here. (tests/mocknvml/Dockerfile:29)
2. The NO_PADDING fix is right, but the sibling hole is still open: `GO_TAGS_FLAG` is only consumed by `GO_BUILD_CMD`, and the default two-pass branch bypasses it — the probe hardcodes `-tags nopadding` and the final `go build` passes no tags at all. So `make BUILD_TAGS=foo` still silently drops `foo` for any tag that isn't `nopadding`, which is the same class of no-op this hunk fixes. (pkg/gpu/mocknvml/Makefile:49)
3. The mid-wait delivery claim doesn't hold for a client whose only NVML activity is the wait. `ClaimXid` returns nothing until `tripped` is set, and that flips only inside `failureInjector.Tick()`, reached from `ConfigurableDevice.tickFailure()` on device getters — never from `PendingXidEvent`. So the 100 ms re-poll can only observe a trip that another thread's guarded calls caused; for the device-plugin health loop these docs cite, it degenerates to a plain sleep. The same wording is in the helm and failure-injection READMEs, so either scope it to clients that also make device calls, or tick the injector from the poll. (pkg/gpu/mocknvml/README.md:344)
4. Nothing exercises the SUCCESS branch of the exported waits: these subtests run against an uninitialized engine, so `pollPendingXid` always reports false, and the container fixture has no `failure.xid` block to deliver. Stubbing `pollPendingXid` to a constant false leaves the suite green, so the Xid-delivery wiring the PR documents is the one part that isn't guarded. (pkg/gpu/mocknvml/bridge/events_test.go:144)
5. This symbol probe is exactly the right guard for the `--defsym` drop you fixed, but nothing runs it — `tests/mocknvml` isn't referenced by any file under `.github/` or by the root Makefile, so it only executes on a manual `make -C tests/mocknvml test`. Issue #456 is already open for a build-time `nm -D` undefined-symbol check; wiring either that or this harness into CI would make the guard actually guard. (tests/mocknvml/bridge_tests.go:139)
6. `waitForDelivery` degenerates into a busy spin when `pollInterval` is non-positive: the clamp assigns it to `remaining` and `time.Sleep` then returns immediately, which is the exact spin this function exists to prevent. Only the 100 ms const reaches it today, so this is just a one-line floor to keep it safe for any future caller. (pkg/gpu/mocknvml/bridge/events.go:128)
7. The checklist ticks CHANGELOG.md but the diff doesn't touch it. A wait that now blocks instead of returning instantly is user-facing enough for a line under `[Unreleased]` / `### Fixed`, alongside the other fix PRs recorded there. (CHANGELOG.md:47) — https://github.com/NVIDIA/k8s-test-infra/blob/16bd78651c2c67fb37d7e56ca94122ea7868fe3b/CHANGELOG.md#L45-L50

tests/mocknvml/Dockerfile
```diff
@@ -22,11 +22,11 @@ COPY vendor/ ./vendor/
...
 
-# Build the mock NVML library
+# Build via the Makefile, not a bare `go build`: it owns the --defsym aliases
+# for the unversioned entry points (nvmlEventSetWait, ...), the SONAME and the
+# symlinks. nopadding keeps the test image small.
 WORKDIR /build/pkg/gpu/mocknvml
-RUN go build -buildmode=c-shared -o libnvidia-ml.so.550.54.15 ./bridge && \
-    ln -sf libnvidia-ml.so.550.54.15 libnvidia-ml.so.1 && \
-    ln -sf libnvidia-ml.so.1 libnvidia-ml.so
+RUN make LIB_VERSION=550.54.15 BUILD_TAGS=nopadding
```
ArangoGutierrez: `make` is incremental and the `$(LIB_REALNAME)` rule declares no prerequisites, so if a `libnvidia-ml.so.550.54.15` is already in the build context it prints "Nothing to be done for 'all'" and never recompiles — I reproduced that with a stale artifact in place. `.gitignore` hides those files but `.dockerignore` has no `.so` exclusion, so `COPY pkg/ ./pkg/` can carry one in; the sibling production build guards this with `make clean && make` in deployments/nvml-mock/Dockerfile, and the same guard would fit here.

pkg/gpu/mocknvml/Makefile
```diff
@@ -41,9 +42,11 @@ PADDING_BYTES   ?=
...
 BUILD_TAGS      ?=
 
-# Detect `nopadding` in BUILD_TAGS or honor NO_PADDING=1.
-PADDING_DISABLED := $(if $(filter nopadding,$(BUILD_TAGS))$(filter 1,$(NO_PADDING)),1,)
-GO_TAGS_FLAG     := $(if $(strip $(BUILD_TAGS)),-tags '$(strip $(BUILD_TAGS))',)
+# NO_PADDING=1 must add the build tag: that is what excludes padding.go.
+# Skipping only the two-pass sizing still compiles in the 14 MiB default.
+GO_BUILD_TAGS    := $(strip $(BUILD_TAGS) $(if $(filter 1,$(NO_PADDING)),nopadding))
+PADDING_DISABLED := $(if $(filter nopadding,$(GO_BUILD_TAGS)),1,)
+GO_TAGS_FLAG     := $(if $(GO_BUILD_TAGS),-tags '$(GO_BUILD_TAGS)',)
```
ArangoGutierrez, line 49: The NO_PADDING fix is right, but the sibling hole is still open: `GO_TAGS_FLAG` is only consumed by `GO_BUILD_CMD`, and the default two-pass branch bypasses it — the probe hardcodes `-tags nopadding` and the final `go build` passes no tags at all. So `make BUILD_TAGS=foo` still silently drops `foo` for any tag that isn't `nopadding`, which is the same class of no-op this hunk fixes.

pkg/gpu/mocknvml/README.md
```diff
@@ -332,14 +332,22 @@ not recover without a reboot. Per-mode behaviour:
...
+`nvmlEventSetWait_v1`/`nvmlEventSetWait_v2` is called after the device
+trips. Combine it with
 `mode: ecc_uncorrectable` to inject a specific Xid (for example `64` for
 ECC double-bit, `79` for "GPU has fallen off the bus") without taking
 the GPU off the API surface. Real NVML reports each critical Xid exactly
 once per occurrence, so the mock delivers the configured code on the
 first wait and reports `NVML_ERROR_TIMEOUT` (no event) on subsequent
 waits — exactly like real hardware.
 
+Those subsequent waits block for the caller's timeout before returning
```
ArangoGutierrez: The mid-wait delivery claim doesn't hold for a client whose only NVML activity is the wait. `ClaimXid` returns nothing until `tripped` is set, and that flips only inside `failureInjector.Tick()`, reached from `ConfigurableDevice.tickFailure()` on device getters — never from `PendingXidEvent`. So the 100 ms re-poll can only observe a trip that another thread's guarded calls caused; for the device-plugin health loop these docs cite, it degenerates to a plain sleep. The same wording is in the helm and failure-injection READMEs, so either scope it to clients that also make device calls, or tick the injector from the poll.

pkg/gpu/mocknvml/bridge/events_test.go
```diff
@@ -0,0 +1,167 @@
...
+				"NULL event data must be rejected")
+			require.Less(t, elapsed, schedulerSlack,
+				"an invalid call must not wait out the timeout")
+		})
+	}
+}
+
+// Exercises the exported entry points, pinning the millisecond conversion and
+// NVML status codes that the waitForDelivery tests above do not see.
+func TestEventSetWait_BlocksForRequestedMilliseconds(t *testing.T) {
```
ArangoGutierrez, line 211: Nothing exercises the SUCCESS branch of the exported waits: these subtests run against an uninitialized engine, so `pollPendingXid` always reports false, and the container fixture has no `failure.xid` block to deliver. Stubbing `pollPendingXid` to a constant false leaves the suite green, so the Xid-delivery wiring the PR documents is the one part that isn't guarded.

tests/mocknvml/bridge_tests.go
```diff
@@ -36,10 +39,114 @@ func bridgeTests(deviceCount int) []testResult {
...
+	if err := lib.Open(); err != nil {
+		return append(results, testResult{"event/dlopen", false, fmt.Sprintf("dlopen libnvidia-ml.so.1: %v", err)})
+	}
+	defer func() {
+		if err := lib.Close(); err != nil {
+			log.Printf("  WARN  event/dlclose: %v", err)
+		}
+	}()
+
+	for _, symbol := range []string{"nvmlEventSetWait", "nvmlEventSetWait_v1", "nvmlEventSetWait_v2"} {
```
ArangoGutierrez, line 139: This symbol probe is exactly the right guard for the `--defsym` drop you fixed, but nothing runs it — `tests/mocknvml` isn't referenced by any file under `.github/` or by the root Makefile, so it only executes on a manual `make -C tests/mocknvml test`. Issue #456 is already open for a build-time `nm -D` undefined-symbol check; wiring either that or this harness into CI would make the guard actually guard.

pkg/gpu/mocknvml/bridge/events.go
```diff
@@ -99,20 +107,70 @@ func pollPendingXid(data *C.nvmlEventData_t) bool {
...
+		return false
+	}
+	deadline := time.Now().Add(timeout)
+	for {
+		remaining := time.Until(deadline)
+		if remaining <= 0 {
+			return false
+		}
+		if remaining > pollInterval {
+			remaining = pollInterval
```
ArangoGutierrez, line 138: `waitForDelivery` degenerates into a busy spin when `pollInterval` is non-positive: the clamp assigns it to `remaining` and `time.Sleep` then returns immediately, which is the exact spin this function exists to prevent. Only the 100 ms const reaches it today, so this is just a one-line floor to keep it safe for any future caller.

tests/mocknvml/Dockerfile
avasilevskii: Fixed: it runs make clean && make now

pkg/gpu/mocknvml/Makefile
avasilevskii: Added `PROBE_TAGS_FLAG` (caller tags + nopadding) for the probe, `$(GO_TAGS_FLAG)` on the padded pass.

pkg/gpu/mocknvml/README.md
avasilevskii: Adjusted.

pkg/gpu/mocknvml/bridge/events_test.go
avasilevskii: Added `pendingXidClaim` seam in `events.go` + two tests: payload delivery (status/eventType/eventData/device handle) and delivery of an Xid injected mid-wait.

tests/mocknvml/bridge_tests.go
avasilevskii: Added new root target `test-mocknvml-bridge` + CI step in `golang.yaml`.

pkg/gpu/mocknvml/bridge/events.go
avasilevskii: Fixed: the value floors to `waitPollInterval` now.

avasilevskii:
/ok to test [9fe2888](https://github.com/NVIDIA/k8s-test-infra/pull/535/commits/9fe28884b34ce9dcb9cf64f4decba1d974ef1444)

pkg/gpu/mocknvml/bridge/events.go
```diff
@@ -99,20 +113,115 @@ func pollPendingXid(data *C.nvmlEventData_t) bool {
...
+		}
+	}
+}
+
+// waitForXid mirrors real nvmlEventSetWait semantics: NVML_SUCCESS as soon as
+// an injected Xid is pending, otherwise NVML_ERROR_TIMEOUT after blocking for
+// up to timeoutms.
+func waitForXid(data *C.nvmlEventData_t, timeoutms C.uint) C.nvmlReturn_t {
+	// NVML rejects a NULL payload; fail fast instead of after the timeout.
+	if data == nil {
```
ArangoGutierrez: The new fast-fail rejects a NULL `data` but leaves `set` unchecked, so `nvmlEventSetWait(NULL, &data, 10000)` now parks the caller for the full ten seconds and answers TIMEOUT where real NVML answers INVALID_ARGUMENT straight away. Cheap to match now that the wait actually blocks, though the test helpers pass `set = nil` today so they would need a non-nil set alongside it.

ArangoGutierrez (commented):
All seven items from the last round are addressed, and the fixes look right. One small thing left.

1. The new fast-fail rejects a NULL `data` but leaves `set` unchecked, so `nvmlEventSetWait(NULL, &data, 10000)` now parks the caller for the full ten seconds and answers TIMEOUT where real NVML answers INVALID_ARGUMENT straight away. Cheap to match now that the wait actually blocks, though the test helpers pass `set = nil` today so they would need a non-nil set alongside it. (pkg/gpu/mocknvml/bridge/events.go:152)

pkg/gpu/mocknvml/bridge/events.go
avasilevskii: Added NULL handing for `set`

Issue #533, opened 2026-07-27
tracking: k8s-native GPU fault injection — GPUFaultInjection CRD + controller, agent-applied (v0.4.0 flagship)


ArangoGutierrez:
Today the only way to mutate simulated GPU runtime state is `kubectl exec` into the nvml-mock DaemonSet pod on a node and run `nvml-mock-ctl`. That works, it is what the demos use, and it is the right tool for a handful of nodes. It is O(N) client-side actions for N nodes, so at the 5k-node scale we are targeting for DCGM and NVSentinel testing it does not hold up, and it gives us no answer to "did the fault actually land everywhere".

Raised by @roma-glushko on #493. This epic is the scale answer; #493 itself is a demo and is unaffected.

## The insight

The mock engine already reads a declarative file (`overrides.yaml`) on a 1s TTL, via `configOverrideStore` in `pkg/gpu/mocknvml/engine/config_override_store.go`. That file is already the right reconciliation target: declarative, idempotent, level-triggered. The engine does not need to change.

What is missing is a **delivery layer**. This epic builds it. That makes the work additive rather than an architecture rewrite.

It also already works through both delivery paths. CDI-injected consumers read the file via the config directory bind, and NRI-injected consumers reach the same host file because the plugin rbinds `/var/lib/nvml-mock` and injects `MOCK_NVML_CONFIG`, so `ConfigOverridePathFor` resolves to the same `overrides.yaml`. #525 puts a test behind that, since today it is only proven by reading the code.

## Direction

`GPUFaultInjection` CRD, central controller, **agent-applied**:

- A `GPUFaultInjection` CR carries a node selector plus a fault spec.
- The **controller** resolves the selector to a node set, applies campaign semantics, and aggregates status. It never writes to a node filesystem.
- The **node agent**, in the existing nvml-mock DaemonSet, watches the CR, filters to its own node, and materializes `overrides.yaml` locally.

That split is what keeps this Kubernetes-native rather than ssh-shaped. One operator action, watch-based fanout with no kubelet in the path, sub-second convergence, and real status.

Alternatives weighed and rejected:

- **Node annotations plus an agent watch.** Cheaper, and a strict subset of the agent-side logic here. No status, no schema validation, and annotation size limits. It cannot answer "did this land on all 4,997 nodes", which is the question the 5k suite exists to ask.
[...]

PR #536, opened 2026-07-27
docs(meps): [MEP0001] Mokka Control Plane

roma-glushko:
## What This PR Does

This is a proposal to shape our achitecture by introducing Control Plane, representing our simulated GPUs as inventory of allocatable sGPU nodes, and manipulating the runtime state of a collection of sGPUs via policies.

This MEP touches on the architecture that should fit https://github.com/NVIDIA/k8s-test-infra/issues/533 but that specific epic is still actual as it goes into details of how fault injection contract should look like while this MEP suggests the architecture that should work there.

The proposal also includes CRD design, examples of CRs and some important implementation decisions.

## Why

- there is no way to realistically simulate GPU rack capacity distribution
- there is no standard way to manipulate sGPU runtime information
- we ask cluster admins to setup cliques and topology which is in most cases should be inferred.

## Checklist


enhancements/meps/0001-mokka-control-plane/README.md
```diff
@@ -0,0 +1,194 @@
...
+
+Control Plane would operate as a K8s Operator and we need to make sure we can make multiple
+replicas active at the same time to fulfill node agent requests and do sGPU distribution.
+
+### Configuration
+
+The current NVML configuration may need to change slightly. 
+Currently, it contains both GPU facts and spec and some runtime information mocks. The runtime portion should be decided by Control Plane.
+
+## Drawbacks
```
smalleni, line 1452: Might be good to call out additional load on the kube control plane due to this operator/control plane here.

enhancements/meps/0001-mokka-control-plane/README.md
```diff
@@ -0,0 +1,194 @@
...
+For example, consider operational overhead, resource consumption, and how
+this will impact our users.
+-->
+
+## Design Details
+
+### sGPU Inventory Distribution
+
+Control Plane holds the currently configured sGPU inventory.
+Then, cluster admins are free to label their CPU nodes with a custom label (for example, `mokka.nvidia.com/sgpu: gb300`) that indicate which sGPU type should be available on that node.
```
avasilevskii: This can be a viable approach for precise manual sGPU cluster bringup, e.g. "I'm an advanced user and I want precisely these nodes to have profile X and those nodes to have profile Y". However, the manual labeling step partly contradicts user-story S1:`I just specify the GPU racks we are waiting for and rough network topology between them and my cluster is representative.`. 

To streamline the process for the more generic case "here's a CPU-only cluster, make as many GB300 18-node cliques with NVL72 topology as possible", we could make the Control Plane also pre-label nodes with concrete `mokka.nvidia.com/sgpu` values based on an input derived from CR. We may want to limit the node pool though (e.g. include only xlarge nodes, do not include monitoring nodes, etc.), but this can be done in the CR as well. WDYT?

Btw, how does the topology work in that case: is that part of the node labeling? Or something the user needs to specify in the CR? Or is it calculated by the Control-Plane based on the number of nodes with certain profile?

enhancements/meps/0001-mokka-control-plane/README.md
avasilevskii: +1
This is something to measure and explicitly document.

enhancements/meps/0001-mokka-control-plane/README.md
roma-glushko: @avasilevskii great feedback!

> However, the manual labeling step partly contradicts user-story S1
> We may want to limit the node pool though (e.g. include only xlarge nodes, do not include monitoring nodes, etc.), but this can be done in the CR as well. WDYT?

I realized that this morning as well. In [the recent commit](https://github.com/NVIDIA/k8s-test-infra/pull/536/changes/acca42b1fafb7c29a8b76d1ea92fdc67126394f3), I've been thinking about `mokka.nvidia.com/sgpu: auto` as a default way to distribute sGPU inventory. This way we remove cluster admin focus from the distribution at all. Is this in line with your concern?

> Btw, how does the topology work in that case: is that part of the node labeling? Or something the user needs to specify in the CR? Or is it calculated by the Control-Plane based on the number of nodes with certain profile?

I'm still working though details as I deepen my own understanding of the network pieces, but things like IMEX domains should be automatically handled by Control Plane (the commit I have referenced has that information as well). 

It's likely that cross-rack networking should be still defined by cluster admin as it's less standard (e.g. could be InifiniBand, could be RoCE, etc.). 

But if there is a way to have a reasonable default topology just using sGPU inventory without requiring to always specify network topology (but have a way to do that if needed) that would be the experience I would aim for. We can also have a concept of network topology profiles e.g. AWS RoCE profile, Azure profile, etc.

enhancements/meps/0001-mokka-control-plane/README.md
roma-glushko: Good call 👍  The load is not going to be zero as we plan to use etcd/K8s API extensively to keep all the state that mokka is supposed to have. That's the simplest approach without introducing additional dependencies like Redis.

roma-glushko: Added this information in https://github.com/NVIDIA/k8s-test-infra/pull/536/changes/db7375dd101a7939f54e4358213a8c6c969ddbd7 ✅ 

Thank you!

enhancements/meps/0001-mokka-control-plane/README.md
```diff
@@ -0,0 +1,1468 @@
...
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+This proposal introduces a new component called Mokka Control Plane that centralizes:
+- distribution of virtual GPU inventory (a new concept) across a K8s cluster
```
avasilevskii, line 24: Nit: should we stick to "simulated GPU" consistently everywhere?

enhancements/meps/0001-mokka-control-plane/README.md
```diff
@@ -0,0 +1,1468 @@
...
+- a single central NVML mock component that acts as a node agent that applies NVML mock configuration and topology from configmaps
+- each NVML instance is an independent component that knows nothing about the existence of any other NVML mock instances
+- if you want to have different GPU profiles for different nodes, you need to label those nodes differently and deploy multiple Mokka helm charts with different node selectors.
+
+![todays-architecture.png](./img/todays-architecture.png)
+
+This architecture has given us a chance to focus on developing the core simulation logic so far.
+However, we are approaching use cases that push it to its limits:
+
+- there is no way to simulate capacity distribution of GPU platforms like GB300. For example, if you want to simulate two GB300 instances in the cluster, you can get at most 2 × 18 = 36 GPU nodes, each with 4 GPUs. Today it's the responsibility of Mokka users to enforce that cap, which may lead to unrealistic cluster topologies.
```
avasilevskii, line 39: Can you elaborate on this limitation? Does it refer to profiles currently provided with Mokka?
Afaik, we can create any amount of GB300 or any other sGPU nodes by using custom profiles. Although it's true that users have to make sure the actual k8s nodes correspond to whatever is configured in Mokka's custom profile.

enhancements/meps/0001-mokka-control-plane/README.md
```diff
@@ -0,0 +1,1468 @@
...
+- [Bad] Each NodeAllocation corresponds to a K8s Node so we will have roughly 5k records when simulating a 5k-node GPU cluster, for example. So NodeAllocations have a high cardinality.
+- [Bad] It's likely that we will update those objects quite often, which adds load to the K8s Control Plane.
+
+2. Use Redis:
+- [Good] Much easier to operate on the state, change it concurrently and atomically and even search compared to etcd.
+- [Good] It scales well. It might be helpful for more advanced functionality
+- [Good] No additional pressure on the Kubernetes Control Plane
+- [Bad] We add an external dependency in a form of Redis. Even though it's the least demanding DB in terms of maintenance, we need to deploy it and potentially make sure it's snapshotting its content for persistence (so PVC would be needed).
+
+Decision: The proposal suggests moving forward with Redis as a medium to store runtime state because it makes implementation easier and removes high-cardinality data load from the K8s Control Plane.
```
avasilevskii, line 125: Agree on that 👍

enhancements/meps/0001-mokka-control-plane/README.md
```diff
@@ -0,0 +1,1468 @@
...
+
+### User Stories
+
+#### S1: Precise Multi-GPU Distribution Simulation
```
avasilevskii, line 129: Does this also include heterogeneous sGPU deployments? E.g. if I want to have a bunch of GB300 and H100 in one cluster.

enhancements/meps/0001-mokka-control-plane/README.md
```diff
@@ -0,0 +1,1468 @@
...
+
+2. Use Redis:
+- [Good] Much easier to operate on the state, change it concurrently and atomically and even search compared to etcd.
+- [Good] It scales well. It might be helpful for more advanced functionality
+- [Good] No additional pressure on the Kubernetes Control Plane
+- [Bad] We add an external dependency in a form of Redis. Even though it's the least demanding DB in terms of maintenance, we need to deploy it and potentially make sure it's snapshotting its content for persistence (so PVC would be needed).
+
+Decision: The proposal suggests moving forward with Redis as a medium to store runtime state because it makes implementation easier and removes high-cardinality data load from the K8s Control Plane.
+
+### User Stories
```
avasilevskii, line 127: Is control-plane and mokka-agent observability part of this MEP or do we leave it out-of-scope for now?

enhancements/meps/0001-mokka-control-plane/README.md
```diff
@@ -0,0 +1,1468 @@
...
+### CRD Design
+
+Here is the list of CRDs that map to our system concepts:
+
+![Mokka Control Plane CRDs](./img/mokka-sgpu-crds.png)
+
+- [Admin/GitOps] sGPU Profile: Specifies custom sGPU profiles
+- [Admin/GitOps] sGPU Inventory: Specifies a set of sGPUs available for node assignment (e.g. Rack Name: sGPU profile × count)
+- [Admin/GitOps] sGPU Runtime Policy: Modifies runtime information for the whole sGPU inventory, a sGPU rack, a node inside a sGPU rack, or a specific sGPU inside the sGPU node.
+- [Control Plane] sGPU Node Allocation (in case of storing [the runtime state in etcd](#runtime-state)): Defines the expected state of sGPU node including sGPU<->K8s Node assignment, last agent fetch time, etc.
```
avasilevskii, line 212: Taking into account that decision above to use a dedicated KV-Store (Redis) for runtime state, does it make sense to align this point with the actual plan?

enhancements/meps/0001-mokka-control-plane/README.md
```diff
@@ -0,0 +1,1468 @@
...
+
+A policy should contain only the fields it wants to control. It's a sparse override.
+
+- Omitted field means inherit
+- Explicit zero is a real value to set e.g. empty list [] or 0 int.
+- A more-specific list replaces a less-specific list
+
+#### SGPUNodeAllocation
+
+This only applies to the architecture where we store [runtime state in Kubernetes CRD](#runtime-state).
```
avasilevskii, line 671: I'm missing a section that would elaborate in the same level of detail on the data layout in Redis.

enhancements/meps/0001-mokka-control-plane/README.md
roma-glushko: > Afaik, we can create any amount of GB300 or any other sGPU nodes by using custom profiles. Although it's true that users have to make sure the actual k8s nodes correspond to whatever is configured in Mokka's custom profile.

@avasilevskii yes, that's the problem that they create any number of sGPU nodes for a GB300, for example, while in reality their node pool would be limited to 18 nodes per rack. 

The new model is not limiting cluster admins in terms of now many sGPU nodes they can simulate. They would just need to add more racks to their SGPUInventory, but the total number of sGPU nodes would be realistic and enforced by Mokka Control Plane.

enhancements/meps/0001-mokka-control-plane/README.md
roma-glushko: @avasilevskii Yes, for sure. SGPUInventory has a rackGroups field where we can pass rack count and rack type (which is any sGPU profile we support)

enhancements/meps/0001-mokka-control-plane/README.md
roma-glushko: @avasilevskii I'm glad you've spotted this! I plan to add it as a followup once we start working on the implementation. I think we will land on the much better/realistic Redis data layout once we get our hands dirty.

For the time being, this should be an appropriate detail resolution to kick off the implementation as important, key architectural decisions are decided upon.

enhancements/meps/0001-mokka-control-plane/README.md
roma-glushko: The "(in case of storing [the runtime state in etcd](#runtime-state))" actually refers to the decision about where to put runtime store. This item is only actual if etcd is a runtime store.

enhancements/meps/0001-mokka-control-plane/README.md
avasilevskii: Fair enough 👍

enhancements/meps/0001-mokka-control-plane/README.md
roma-glushko: @avasilevskii I would not want to solve all problems of this project in one MEP 😄 

In terms of observability there are three corner stones: logs, metrics, traces.

Emitting logs with multiple log levels e.g. debug, info, error, etc. goes without saying. That's the baseline o11y signals we will emit.

Traces may not be interested for a wide audience, so we should consider whether we want to have them. Optional instrumentation is a good stand here I guess.

Metrics are what we are curious the most I guess. We would need to decide how we want to emit metrics e.g. using OpenTelemetry or plain Prometheus scrape endpoints and we also need to agree which are the metrics that would be helpful for the 5k scale test (then generalize them for a broader audience). Grafana dashboards will depends on the metric set we pick.

That's my thoughts on o11y in Mokka in general.

faganihajizada:
The architecture document from #601 is now merged. It sets three constraints that apply to this MEP.

## 1. Node agent scope

The MEP scopes the agent to "purely what NVML mock does today: applying the expected state of NVML and networking".

The architecture document is wider. It says the mock gives "Same libraries and file footprints". It also puts "PCI and sysfs trees, and kernel or driver footprints under `/proc` and `/sys` (including module state such as `lsmod`)" into the contract surface. The two scopes do not agree. With the narrow scope, no component owns the driver footprint. The footprint is not NVML, so the agent does not own it. The footprint is not inventory, so the Control Plane does not own it.

Shoulw we widen the agent scope to the low-level surfaces below the contract surface. The MEP already describes the agent as "a virtual device driver (for GPU and network card)". That description is correct. Only the scope sentence needs some adjustment.

## 2. Runtime state shape

Every field in `spec.runtime` describes one GPU: `deviceState`, `modes`, `telemetry` and `errors.xid`.

Driver module state is different. It is a fact about the node. It has two values: the module is loaded, or the module is not loaded. `targetRef` has `nodeIndexes`, so you can select the node. But `spec.runtime` has no field for module state.

The architecture document sets this requirement: "each simulated surface has to be controllable independently. A GPU can show up in the PCI tree with no driver footprint; that is a normal failure mode, and the mock needs to reproduce it."

The current schema cannot express that state, because the state is not a property of a GPU.

## 3. Delivery

Should we add a constraint? The agent must write the file surfaces to the paths that the consumers read.

The architecture document gives the reason. A flag that sends a consumer to a substitute path "no longer tests that the consumer reads the real path". `LD_PRELOAD` "does not work for Go binaries". The document concludes that "file surfaces need to be mounted into the container instead of intercepted".

tjrasche:
I think the user-facing goal makes
sense: describe a finite logical GPU datacenter, project it onto ordinary
Kubernetes Nodes, and expose enough driver, health, and topology surfaces for
the real software stack above the GPUs to operate unmodified.

I may be missing some constraints, but I wonder whether we can achieve that
with a simpler Kubernetes-native state model, without Redis or an additional
agent REST service.

One thing I found difficult in the current proposal is that "runtime state"
combines several concepts with different lifecycles:

- Node-to-rack assignment is durable desired state.
- Agent heartbeat is an ephemeral health observation.
- "GPU 2 should report XID 79" is desired simulation policy.
- The current temperature or counter value is ephemeral node-local data.

These do not appear to require the same storage or coordination mechanism.

A possible alternative model could be:

```text
SGPUProfile + SGPUInventory
            │
            ▼
 Kubernetes controller
            │
            ▼
 controller-generated SGPURack
            │
            ▼
 Node labels/annotations
            │
            ▼
 local Mokka agent
            ▲
            │
  SGPURuntimePolicy
```

`SGPUInventory` would remain the user-authored declaration of rack groups and
capacity. The controller would expand it into stable logical racks and assign
eligible Kubernetes Nodes to bounded rack slots.

A controller-generated `SGPURack` could represent the rendered rack:

```yaml
spec:
  profileRef:
    name: custom-2gpu-18node-rack

  identity:
    rackGroup: training
    rackIndex: 42
    fabricUUID: ...

  slots:
    - index: 0
      nodeRef:
        name: worker-001
        uid: ...
    # ...
    - index: 17
      nodeRef:
        name: worker-018
        uid: ...
```

The Rack CR would provide a stable identity, membership, topology, conditions,
and a natural runtime-policy target. The controller could project each slot
assignment onto the corresponding Node using small labels and annotations. The
Mokka DaemonSet could select assigned Nodes, and its agent would only need to
determine "I am rack 42, slot 5, using profile X."

Existing assignments should remain fixed. The allocator would only pair
unassigned Nodes with free slots; it would not repeatedly sort all Nodes and
reshuffle racks after membership changes.

For runtime behavior, it seems sufficient to persist the desired policy rather
than a materialized runtime document:

```text
profile defaults
+ rack identity
+ node slot
+ matching runtime policies
=
effective node configuration
```

For example, a policy might say that rack 42 should simulate a temperature
around 85°C, or that GPU 2 on node slot 5 should report XID 79. The local agent
can compile that with the profile defaults and write the existing Mokka
base/override configuration.

Individual samples and counters can remain node-local:

```text
Policy: base temperature 85°C, variance 5°C
Local observations: 83°C, 87°C, 84°C, ...
```

Those observations would be exposed through mock NVML, `nvidia-smi`, DCGM, and
the other surfaces that consumers already use. They would not need to be
written back to Kubernetes or Redis.

Under this model, Kubernetes contains only:

- Profiles and inventory intent.
- Rendered racks and stable placement.
- Runtime policies.
- Small aggregate conditions and revision information.

Everything else is deterministic or derived locally. Agent disconnection would
not release a rack slot, and agent requests would not double as heartbeats.

The agents could initially consume the relevant Kubernetes resources directly
using exact-name or label-selected watches. This does create an API-server
fanout concern at the 100k-node target, and I think that deserves a benchmark.
If direct Kubernetes delivery is demonstrably insufficient, a read-only
distribution proxy or optional reconstructable cache could be introduced
later. It would not need to own allocation, identity, leases, or runtime state.

Would it make sense to split the work into two stages?

1. Static datacenter materialization: profiles, inventory, rendered racks,
   stable Node assignments, GPU identities, and topology.
2. Mutable simulation scenarios: runtime policies compiled and applied locally.

This would let us validate the rack and scheduling model independently before
committing to Redis, custom heartbeat semantics, active-active allocation, and
a new agent protocol.

PR #576, opened 2026-08-02
docs(demo): pin the kube context in the demo READMEs and the two scripts that still omit it

### LINKED ISSUE #574: docs(demo): pin the kube context in the demo READMEs and the two scripts that still omit it
While working on #546 we found that `docs/demo/node-wide-injection/README.md` documents bare `kubectl` commands with no context pinned. The failure mode is worse than a command that doesn't work: a documented `kubectl -n nvml-mock-system get daemonset nvml-mock nvml-mock-nri` will happily return a real DaemonSet from whatever cluster the reader's kubeconfig happens to point at. They see plausible output, conclude the demo worked, and are looking at an unrelated cluster.

I checked whether the other five demo docs share it. Four of them do, and the problem extends to two of the scripts as well.

### Scripts

The idiom already exists in the repo. `standalone/demo.sh` and `node-wide-injection/run.sh` both do this:

```sh
KUBE_CONTEXT="kind-${CLUSTER_NAME}"
# Wrap kubectl so every call is pinned to the demo's kind context without
# repeating --context at each call site. helm keeps --kube-context inline
kubectl_ctx() { command kubectl --context "${KUBE_CONTEXT}" "$@"; }
```

and pass `--kube-context "${KUBE_CONTEXT}"` to helm.

`compute-domain/run.sh` and `failure-injection/run.sh` never got it. Both set `CLUSTER_NAME` and stop there — no `KUBE_CONTEXT`, no wrapper, and their `helm upgrade` calls (`compute-domain/run.sh:173,336`, `failure-injection/run.sh:102,154`) run against the ambient context.

So this isn't an open design question. It's a rollout that stopped halfway.

### READMEs

| File | Pinned? | Unpinned commands |
|---|---|---|
| `README.md` | n/a | none — the single `kubectl` mention is a requirements line, not a command |
| `compute-domain/README.md` | no | `kubectl` :154, :156, :160, :256; `helm install` :146 |
| `failure-injection/README.md` | no | `kubectl exec` :105 |
| `standalone/README.md` | no | `helm uninstall nvml-mock -n nvml-mock-system` :43 |
| `with-fgo/README.md` | no | `kubectl` :77, :80, :83, :85, :88, :89, :96; `helm` :35, :51 |

Three details worth calling out:

**`compute-domain/README.md:181` contradicts itself.** It tells the reader that renaming the cluster means updating "Cluster name in every `kind` / `kubectl --context` / `kind load` call below" — but no command below it uses `--context` at all. The doc describes a convention it doesn't follow.

**Two of the unpinned commands are destructive.** `standalone/README.md:43` is `helm uninstall nvml-mock -n nvml-mock-system`, and `node-wide-injection/README.md:112-114` includes `kubectl delete namespace nvml-mock-system`. Run against the wrong context those do real damage, and `--ignore-not-found` means the wrong-cluster case is the one that produces output. `standalone` is the sharper case: its script pins, its README doesn't, so the documented teardown is less safe than the scripted one.

**`failure-injection/README.md` promises isolation it doesn't enforce.** It says the demo lives in its own cluster "so it never collides with the standalone demo or any other nvml-mock deployment" — which is exactly the guarantee an unpinned context removes.

### Suggested fix

- Add the `KUBE_CONTEXT` / `kubectl_ctx` idiom to `compute-domain/run.sh` and `failure-injection/run.sh`, copying `standalone/demo.sh` verbatim.
- Add `--context kind-<cluster>` to every runnable `kubectl` in the four READMEs, and `--kube-context` to every `helm`.
- Fix the stale `--context` reference at `compute-domain/README.md:181` so it matches whatever the commands end up doing.

Worth a moment's thought on whether a doc test could keep this from regressing — a check that every fenced `kubectl`/`helm` line under `docs/demo/` carries a context flag would be cheap and would have caught all of these. Not proposing it as part of this issue, just noting it's available.

LuisMend12:
## Summary

`compute-domain/run.sh` and `failure-injection/run.sh` set `CLUSTER_NAME` and stop there — no `KUBE_CONTEXT`, no wrapper, so every `kubectl`/`helm` call in both scripts runs against the ambient kubeconfig context instead of the demo's own Kind cluster. A documented/scripted command that happens to match a real DaemonSet, namespace, or release name on the reader's actual cluster returns plausible output and looks like it worked — while it silently operated on the wrong cluster.

This brings both scripts in line with the `KUBE_CONTEXT` / `kubectl_ctx()` idiom already established in `standalone/demo.sh` and `node-wide-injection/run.sh`:
- Added `KUBE_CONTEXT="kind-${CLUSTER_NAME}"` and a `kubectl_ctx() { command kubectl --context "${KUBE_CONTEXT}" "$@"; }` wrapper to each script.
- Replaced every bare `kubectl` call with `kubectl_ctx` (mentions of `kubectl` inside prose comments were left untouched).
- Added `--kube-context "${KUBE_CONTEXT}"` to every `helm` call.

Also pinned the manual/documented commands that were still unpinned:
- **`compute-domain/README.md`**: the `helm install` command and the four `kubectl` commands in the manual-reproduction and troubleshooting sections. This also makes the existing "Custom cluster name" note truthful — it already claimed every command below it uses `--context`, which wasn't the case until now.
- **`failure-injection/README.md`**: the interactive `kubectl exec` example.
- **`standalone/README.md`**: `helm uninstall` — a destructive command, and previously the one command in this particular demo that was *less* safe than its own script.
- **`with-fgo/README.md`**: both `helm` calls and every `kubectl` call in the verification section.

`node-wide-injection/README.md` was already fully pinned (its `run.sh` already carries the idiom), so no changes were needed there.

Fixes #574

## Test plan
- [x] `bash -n` on both modified scripts — no syntax errors
- [x] Confirmed no bare (unpinned) `kubectl` call sites remain in either script or any of the four modified READMEs (only prose mentions of `kubectl` were left alone)
- [x] Confirmed every `helm` invocation across the changed files now carries `--kube-context`/`--kube-context` consistent with the rest of the codebase's idiom
- Not run against a live Kind cluster in this environment — the change only adds context-pinning flags to existing, already-tested commands; no command's arguments or logic otherwise changed


docs/demo/compute-domain/README.md
```diff
@@ -144,20 +144,21 @@ kind load docker-image nvml-mock:compute-domain-imex --name nvml-mock-compute-do
 #    chart pulls the default upstream image which does not have the
 #    real IMEX layer baked in).
 helm install nvml-mock deployments/nvml-mock/helm/nvml-mock \
+    --kube-context kind-nvml-mock-compute-domain \
```
ArangoGutierrez, line 211: Renaming the cluster now means updating this flag too, but the "Custom cluster name" checklist a little further down still lists only `kind` / `kubectl --context` / `kind load`. That was issue #574's third item, so adding `helm --kube-context` to it would close the last piece.

ArangoGutierrez (commented):
Nothing here blocks merge.

That's true of the change itself — the sweep is complete. No bare `kubectl` is left in either script, all four `helm` calls carry `--kube-context`, every runnable command in the four READMEs is pinned, and `bash -n` is clean on both. `node-wide-injection` and `nv-sentinel` were already covered, as the description says.

Separately, and before it can merge: the commit has no `Signed-off-by` line, which is what the red DCO check is about, and `main` also requires GPG-signed commits. `git commit --amend -s -S` and a force-push clear both.

One leftover from the issue:

1. [should-fix] Renaming the cluster now means updating this flag too, but the "Custom cluster name" checklist a little further down still lists only `kind` / `kubectl --context` / `kind load`. That was issue #574's third item, so adding `helm --kube-context` to it would close the last piece. (docs/demo/compute-domain/README.md:147)

ArangoGutierrez:
image


hey @LuisMend12  please address the DCO warning so we can move on with this PR

docs/demo/compute-domain/README.md
LuisMend12: Done in 1e40c0e.

giuliocalzo:
@LuisMend12 the PR looks great! can you please sign all your commits?

giuliocalzo:
@LuisMend12 the commits missing signature are the first one

image

can you please amend those commits and sign all?

LuisMend12:
signed all files

giuliocalzo:
thanks @LuisMend12 for your contribution!

PR #599, opened 2026-08-03
docs,test(nri): document ambient injection behavior and cover plain pod

### LINKED ISSUE #585: docs+test(nri): state that ambient injection defeats "non-GPU pod sees no GPUs", and cover the un-annotated pod under the device plugin
### What is true today

With `nri.enabled=true`, the plugin mounts the mock overlay and injects the mock
environment into **every** container in every non-excluded namespace. Device
nodes are a separate opt-in (`nvml-mock.nvidia.com/devices: "true"`), but the
overlay and env are ambient. That is the feature, and the e2e suite asserts it
deliberately:

```
It("keeps the workload pod plain (no nvidia.com/gpu request)")
It("reports the profile GPUs via nvidia-smi in the injected pod")
```

The mechanism is `detectVisibleDevices` (`pkg/gpu/mocknvml/engine/engine.go`):

```go
// No filtering if all devices exist or none exist.
if absent == 0 || len(present) == 0 {
    return nil
}
```

A container with **no** `/dev/nvidiaN` nodes takes the `len(present) == 0` branch
and gets `nil`, which means *no filtering* — the full configured GPU set.

### Gap 1 — the consequence is not documented (docs)

The chart README states the mechanism ("Injects cluster-wide at
container-creation time", values table `nri.enabled`) and offers
`nri.excludedNamespaces`. It never states what that means for a test author:

> **A pod that requests no GPUs will still report every GPU while `nri.enabled=true`.**
> If your test asserts that a non-GPU workload sees no GPUs — scheduler
> isolation, admission behaviour, a negative control of any kind — either leave
> NRI disabled or put that workload in an excluded namespace.

This is the same defect class the pre-`v0.3.0` audit kept finding: a true
statement whose implication is left for the reader to derive. The `v0.3.0`
release notes have the same shape — "ordinary pods can see mock GPUs without a
GPU resource request" reads as a capability, and it is equally a constraint.

Add the warning to the chart README's NRI section, to
`docs/demo/node-wide-injection/README.md`, and to the `nri.enabled` values-table
row.

### Gap 2 — one combination is asserted nowhere (test)

`scenario_nri_test.go` covers, under `when the device plugin also serves the node`:

| pod | annotated | requests GPU | covered |
|---|---|---|---|
| `nri-dp-single` | no | yes | yes — sees exactly its allocated GPU |
| `nri-dp-a` / `nri-dp-b` | no | yes | yes — isolated from each other |
| annotated + request | yes | yes | yes — suppression, mutation-checked |
| annotated, no request | yes | no | yes — `nri-dp-optin`, sees every GPU |
| **un-annotated, no request** | **no** | **no** | **not covered** |

The plain-pod specs that do cover an un-annotated non-requesting pod run in a
context with **no device plugin installed**, so they do not answer what happens
in the composed case.

**This is written as an experiment, not a claim.** Reading
`detectVisibleDevices` predicts the pod sees every GPU, but nobody has run it,
and the audit's own rule applies: a diagnosis is a hypothesis until executed.

Run it: in the `when the device plugin also serves the node` context, schedule a
pod with no annotation and no `nvidia.com/gpu` request onto `gpuNode`, and
record what `nvidia-smi -L` reports. Then assert whatever is true, so the
behaviour stops being derivable-only-from-source.

### Why this matters beyond documentation

Mokka's purpose is to be a substrate other people test *against*. A test that
passes for the wrong reason on this platform is worse than one that fails: the
GPU Operator path is safe today only because `nri.enabled` defaults to `false`
and no scenario except `scenario_nri_test.go` turns it on. That is a property of
current defaults, not a guarantee anyone has written down.

### Not affected, checked while filing

- **GPU Operator / container toolkit testing.** `nri.enabled` appears in exactly
  one e2e file. A default `helm template` renders zero NRI objects.
- **Scheduler isolation for GPU-requesting pods.** Correct and mutation-checked;
  `--pass-device-specs` gives each pod its allocated node, so
  `detectVisibleDevices` has both a present and an absent device and filters.
  MEP-0002:427 already names breaking that oracle as "the single most important
  constraint, because it fails green".

Shankar-v27:
## What This PR Does

Fixes #585.

This PR addresses both gaps described in the issue:

- documents the user-visible behavior of ambient NRI injection in:
  - the Helm chart README,
  - the `nri.enabled` values table,
  - and the node-wide injection demo documentation.
- adds the missing e2e scenario covering an unannotated pod with no GPU request when the NVIDIA device plugin also serves the node.

The new test verifies:
- the ambient overlay is injected into the workload, and
- the runtime GPU visibility reported by `nvidia-smi -L`.

## Why

Issue #585 notes that:

- the documentation does not explain that workloads without GPU requests may still report mock GPUs when `nri.enabled=true`, and
- the composed device-plugin scenario for an unannotated pod without a GPU request was not covered by the e2e suite.

This PR documents that behavior and adds the missing test coverage.

## Testing

Executed locally:

- ✅ `go test ./pkg/nri/nvmlmock/... ./cmd/nvml-mock-nri/...`
- ✅ `go test -tags=e2e ./tests/e2e/go/framework/...`
- ✅ `go test -c -tags=e2e ./tests/e2e/go`
- ✅ Verified the new spec is selected by Ginkgo using `-ginkgo.label-filter=nri-dp-plain`

The full Kind-backed e2e scenario could not be executed locally because the required `kind` binary was not available.

## Checklist

- [ ] New code has SPDX license headers (not applicable)
- [ ] CHANGELOG.md updated (not user-facing)

### DIFF
--- a/deployments/nvml-mock/helm/nvml-mock/README.md
+++ b/deployments/nvml-mock/helm/nvml-mock/README.md
@@ -33,7 +33,12 @@ When `nri.enabled=true` (opt-in; default `false`), the chart also deploys
 `nvml-mock-nri`, a node-local containerd NRI plugin. It mounts the host overlay
 into newly created containers at `/opt/nvml-mock` and injects the mock
 environment at runtime, so plain pods can run `nvidia-smi` without GPU resource
-requests or pod-spec mutation. Because it injects cluster-wide, it is off by
+requests or pod-spec mutation. The overlay and environment are injected ambiently into
+containers in non-excluded namespaces, while host device nodes (`/dev/nvidia*`) remain opt-in
+(via `nvidia.com/gpu` requests or the `nvml-mock.nvidia.com/devices: "true"` annotation).
+Unannotated pods without GPU requests will still report GPUs if `nvidia-smi` is run inside them.
+Test suites that rely on non-GPU pods seeing zero GPUs should either keep NRI disabled (`nri.enabled=false`)
+or run within an excluded namespace (`nri.excludedNamespaces`). Because it injects cluster-wide, it is off by
 default. Kind clusters must have containerd NRI enabled; see
 [`docs/demo/node-wide-injection`](../../../../docs/demo/node-wide-injection).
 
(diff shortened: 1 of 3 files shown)


roma-glushko (approved):
Looks good to me 👍

roma-glushko:
@Shankar-v27 thanks for the contribution! 

We typically want to have two things for each commit:
- DCO/Signed-off-by ✅ 
- Github Commit Verified Badge ❌  (you would need to sign commits either with GPG or SSH key)

Could you please repush your changes under a verified commit? 🙏 

We should be ready to merge it right after.

Shankar-v27:
Thanks! I've re-signed the commit with an SSH signing key and force-pushed the updated commit. The commit should now show as Verified. Thanks again for the review!

roma-glushko:
@Shankar-v27 the PR has been merged 🎉  
Welcome to the Mokka contributors list 😎

Shankar-v27:
> @Shankar-v27 the PR has been merged 🎉 Welcome to the Mokka contributors list 😎

Thank you! 🎉 Really appreciate the review and support. Happy to be part of the Mokka contributors list 😂, and I look forward to contributing more! 😂

PR #601, opened 2026-08-04
docs: Updated General Architecture

roma-glushko:
## What This PR Does

Updated the general architecture approach that we use in Mokka.

## Why

Share the same vision across all team members and contributors.

## Checklist


docs/architecture.md
```diff
@@ -1,6 +1,19 @@
-# Architecture Guide
+# Architecture
 
-Deep dive into Mock NVML's design and implementation.
+The main idea of Mokka is to simulate very low-level system contacts,
```
ArangoGutierrez: "system contacts" looks like a typo for "system contracts" - the diagram labels this the contract surface, and the closing paragraph says contracts too.

ArangoGutierrez (commented):
Good diagram, and the layered framing reads well. One blocker before merge, plus a typo.

1. [must-fix] This is the whole working canvas rather than just the diagram used here: 153 of its 1315 elements make the picture embedded in the doc, and 979 of the rest (560 text labels) are unrelated scenes. Those include the AIHub sync data model with its internal API paths and sacctmgr provisioning commands, and an identity topology naming ITSS LDAP, SailPoint, Azure Entra ID, a Kerberos primary domain controller, Teleport, CorpNet Santa Clara, and the NCP VPC behind its VPN tunnel and firewall. This repo is public, and nothing in the tree references the .excalidraw at all, only the PNG. Simplest fix is to drop it from the PR, or re-export a canvas holding just the Mokka scene. Better to do it before merge than after, since undoing it on a public repo means a history rewrite rather than a delete commit. (docs/img/mokka-general-architecture.excalidraw:1) — https://github.com/NVIDIA/k8s-test-infra/blob/bf617dc4bff63d1f4c0c9f1cd087fbbbaf22fe74/docs/img/mokka-general-architecture.excalidraw
2. [should-fix] "system contacts" looks like a typo for "system contracts" - the diagram labels this the contract surface, and the closing paragraph says contracts too. (docs/architecture.md:3)

roma-glushko:
@ArangoGutierrez the excalidraw scene file is removed. The source of the image is embedded directly into the PNG file. No need for the excalidraw file.

docs/architecture.md
roma-glushko: Fixed in https://github.com/NVIDIA/k8s-test-infra/pull/601/commits/de1b67ee4657cf5ede13ac1d4d3bf44172adebe6 ✅  Thanks!

faganihajizada (dismissed):

faganihajizada (approved):

faganihajizada (approved):

PR #611, opened 2026-08-06
docs: Set up AGENTS.md file

roma-glushko:
## What This PR Does

Setting up AGENTS.md file.

## Why

Sharing the important assumptions between agent instances.

## Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,32 @@
+# Mokka
+
+Mokka helps simulate expensive GPU infrastructure on CPU nodes by
+simulating the presence of GPU and network devices at the driver level.
+
+The idea is to simulate the driver footprint at a low level,
+so all consumers and applications up the stack work without modifications.
+
+## Principles
+
+- Simple and minimal code
+- Strive for high cohesion and low coupling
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,32 @@
...
+- Simple and minimal code
+- Strive for high cohesion and low coupling
+- Write idiomatic Golang code and project structure
+- Red-green test-driven development
+- Code should communicate your intent
+- Code should be composable and testable
+
+## Scale
+
+- The system should work well for clusters with tens of thousands or a few million nodes.
```
avasilevskii: I would reduce the scale a bit here and cap it 100k nodes to avoid over-engineered solutions. In fact, k8s cannot scale to even 1M nodes yet, e.g. see https://github.com/bchess/k8s-1m

roma-glushko: > In fact, k8s cannot scale to even 1M nodes yet, e.g. see https://github.com/bchess/k8s-1m

That's an illuminating read 👀  thanks for sharing.

Let's cap it to 100k then 👍

PR #621, opened 2026-08-07
feat(demo): deliver compute-domain mocks through NRI

### LINKED ISSUE #551: docs(demo): compute-domain demo predates the NRI foundation and does not use it
## Summary

`docs/demo/compute-domain` predates the NRI foundation. It installs the `gb200` profile with a
purpose-built demo image and never enables NRI, so the multi-node ComputeDomain story — the
headline payoff of the NRI direction — is still demonstrated on the older delivery path.

Carved out of #437 when that issue closed.

## What the demo does today

`docs/demo/compute-domain/run.sh` installs the chart with:

```
--set image.repository=nvml-mock
--set image.tag=compute-domain-imex
--set gpu.profile=gb200
```

plus a topology document. It builds a local demo image layering the real `nvidia-imex` behind
`imex-nogpu-shim`, and Scenario 2 forms a genuine IMEX domain over the pod network, asserting
`nvidia-imex-ctl -q` readiness including peer-death detection.

There is no `nri.enabled`, and no `nvml-mock.nvidia.com/imex-channels` annotation anywhere in it.

## Why it matters

This is a consistency gap, not a defect — the demo works and proves something real. But #441
framed NRI as the delivery mechanism that makes workloads testable unmodified, and #437 called
the multi-node ComputeDomain case "the main payoff of the NRI direction." A reader following the
compute-domain demo sees none of that. The two most load-bearing demos for the v0.3.0 story
currently use two different delivery mechanisms.

The channel-injection path itself is covered: #549 injects
`/dev/nvidia-caps-imex-channels/channel0..N-1` into annotated containers and asserts consistency
across two kind nodes on all six GPU profiles. What is missing is the demo tying that to a real
IMEX domain.

## What to do

Either port the compute-domain demo to the NRI path — `nri.enabled=true` plus the channel
annotation, dropping whatever bespoke wiring NRI now makes unnecessary — or document in its README
why it deliberately stays on the older path.

Porting is the better outcome if the real `nvidia-imex` daemon can consume NRI-injected channel
nodes. That is the open question and should be answered by running it, not by reading: the
channels are presence-only (statable and listable, not openable), and whether that satisfies the
real daemon at domain-formation time is unverified.

If it turns out the real daemon needs openable channel devices, say so — that is a genuine
finding about the limits of the mock, and it belongs in the docs rather than being worked around.

## Acceptance

- The compute-domain demo either runs on the NRI path, or its README states why it does not.
- If ported, the existing domain-formation and peer-death assertions still pass.

## References

- #437 — closed; channel injection shipped in #549
- #441 — the NRI epic
- #304 — replaced the fake IMEX binaries with the real daemon in NO GPU mode

tjrasche:
## What This PR Does

Reworks the ComputeDomain demo so containerd NRI delivers the mock GPU stack to
an ordinary real-IMEX workload instead of baking mock libraries, files, mounts,
and environment variables into that workload.

The demo now:

- keeps the chart-managed `nvml-mock:compute-domain` DaemonSet as the node-local
  staging layer;
- adds a `workload` image target with Ubuntu, real `nvidia-imex`, and
  `imex-nogpu-shim`, but no nvml-mock payload;
- deploys an ordinary IMEX DaemonSet in a separate namespace and lets the NRI
  plugin inject the staged mock overlay, topology configuration, and annotated
  IMEX channel devices;
- selects unused character-device majors across every Kind node before enabling
  mock IMEX channels;
- gives the workload a peer-only ingress policy for the IMEX ports;
- rejects reused legacy clusters whose containerd NRI configuration is disabled
  or uses an incompatible socket, with `FORCE_RECREATE=true` as an explicit
  destructive opt-in;
- restarts staging, NRI, and workload DaemonSets in dependency order so reruns
  pick up same-tag images and topology changes; and
- verifies the injected channel device and the complete fabric identity of every
  reported GPU before exercising IMEX peer formation, liveness, and topology
  rebinding.

## Why

The previous demo ran the real IMEX overlay inside the nvml-mock DaemonSet. It
proved that the mock and the real IMEX protocol could work together, but it did
not prove the intended integration: an otherwise ordinary workload receiving
the mock driver surface, per-node topology identity, and channel devices through
NRI at container creation.

Separating node-local staging from the workload more closely represents how the
upstream ComputeDomain components consume nvml-mock. The reuse checks and ordered
rollouts also make repeated demo runs deterministic instead of silently reusing
an incompatible cluster or stale workload state.

## Validation

- `bash -n docs/demo/compute-domain/run.sh`
- extracted and syntax-checked the README manual-reproduction Bash block
- `git diff --check upstream/main...HEAD`
- strict YAML/TOML parsing and typed Kubernetes decoding of the changed manifests
- safe stub coverage for compatible/legacy/forced cluster paths, dynamic majors,
  NRI socket parsing, and adversarial multi-GPU fabric output
- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- `make check-modules`
- built and inspected the `workload` image target: real IMEX, ctl, config, and
  shim are present; mock libraries, files, and `check-fabric` are absent

A full Kind demo and Helm/Tilt validation were not run locally because `kind`,
`kubectl`, `helm`, and `tilt` are unavailable in this workspace.

## Checklist

### DIFF
--- a/deployments/nvml-mock/Dockerfile.compute-domain-daemon
+++ b/deployments/nvml-mock/Dockerfile.compute-domain-daemon
@@ -23,8 +23,8 @@
 #       --nogpu appended: the real daemon then runs the real gRPC peer
 #       protocol without GPUs.
 #
-#   demo — the same real-IMEX trio layered onto the nvml-mock image;
-#       used by docs/demo/compute-domain/run.sh Scenario 2.
+#   demo — compatibility target with the same real-IMEX trio layered
+#       onto the nvml-mock image; used by the local Tilt workflow.
 #
 # LOCAL BUILD ONLY — never publish these images: they repackage the
 # proprietary nvidia-imex binary (installed from Ubuntu jammy
@@ -81,7 +81,7 @@ COPY --from=shim /out/imex-nogpu-shim         /usr/bin/nvidia-imex
 # daemon (default target): real IMEX on top of the upstream
 # compute-domain-daemon image. Upstream renders its own imexd.cfg from
 # its template, so no config.cfg is copied here.
-FROM ${COMPUTE_DOMAIN_DAEMON_IMAGE}
+FROM ${COMPUTE_DOMAIN_DAEMON_IMAGE} AS daemon
 COPY --from=imex /usr/bin/nvidia-imex         /usr/bin/nvidia-imex.real
 COPY --from=imex /usr/bin/nvidia-imex-ctl     /usr/bin/nvidia-imex-ctl
 COPY --from=shim /out/imex-nogpu-shim         /usr/bin/nvidia-imex
--- a/docs/demo/compute-domain/demo-workload.yaml
+++ b/docs/demo/compute-domain/demo-workload.yaml
@@ -1,0 +1,65 @@
+# Copyright 2026 NVIDIA CORPORATION
+# SPDX-License-Identifier: Apache-2.0
+#
+# DaemonSet placement is intentional: scenarios 1 and 3 select pods by node and
+# require exactly one NRI-injected demo workload pod on every GPU worker.
+# The demo workload has no mock volume, mock environment variables, GPU request,
+# or RuntimeClass. containerd's NRI plugin supplies the mock overlay, topology
+# identity, and the annotated IMEX channel device nodes.
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: compute-domain-demo-workload
+  labels:
+    app.kubernetes.io/name: compute-domain-demo-workload
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: compute-domain-demo-workload
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: compute-domain-demo-workload
+      annotations:
+        nvml-mock.nvidia.com/imex-channels: "true"
+    spec:
+      automountServiceAccountToken: false
+      nodeSelector:
+        nvidia.com/gpu.present: "true"
+      containers:
+        - name: demo-workload
+          image: nvml-mock:compute-domain-workload
+          imagePullPolicy: IfNotPresent
+          command: ["/bin/sh", "-c", "trap : TERM INT; sleep infinity & wait"]
+          resources:
+            requests:
+              cpu: 10m
+              memory: 32Mi
+            limits:
+              cpu: 100m
+              memory: 128Mi
+---
+# The real IMEX daemons listen on both the peer and command/status ports. A
+# podSelector-only peer is namespace-local, so only other pods from this
+# demo workload in compute-domain-workload can connect. Selecting Ingress alone
+# intentionally leaves all egress unrestricted.
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: compute-domain-demo-workload
+spec:
+  podSelector:
+    matchLabels:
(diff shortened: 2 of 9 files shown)


deployments/nvml-mock/Dockerfile.compute-domain-daemon
```diff
@@ -78,10 +82,18 @@ COPY --from=imex /usr/bin/nvidia-imex-ctl     /usr/bin/nvidia-imex-ctl
...
+# than from the workload image itself.
+FROM imex AS workload
+RUN mv /usr/bin/nvidia-imex /usr/bin/nvidia-imex.real
+COPY --from=shim /out/imex-nogpu-shim         /usr/bin/nvidia-imex
+
 # daemon (default target): real IMEX on top of the upstream
 # compute-domain-daemon image. Upstream renders its own imexd.cfg from
 # its template, so no config.cfg is copied here.
-FROM ${COMPUTE_DOMAIN_DAEMON_IMAGE}
+FROM ${COMPUTE_DOMAIN_DAEMON_IMAGE} AS daemon
```
roma-glushko, line 84: I know it was there before but why to keep it as a part of this image? This image is a NRI target demo application, so it makes a lot of sense to represent that by having a separate Dockerfile for it.

docs/demo/compute-domain/demo-workload.yaml
```diff
@@ -0,0 +1,63 @@
+# Copyright 2026 NVIDIA CORPORATION
+# SPDX-License-Identifier: Apache-2.0
+#
+# A normal workload: it has no mock volume, mock environment variables, GPU
+# request, or RuntimeClass. containerd's NRI plugin supplies the mock overlay,
+# topology identity, and the annotated IMEX channel device nodes.
+apiVersion: apps/v1
+kind: DaemonSet
```
roma-glushko, line 10: Does it have to be a `Deamonset`? The type `Daemonset` + `compute-domain-imex` as a name make it feel like it's a part of the mokka system (like a part of the imex mocking functionality) while in reality it's just a demo application to showcase NRI in action.

I would probably call it demo-workload + maybe make it a regular `Deployment`.

docs/demo/compute-domain/run.sh
roma-glushko, line 1: Is this the right time to start ditching those sh files in favor of tilt?

tjrasche: I think it's time but that should be a different PR

docs/demo/compute-domain/demo-workload.yaml
tjrasche: Updated in 797dedf8. I kept the DaemonSet intentionally because scenarios 1 and 3 need exactly one NRI-injected workload pod on every GPU worker and select those pods by node; a Deployment would need a fixed replica count plus placement constraints. I renamed the manifest, resources, container, and labels to compute-domain-demo-workload so it is clearly a demo workload rather than a Mokka component.

deployments/nvml-mock/Dockerfile.compute-domain-daemon
tjrasche: Fixed in 0ce41ebf. I moved the NRI target workload into docs/demo/compute-domain/Dockerfile and left Dockerfile.compute-domain-daemon focused on its existing demo and daemon overlays. The run script and documentation now build the dedicated nvml-mock:compute-domain-workload image.

roma-glushko (approved):

PR #630, opened 2026-08-11
fix(mocknvml): make nvidia-smi report configured GPU processes correctly

giuliocalzo:
## What This PR Does

Makes `nvidia-smi` report a mock GPU's configured `processes:` correctly — in the default table's Processes box, `-q`, and `--query-compute-apps` — and stops it inventing processes that were never configured.

Both bugs came from the same place: nvidia-smi enumerates processes through NVML's **internal export table**, not the public `nvmlDeviceGet*RunningProcesses` APIs, and our catch-all stub for that table never wrote anything back.

- **Phantom processes.** The stub returned `NVML_SUCCESS` without touching the caller's count, so nvidia-smi rendered its uninitialized buffer as hundreds of `PID 0 / empty name / 0 MiB` rows per GPU. It now writes back a real count.
- **Hidden processes.** The entry is not the public 24-byte `nvmlProcessInfo_t`. Probing real nvidia-smi 580.65.06 (stamp each word with its own offset, read back which offset each row came from) showed a `nvmlProcessDetail_v1_t`-style header followed by an **inline 4096-byte name buffer** — a 4128-byte stride. With the wrong stride only the first process rendered; with an empty inline name the default table dropped the rows entirely.
- **Runtime control** needs no new subcommand: `nvml-mock-ctl set --gpu 0 'processes=[{pid: 100, type: C, name: train.py, used_memory_mib: 8192}]'`.

Known gaps, documented in the CHANGELOG and `docs/configuration.md`: the `Type` column always reads `M+C+G` (the entry carries no type field), and `pmon` uses an unmapped entry point — it has never listed processes on the mock, and now fails loudly (`Not supported on the device(s)`, non-zero) where `main` printed nothing and exited 0.

<details>
<summary>Two self-inflicted regressions caught before merge</summary>

**`nvidia-smi topo -m` (commit 5).** The first CI run failed every `e2e`/`e2e-dra` job on the pre-existing NVLink topology spec. Not a stale base — the branch was already on `main`'s tip. Commit 1 had switched the stub's catch-all from `NVML_SUCCESS` to `NVML_ERROR_NOT_SUPPORTED`; every slot of the table points at that one stub, so it also answered the calls `topo -m` probes, and `topo -m` aborts on any error. The catch-all was never needed for the phantom fix and is back on `NVML_SUCCESS`.

**Wrong name across GPUs (commit 6).** `mockInternalFillProcessList` resolved names through an engine-wide pid search — a leftover from when names were thought to arrive only via `nvmlSystemGetProcessName`, which takes a bare pid. With the same pid on two GPUs, GPU 1 rendered GPU 0's name. It now resolves against the device it already holds.

Commit 6 also corrected comments left over from superseded hypotheses, dropped a debug print added while chasing the topology bug, and moved both CHANGELOG entries out of the released `0.3.0` section into `[Unreleased]` (removing a duplicate `### Fixed` heading).

</details>

## Why

Mock GPU nodes could not be used to exercise anything process-aware: consumers saw either phantom processes or none at all, no matter what the profile configured.

Verified against real nvidia-smi 580.65.06 by diffing 18 subcommands between an image built from `main` and one from this branch, under both an empty and a 3-process config — no differences other than `pmon`. E2E covers it: a `runtime-control` scenario drives the list through `nvml-mock-ctl` and asserts all three views (three processes on purpose, since a single-process test passes even with the broken stride), plus a phantom-process guard in the shared `NvidiaSMI` assertion.

## Checklist

- [x] Linter passes (`golangci-lint run`) — 0 issues
- [x] Documentation updated (if applicable) — `docs/configuration.md`


pkg/gpu/mocknvml/engine/device.go
```diff
@@ -483,6 +483,19 @@ func (d *ConfigurableDevice) GetComputeRunningProcesses() ([]nvml.ProcessInfo, n
 	return procs, nvml.SUCCESS
 }
 
+// ProcessNameByPID returns the configured name of a process on this device, or
+// "" when the device has no process with that pid. nvml.ProcessInfo carries no
+// name, so callers that need one alongside a running-process list look it up
+// here rather than searching every device (two GPUs may host the same pid).
+func (d *ConfigurableDevice) ProcessNameByPID(pid uint32) string {
```
roma-glushko: does it make sense to return just name vs process struct? The second sounds like a more general approach useful for broader set of use cases.

roma-glushko (dismissed):

roma-glushko (approved):
The updates look good to me 👍

PR #645, opened 2026-08-11
fix(nvml-mock): gate T.Limit temperature fields on Ada+

### LINKED ISSUE #635: [Bug]: nvidia-smi -q reports impossible temperature limits on pre-Ada profiles because T.Limit fields are not architecture-gated
## Component / version

## What happened

The mock answers the T.Limit temperature threshold field IDs (`NVML_FI_DEV_TEMPERATURE_SHUTDOWN_TLIMIT`, `..._SLOWDOWN_TLIMIT`, `..._GPU_MAX_TLIMIT`, IDs 193–196) for every device that has a `thermal:` block, regardless of GPU architecture. Real hardware only reports these on Ada and later; Ampere and earlier report absolute thresholds through `nvmlDeviceGetTemperatureThreshold`.

nvidia-smi prefers the T.Limit fields whenever they are present, so on pre-Ada profiles two things go wrong at once:

1. The row **labels** switch to the "T.Limit" form, which real Turing or Ampere hardware never produces.
2. The **values** are rendered as signed margins from a reference rather than absolute temperatures, producing negative and inverted limits.

As a side effect, the absolute thresholds configured in every profile are never displayed at all.

### Example output

`profiles/a100.yaml` configures `shutdown_threshold_c: 92`, `slowdown_threshold_c: 87`, `max_operating_c: 83`, `temperature_gpu_c: 33`. `nvidia-smi -q` reports:

```
    Temperature
        GPU Current Temp                  : 33 C
        GPU T.Limit Temp                  : 54 C
        GPU Shutdown T.Limit Temp         : -5 C
        GPU Slowdown T.Limit Temp         : 0 C
        GPU Max Operating T.Limit Temp    : 4 C
        GPU Target Temperature            : 83 C
        Memory Current Temp               : 31 C
```

Every value is `slowdown − X`: `87−92 = −5`, `87−87 = 0`, `87−83 = 4`, `87−33 = 54`. A shutdown limit below the slowdown limit, and a negative temperature, are not physically meaningful.

The `t4` profile (Turing, 2018 hardware) produces the same shape:

```
        GPU Current Temp                  : 30 C
        GPU T.Limit Temp                  : 63 C
        GPU Shutdown T.Limit Temp         : -3 C
        GPU Slowdown T.Limit Temp         : 0 C
        GPU Max Operating T.Limit Temp    : 4 C
```

### Root cause

`tlimitThresholdFieldValue` in `pkg/gpu/mocknvml/engine/field_values.go` returns the offsets whenever `c.Thermal != nil`, with no architecture check — even though the function's own doc comment states these are reported by "Ada and later hardware ... instead of the legacy `nvmlDeviceGetTemperatureThreshold` scalars".

The offsets themselves are intentional and correct: NVSentinel's `GpuThermalMarginWatch` needs the slowdown entry to arm, and the live margin is measured against the same reference. Only the missing architecture gate is wrong.

Affected profiles: `t4` (turing) and `a100` (ampere). `l40s`, `h100`, `b200`, `gb200`, and `gb300` are Ada or later and behave correctly today.

## Expected behaviour

- On pre-Ada architectures, field IDs 193–196 return `NVML_ERROR_NOT_SUPPORTED`, and `nvidia-smi -q` shows the absolute rows `GPU Shutdown Temp : 92 C`, `GPU Slowdown Temp : 87 C`, `GPU Max Operating Temp : 83 C` for `a100`.
- On Ada and later, behaviour is unchanged, including `GetMarginTemperature` and the NVSentinel margin path.
- No profile ever renders a negative or inverted temperature limit.

## Steps to reproduce

```bash
# build the mock image at this commit, then:
  nvml-mock:latest nvidia-smi -q -d TEMPERATURE

# same defect on a Turing profile
#   ...profiles/t4.yaml...
# contrast with an Ada+ profile, where the T.Limit form is correct
#   ...profiles/h100.yaml...
```

## Environment

```yaml
```

## Acceptance criteria

- [ ] Field IDs 193–196 are gated on Ada and later; pre-Ada devices return `NVML_ERROR_NOT_SUPPORTED` for them.
- [ ] `nvidia-smi -q` on `a100` and `t4` shows absolute `GPU Shutdown Temp`, `GPU Slowdown Temp`, and `GPU Max Operating Temp` rows equal to the configured thresholds.
- [ ] `h100`, `gb200`, `gb300`, `b200`, and `l40s` output is byte-for-byte unchanged, and `GetMarginTemperature` behaviour is unchanged on all profiles.
- [ ] `TestGetFieldValue_TlimitThresholds` declares an Ada-or-later architecture, and gains a pre-Ada case asserting `NOT_SUPPORTED`.
- [ ] **Dedicated e2e test.** A new Ginkgo spec under `tests/e2e/go/` with `Label("nvidia-smi")` that parses `nvidia-smi -q -d TEMPERATURE` on a pre-Ada profile (`a100`) and an Ada-or-later profile (`h100`) and asserts: the pre-Ada output uses the absolute row labels with values matching the profile's configured thresholds; the Ada+ output retains the `T.Limit` labels; and in both cases no reported temperature limit is negative or ordered below the slowdown limit. The test must fail against the current `main`.

## Out of scope

giuliocalzo:
## What This PR Does

Gates `NVML_FI_DEV_TEMPERATURE_*_TLIMIT` (field IDs 193–196) on Ada and later in nvml-mock. Pre-Ada profiles (`t4`, `a100`) return `NVML_ERROR_NOT_SUPPORTED` so `nvidia-smi -q` falls back to absolute `GPU Shutdown/Slowdown/Max Operating Temp` rows from `nvmlDeviceGetTemperatureThreshold`.

Adds unit coverage for the pre-Ada path, a standalone e2e `nvidia-smi` assertion that checks architecture-correct threshold labels (absolute vs T.Limit), and a CHANGELOG entry.

## Why

Fixes #635 — without the architecture gate, signed T.Limit margins were rendered as impossible (negative / inverted) absolute temperatures on Ampere and Turing profiles.

## Checklist

- [x] Tests pass (`go test -race` on engine + e2e profile/assertions packages)

## Test plan

- [x] `go test -race ./pkg/gpu/mocknvml/engine/ -run 'TestGetFieldValue_Tlimit|TestGetMarginTemperature'`
- [x] `go test -race ./tests/e2e/go/profile/ ./tests/e2e/go/assertions/`
- [x] `DiffTemperatureQuery` rejects the buggy a100 sample from #635 and accepts fixed pre-Ada / Ada+ samples
- [ ] E2E: `make e2e E2E_PROFILES=a100,h100 E2E_LABEL=nvidia-smi` (or CI standalone matrix)


giuliocalzo:
The `e2e (t4)` and `e2e (a100)` failures on the first commit were real, and the new spec caught a gap in the original fix.

Gating only field IDs 193–196 made the T.Limit threshold rows report `N/A`, but nvidia-smi still chose the T.Limit layout and never printed the absolute rows, because it picks the temperature section shape from `nvmlDeviceGetMarginTemperature` (the `GPU T.Limit Temp` row), which the mock still answered on pre-Ada:

```
GPU Current Temp                  : 63 C
GPU T.Limit Temp                  : 33 C
GPU Shutdown T.Limit Temp         : N/A
```

`nvmlDeviceGetMarginTemperature` reports headroom to the T.Limit reference, which pre-Ada hardware does not have, so it is now gated on the same Ada+ check via a shared `reportsTLimit()` helper.

Verified with the real nvidia-smi in the mock image (`nvidia-smi -q -d TEMPERATURE`):

```
# t4 (turing)                          # a100 (ampere)                        # h100 (hopper, unchanged)
GPU Current Temp        : 30 C         GPU Current Temp        : 33 C         GPU Current Temp               : 34 C
GPU T.Limit Temp        : N/A          GPU T.Limit Temp        : N/A          GPU T.Limit Temp               : 53 C
GPU Shutdown Temp       : 96 C         GPU Shutdown Temp       : 92 C         GPU Shutdown T.Limit Temp      : -5 C
GPU Slowdown Temp       : 93 C         GPU Slowdown Temp       : 87 C         GPU Slowdown T.Limit Temp      : 0 C
GPU Max Operating Temp  : 89 C         GPU Max Operating Temp  : 83 C         GPU Max Operating T.Limit Temp : 4 C
```

Two notes on the issue's acceptance criteria:

- It asked for `GetMarginTemperature` to be unchanged on all profiles. That is not compatible with the primary expectation (absolute rows on `a100`/`t4`), since the margin API is what drives the layout. It is unchanged on Ada+, so the NVSentinel `GpuThermalMarginWatch` path is untouched — the thermal-margin demo defaults to `h100`.
- Pre-Ada output keeps a `GPU T.Limit Temp : N/A` row: the same nvidia-smi build prints that line whenever the query is unsupported. The e2e assertion only rejects a T.Limit row carrying a value, which is the actual defect shape.

tests/e2e/go/assertions/nvidiasmi_temperature_test.go
```diff
@@ -0,0 +1,80 @@
...
+        GPU Target Temperature            : 83 C
+        Memory Current Temp               : 32 C
+        Memory Max Operating T.Limit Temp : N/A
+`
+
+func TestDiffTemperatureQuery_RejectsBuggyPreAdaOutput(t *testing.T) {
+	problems := DiffTemperatureQuery(buggyPreAdaTemperatureQuery, false, 92, 87, 83)
+	require.NotEmpty(t, problems, "buggy main output must fail the pre-Ada check")
+	joined := strings.Join(problems, "; ")
+	assert.Contains(t, joined, "T.Limit")
```
roma-glushko, line 60: Any good reason to use assert over require here? 👀

tests/e2e/go/profile/profile.go
```diff
@@ -212,3 +229,27 @@ func (p Profile) FabricMgr() bool { return p.hasSwitches || p.fabricAuto }
...
+// SlowdownThresholdC is thermal.slowdown_threshold_c from the profile.
+func (p Profile) SlowdownThresholdC() int { return p.slowdownThresholdC }
+
+// MaxOperatingC is thermal.max_operating_c from the profile.
+func (p Profile) MaxOperatingC() int { return p.maxOperatingC }
+
+// ReportsTLimitTemp is true when real hardware of this architecture reports the
+// GPU T.Limit temperature field IDs (Ada and later). Pre-Ada profiles keep the
+// legacy absolute threshold rows via nvmlDeviceGetTemperatureThreshold.
+func (p Profile) ReportsTLimitTemp() bool {
```
roma-glushko, line 248: Looks like we have this logic in two places e.g. https://github.com/NVIDIA/k8s-test-infra/pull/645/changes#diff-1242c1b803d219a1f9c0d5e7cd78de36988663ddd532f5a040f25e99351a4a64R654.

Wondering if there is a way to dry it, so there is one place to change the condition not two.

roma-glushko (approved):

tests/e2e/go/assertions/nvidiasmi_temperature_test.go
giuliocalzo: `require.NotEmpty` establishes the precondition (there must be problems to inspect) and stops the test if it fails, since joining an empty slice would make the rest meaningless. The two `assert.Contains` checks after it are independent, so keeping them as `assert` lets a single run report every missing diagnostic instead of stopping at the first — more useful output when the matcher shape changes.

tests/e2e/go/profile/profile.go
giuliocalzo: This is intentional duplication. The `profile` package is deliberately an independent e2e oracle derived from the chart YAML — reusing the engine's `reportsTLimit()` predicate would couple the expected behavior to the implementation under test, so the same predicate bug would pass on both sides and the e2e would stop being a real check. They also operate on different types: the engine compares NVML arch enums, while the profile loader classifies the raw `architecture` string from YAML.

`TestDerivations` pins `ReportsTLimitTemp` for every shipped profile, so a drift between the two is caught by that table rather than hidden. Happy to add a short comment on the profile method pointing at the engine gate as the production counterpart if that helps discoverability.

PR #647, opened 2026-08-12
fix(nvml-mock): wire stubbed encoder/FBC/accounting exports

### LINKED ISSUE #636: [Bug]: eight engine methods are unreachable behind generated stub exports, silently dropping configured encoder_stats and fbc_stats
## Component / version

## What happened

Eight `ConfigurableDevice` methods are fully implemented in the engine, and some of them read values that every profile explicitly configures — but their C exports are generated stubs that return `NVML_ERROR_NOT_SUPPORTED`. The engine code and the profile configuration are therefore unreachable through the ABI: a user sets a value, the mock accepts it without complaint, and `nvidia-smi` reports `N/A`.

Unreachable engine methods and their stubbed exports:

| Engine method | Stubbed export | Visible in `-q` |
|---|---|---|
| `GetEncoderStats` | `nvmlDeviceGetEncoderStats` | yes — `Encoder Stats` |
| `GetFBCStats` | `nvmlDeviceGetFBCStats` | yes — `FBC Stats` |
| `GetAccountingBufferSize` | `nvmlDeviceGetAccountingBufferSize` | yes — `Accounting Mode Buffer Size` |
| `GetViolationStatus` | `nvmlDeviceGetViolationStatus` | no |
| `GetRetiredPages_v2` | `nvmlDeviceGetRetiredPages_v2` | no |
| `GetEncoderSessions` | `nvmlDeviceGetEncoderSessions` | no |
| `GetEncoderCapacity` | `nvmlDeviceGetEncoderCapacity` | no |

The silent-drop behaviour is the important part: `encoder_stats` and `fbc_stats` are present in every profile under `device_defaults`, are parsed into `EncoderStatsConfig` / `FBCStatsConfig`, are served by `GetEncoderStats` / `GetFBCStats`, and still never reach `nvidia-smi`.

### Example output

`profiles/a100.yaml` configures:

```yaml
  encoder_stats:
    session_count: 0
    average_fps: 0
    average_latency_us: 0

  fbc_stats:
    session_count: 0
    average_fps: 0
    average_latency_us: 0
```

`nvidia-smi -q` reports seven `N/A` rows where real hardware reports numbers:

```
    Encoder Stats
        Active Sessions                   : N/A
        Average FPS                       : N/A
        Average Latency                   : N/A
    FBC Stats
        Active Sessions                   : N/A
        Average FPS                       : N/A
        Average Latency                   : N/A
...
    Accounting Mode Buffer Size           : N/A
```

A real A100 reports `0 / 0 / 0` for both stat blocks and a real accounting buffer size.

Confirmed with `MOCK_NVML_DEBUG=1`, which shows nvidia-smi asking for each of these once per GPU and getting nothing:

```
   8 nvmlDeviceGetFBCStats called (NOT IMPLEMENTED)
   8 nvmlDeviceGetEncoderStats called (NOT IMPLEMENTED)
   8 nvmlDeviceGetAccountingBufferSize called (NOT IMPLEMENTED)
```

## Expected behaviour

- Each of the eight exports delegates to its existing engine method, the way other hand-written bridge functions in `pkg/gpu/mocknvml/bridge/device.go` do.
- `nvidia-smi -q` reports the configured `encoder_stats` and `fbc_stats` values, and a real accounting buffer size.
- More generally: a value set in a profile is either surfaced or rejected at load time, never accepted and silently dropped.

## Steps to reproduce

```bash
  nvml-mock:latest nvidia-smi -q | grep -A 3 -E "Encoder Stats|FBC Stats"

# ground truth for the unanswered calls
```

## Environment

```yaml
```

## Acceptance criteria

- [ ] All eight exports are hand-written in the bridge, delegate to their existing engine methods, and are removed from `stubs_generated.go`.
- [ ] Each keeps the established bridge conventions: `bridgeVersionCheck` for driver-version gating, `NVML_ERROR_INVALID_ARGUMENT` on a nil out-parameter or unknown handle, and `toReturn` for engine returns.
- [ ] `nvidia-smi -q` reports the configured `encoder_stats` and `fbc_stats` values rather than `N/A`, and non-default values set in a profile are visible in the output.
- [ ] A regression guard that fails when an engine method exists but its export is still a generated stub, so this class of gap cannot silently reappear.
- [ ] **Dedicated e2e test.** A new Ginkgo spec under `tests/e2e/go/` with `Label("nvidia-smi")` that deploys a profile with non-zero `encoder_stats` and `fbc_stats` (for example `session_count: 2`, `average_fps: 30`, `average_latency_us: 1500`) and asserts `nvidia-smi -q` reports exactly those numbers in the `Encoder Stats` and `FBC Stats` blocks, and that `Accounting Mode Buffer Size` is a number rather than `N/A`. Using non-default values is required so the test cannot pass by coincidence against a zeroed default. The test must fail against the current `main`.

## Out of scope

giuliocalzo:
## Summary
- Wire eight NVML device exports that already had engine implementations but were still generated stubs (`GetEncoderStats`, `GetFBCStats`, `GetAccountingBufferSize`, `GetEncoderCapacity`, `GetEncoderSessions`, `GetFBCSessions`, `GetRetiredPages_v2`, `GetViolationStatus`), so configured `encoder_stats` / `fbc_stats` and the accounting buffer size reach `nvidia-smi -q` instead of silently reporting `N/A`.
- Add a regression guard that fails when a `ConfigurableDevice` method is left behind a `stubReturn` export.
- Add a standalone e2e assertion that pins non-zero encoder/FBC stats via `nvml-mock-ctl`, unmarshals `nvidia-smi -q -x` into typed XML structs, and validates every GPU without regex matching human-readable output.
- Changelog entry under `[Unreleased]`.

Fixes #636

## Test plan
- [x] `go test ./pkg/gpu/mocknvml/bridge/ ./pkg/gpu/mocknvml/engine/ ./tests/e2e/go/assertions/ -count=1`
- [x] `go run ./cmd/generate-bridge --validate`
- [x] `go test -c -tags e2e ./tests/e2e/go/`
- [ ] E2E: `make e2e E2E_PROFILES=a100` (or focused `nvidia-smi` label) confirms Encoder/FBC Stats and Accounting Mode Buffer Size match the non-zero override


roma-glushko (approved):
LGTM 👍

PR #648, opened 2026-08-12
ci(e2e): cancel in-progress Go E2E runs on the same ref

giuliocalzo:
## Summary
- Add a `concurrency` group to `nvml-mock-e2e-go.yaml` keyed by workflow + ref
- Set `cancel-in-progress: true` so newer runs cancel pending/in-progress ones (same pattern as `helm.yaml`)

## Test plan
- [ ] Open/update a PR that triggers Go E2E and push a second commit while the first run is pending or in progress
- [ ] Confirm the earlier workflow run is cancelled and the latest run proceeds


roma-glushko (approved):
Great idea 👍

PR #651, opened 2026-08-12
fix(mocknvml): surface JPEG and OFA utilization in nvidia-smi

### LINKED ISSUE #637: [Bug]: utilization.jpeg and utilization.ofa are parsed but never surfaced by nvidia-smi
## Component / version

## What happened

Every profile configures JPEG and OFA (optical flow accelerator) utilization, and both keys are parsed into the config struct — but there is no device getter for either, and both C exports are generated stubs. The configured values are accepted and then dropped, so `nvidia-smi` reports `N/A` for the two rows.

This is the sibling of the encoder and FBC stats gap, but with a different cause: those have engine methods hidden behind stubbed exports, whereas JPEG and OFA have no engine method at all.

### Example output

`profiles/a100.yaml` configures:

```yaml
  utilization:
    gpu: 0
    memory: 0
    encoder: 0
    decoder: 0
    jpeg: 0
    ofa: 0                          # Optical flow accelerator
```

`nvidia-smi -q` surfaces the first four and drops the last two:

```
    Utilization
        GPU                               : 21 %
        Memory                            : 22 %
        Encoder                           : 0 %
        Decoder                           : 0 %
        JPEG                              : N/A
        OFA                               : N/A
```

A real A100 reports `0 %` for both, since it has NVJPG and OFA engines.

`MOCK_NVML_DEBUG=1` confirms nvidia-smi asks once per GPU and gets nothing:

```
   8 nvmlDeviceGetOfaUtilization called (NOT IMPLEMENTED)
   8 nvmlDeviceGetJpgUtilization called (NOT IMPLEMENTED)
```

For reference, `Encoder` and `Decoder` work because `nvmlDeviceGetEncoderUtilization` and `nvmlDeviceGetDecoderUtilization` are implemented — the two new getters should follow those directly.

## Expected behaviour

- `nvmlDeviceGetJpgUtilization` and `nvmlDeviceGetOfaUtilization` return the configured `utilization.jpeg` and `utilization.ofa` values, alongside a sampling period, matching the shape of the existing encoder and decoder getters.
- `nvidia-smi -q` reports `JPEG : 0 %` and `OFA : 0 %` for a profile configuring zero, and the configured percentage otherwise.
- Architectures with no JPEG or OFA engine may still report `NOT_SUPPORTED`, if the profile set is extended to include such hardware.

## Steps to reproduce

```bash
  nvml-mock:latest nvidia-smi -q -d UTILIZATION
```

## Environment

```yaml
```

## Acceptance criteria

- [ ] `ConfigurableDevice` gains JPEG and OFA utilization getters that read `utilization.jpeg` and `utilization.ofa`, modeled on the existing encoder and decoder getters.
- [ ] `nvmlDeviceGetJpgUtilization` and `nvmlDeviceGetOfaUtilization` are hand-written in the bridge and removed from `stubs_generated.go`.
- [ ] Unit tests cover both getters, including the case where the `utilization` block is absent from the config.
- [ ] **Dedicated e2e test.** A new Ginkgo spec under `tests/e2e/go/` with `Label("nvidia-smi")` that deploys a profile setting `utilization.jpeg` and `utilization.ofa` to distinct non-zero values (for example `jpeg: 35`, `ofa: 12`) and asserts `nvidia-smi -q -d UTILIZATION` reports exactly `JPEG : 35 %` and `OFA : 12 %`. Distinct non-zero values are required so the test cannot pass against a zeroed default or by reading the wrong field. The test must fail against the current `main`.

## Out of scope

giuliocalzo:
## Summary

`utilization.jpeg` and `utilization.ofa` were parsed into the device config and then dropped, so `nvidia-smi` reported `N/A` for both on hardware that has NVJPG and OFA engines. `ConfigurableDevice` had no getter for either, and `nvmlDeviceGetJpgUtilization` / `nvmlDeviceGetOfaUtilization` were generated stubs returning `NVML_ERROR_NOT_SUPPORTED`.

- Add `GetJpgUtilization` and `GetOfaUtilization`, each reading its config field, modeled on the encoder and decoder getters beside them.
- Hand-write the two C exports in the bridge; the generator drops their stubs.
- The values are hot-reloadable through the existing config-override path, so `nvml-mock-ctl set --gpu <idx> utilization.jpeg=35 utilization.ofa=12` works without a restart.

Before and after, same profile (`a100`, which configures `jpeg: 0` / `ofa: 0`):

```
<jpeg_util>N/A</jpeg_util>   →   <jpeg_util>0 %</jpeg_util>
<ofa_util>N/A</ofa_util>     →   <ofa_util>0 %</ofa_util>
```

Fixes #637

### Also fixed: the internal export-table shim wrote through a pointer it was never given

Found while moving the assertion onto `nvidia-smi -q -x`, which faulted before reaching the utilization block. Every slot of the table nvidia-smi fetches pointed at one stub, so the process-list call had to be recognised from the shape of its arguments. Slots that take fewer arguments leave a stale register where the stub read the array pointer, and any configured `processes:` entry was then written through it:

```
main.writeProcessEntry(...)
	/src/pkg/gpu/mocknvml/bridge/internal.go:203
main.mockInternalFillProcessList(0x217a7ea0, 0xffffa84a7a54, 0xffff)
```

On arm64 that faulted, so `nvidia-smi -q`, `-q -x` and `--id=N -q` all died with a SIGSEGV mid-output as soon as a process was configured; on amd64 (what CI runs) the write landed on whatever the stale value addressed, which is why this went unnoticed. Each slot now gets its own trampoline, so a call is attributed to the entry point nvidia-smi asked for, and only the three slots that carry a process array are filled. Slots 213-215 (process lists) and 81 (device handle by index) were identified by logging which slots each nvidia-smi view hits.

### Also: one XML schema for every nvidia-smi assertion

The encoder/FBC checks from #647 and the new JPEG/OFA and process checks each decoded `-q -x` into their own types. They now share one schema, one decode entry point and one scalar reader, so an assertion that needs another element adds a field rather than a parser. `nvidiasmi_temperature.go` keeps its text-table parsing, which exists because the defect it guards is which row labels nvidia-smi prints at all.

## Test plan

Each layer was watched failing before it was implemented.

- [x] `pkg/gpu/mocknvml/engine`: unit tests for both getters — configured values, and a config carrying no `utilization` block. Without the getters they panic on the embedded mock's nil method.
- [x] `tests/mocknvml`: the docker-based harness drives both calls over the real C ABI against a fixture setting `jpeg: 35` / `ofa: 12`. With the bridge exports removed it fails with `GetJpgUtilization -> 0%, ERROR_NOT_SUPPORTED`.
- [x] `tests/mocknvml`: new export-table tests call the table's function pointers the way nvidia-smi does. The process-list slots must report the configured process; **every other slot must leave a guard buffer untouched and report zero entries**, which fails against the previous dispatch on any architecture (`slot 1 wrote into a buffer it was never given`).
- [x] `tests/e2e/go/assertions`: the XML parser is unit-tested against output captured from three real images — current, pre-fix (`N/A` on all 8 GPUs) and truncated — plus transposed and zeroed expectations, missing elements and an out-of-range GPU index.
- [x] `tests/e2e/go/profile`: new `JPEGUtilizationPct()` / `OFAUtilizationPct()` accessors are pinned against a fixture with distinct non-zero values, since every shipped profile configures 0 for both.
- [x] New e2e spec `Label("nvidia-smi")` asserts the deployed profile's own percentages, then pins distinct non-zero values (35 / 12) at runtime and reads both back. Distinct values are deliberate: equal or zeroed ones would pass even if a reading came from the wrong field.
- [x] The e2e process spec also checks the unscoped `nvidia-smi -q -x` view, the shape that faulted.
- [x] Verified end to end against the real `nvidia-smi` in the built image: `35 %` / `12 %` on all 8 GPUs, `<pid>4242</pid>` present, and `nvidia-smi`, `-q`, `-q -x`, `topo -m` and `--query-compute-apps` all exit 0 both with and without configured processes.
- [x] `make lint-fix` clean, `make test` green, `make test-mocknvml-bridge` green.
- [ ] Go E2E workflow (not run locally; the specs above are the new coverage).

## Notes for reviewers

- Assertions read `-q -x` rather than the `-q` table: the XML is nvidia-smi's machine-readable contract, so they key off DTD element names instead of column positions. `-d` cannot be combined with `-x`, so the query is plain `-q -x` and only stdout is parsed.
- The e2e pins the distinct non-zero pair through `nvml-mock-ctl set` rather than a second Helm install: a mid-suite `helm upgrade` rolls the DaemonSet and would invalidate the pod reference later specs in that ordered context still use. Both paths go through the same config merge the getters read.
- The sampling period returns 0, mirroring the existing encoder and decoder getters. `nvidia-smi` never displays it; real hardware returns a non-zero period, so this is a consistency choice rather than fidelity.
- Encoder and decoder utilization still return a hardcoded 0 rather than reading `utilization.encoder` / `utilization.decoder`. Out of scope here, and invisible while every profile configures 0.
- The export-table fix can be split into its own PR if you would rather review it separately; it is a self-contained commit.


roma-glushko (approved):
Looks good to me as much as I understand what's going on 😃

PR #652, opened 2026-08-13
fix(mocknvml): derive Board ID and report Device/Host Max PCIe generation

### LINKED ISSUE #638: [Bug]: PCIe Generation Host Max reports 0 and Board ID reports 0x0
## Component / version

## What happened

Two scalar identity fields report values that no real GPU would produce. They are grouped here because each is a single-value fix in one place; split the issue if that is preferred.

**1. `PCIe Generation → Host Max : 0`.** `nvmlDeviceGetGpuMaxPcieLinkGeneration` is a generated stub. Rather than rendering the failure as `N/A`, nvidia-smi prints its zero-initialised buffer, so the field reads `0` — a PCIe generation that does not exist. Note the inconsistency within the same block: `Max` and `Current` report `4` correctly, `Device Max` degrades honestly to `N/A`, and only `Host Max` invents a `0`.

**2. `Board ID : 0x0`.** Real GPUs report a non-zero board ID derived from the PCI topology, so a fleet of eight mock GPUs is indistinguishable by board ID.

### Example output

`profiles/a100.yaml`, GPU 0:

```
    PCI
        Bus                               : 0x07
        Device                            : 0x00
        Domain                            : 0x0000
        ...
        GPU Link Info
            PCIe Generation
                Max                       : 4
                Current                   : 4
                Device Current            : 4
                Device Max                : N/A
                Host Max                  : 0
            Link Width
                Max                       : 16x
                Current                   : 16x
```

and:

```
    Board ID                              : 0x0
```

A real A100 reports `4` for `Host Max` and a non-zero `Board ID`.

`MOCK_NVML_DEBUG=1` confirms the unanswered call:

```
   8 nvmlDeviceGetGpuMaxPcieLinkGeneration called (NOT IMPLEMENTED)
```

## Expected behaviour

- `nvmlDeviceGetGpuMaxPcieLinkGeneration` returns the profile's configured maximum PCIe generation, so `Host Max` reads `4` on `a100` and matches `max_link_gen` on other profiles (for example `6` on `gb200`).
- `Board ID` is non-zero and distinct per device, derived from the configured PCI bus and device numbers the way real NVML does.
- No `-q` field reports a value outside its physically valid range.

## Steps to reproduce

```bash
  nvml-mock:latest nvidia-smi -q | grep -E "Board ID|Host Max|Device Max|PCIe Generation" -A 1
```

## Environment

```yaml
```

## Acceptance criteria

- [ ] `nvmlDeviceGetGpuMaxPcieLinkGeneration` is hand-written in the bridge, reads the profile's configured max PCIe generation, and is removed from `stubs_generated.go`.
- [ ] `Board ID` is non-zero and unique per device across an eight-GPU profile, derived from the configured PCI bus and device numbers.
- [ ] Unit tests cover both values, including a profile that does not configure `max_link_gen`.
- [ ] **Dedicated e2e test.** A new Ginkgo spec under `tests/e2e/go/` with `Label("nvidia-smi")` that asserts, for every GPU in the profile: `Host Max` equals the profile's configured PCIe generation and is never `0`; `Board ID` is non-zero; and the set of board IDs across all GPUs contains no duplicates. Running it against both `a100` (Gen4) and `gb200` (Gen6) pins that the value comes from config rather than a hardcoded constant. The test must fail against the current `main`.

## Out of scope

The roughly 60 `nvidia-smi -q` fields that are faithfully `N/A` on this hardware — Windows driver model, vGPU and GRID licensing, Confidential Compute protected memory, SXM fan speed, bridge chip, and A100 auto-boost. `Device Max : N/A` is left as-is here, since degrading to `N/A` is honest behaviour rather than a wrong value.

giuliocalzo:
## What This PR Does

`nvidia-smi` reported three per-GPU PCIe identity values no real GPU can produce. All three now track the profile:

| Reading | Before | After |
| --- | --- | --- |
| `Board ID` | `0x0` on every GPU | derived from the PCI address (`0000:07:00.0` → `0x700`) |
| `Device Max` PCIe generation | `N/A` | the profile's `pcie.max_link_gen` |
| `Host Max` PCIe generation | `Gen0` | the profile's `pcie.max_link_gen` |

- **Board ID** was hardcoded to 0. It is now derived the way NVML does — `(domain << 16) | (bus << 8) | (device << 3) | function` — so the eight GPUs of a DGX node no longer all claim the same board.
- **Device Max** came from `nvmlDeviceGetGpuMaxPcieLinkGeneration`, a generated stub returning `NVML_ERROR_NOT_SUPPORTED`. It is now hand-written in the bridge and reads config, so the generation tracks the profile: Gen3 on `t4` through Gen6 on Blackwell.
- **Host Max** has no public NVML API behind it — none of the nine `nvmlDeviceGet*Pcie*` functions is host-scoped, which is why implementing the generation getters never moved the row. nvidia-smi reads it through **slot 230 of the internal export table**, where the catch-all C stub wrote a zero count over the caller's reading. That is the same class of bug as the phantom processes fixed in #630: a stub acknowledging a call it does not understand and clobbering the output. The bridge now serves that slot from `pcie.max_link_gen`, modelling a host that keeps up with the GPU, so `Max`, `Device Max` and `Host Max` agree. Profiles with no `pcie` block stay on the existing zero path rather than having a generation invented for them.

The slot was identified by its position in the `-q` call sequence under `MOCK_NVML_DEBUG` — it fires between `nvmlDeviceGetGpuMaxPcieLinkGeneration` (`Device Max`) and `nvmlDeviceGetMaxPcieLinkWidth` (`Link Width`), exactly where `Host Max` prints:

```
[NVML]   nvmlDeviceGetGpuMaxPcieLinkGeneration -> 4  # Device Max
[C-STUB] slot 230 device list query -> 0 entries     # Host Max  <-- clobbered here
[NVML]   nvmlDeviceGetMaxPcieLinkWidth -> 16         # Link Width Max
```

Writing a marker value into the slot moved the `Host Max` row of `-q`, the `<max_host_link_gen>` element of `-q -x` and the `pcie.link.gen.hostmax` query field together, confirming all three surfaces come from it.

### Why a bridge test guard changed

`tests/mocknvml` guarded every internal slot outside the process-list range with a single blanket rule: leave the caller's buffer alone and write a **zero** count. That is correct for a list count and wrong for a scalar getter, so slot 230 now has its own test — returns SUCCESS, reports the fixture's Gen4, guard buffer untouched — and the blanket rule keeps covering every slot whose meaning is still unknown.

That guard exists for a fault that appeared "as soon as a process was configured", so the change was re-checked against `tests/mocknvml/util-test-config.yaml`, which configures a process alongside `max_link_gen: 4`. The only difference in full `nvidia-smi -q` output is the `Host Max` row, `--query-compute-apps` is unchanged with no phantom rows, and `pmon` fails identically before and after.

### E2E coverage

New spec `reports valid per-GPU PCIe identity via nvidia-smi -q -x` in the standalone scenario, backed by the pure `PCIeIdentityProblems` check in the `tests/e2e/go/assertions/nvidiasmi` package (#658). It requires all three maxima to equal the profile's configured `pcie.max_link_gen` and board IDs to be non-zero and unique across the node. `board_id` and the `pci_gpu_link_info/pcie_gen` subtree join `schema.go`, the one place that package names elements, so its tri-state reading keeps an absent element distinguishable from `N/A` and from a genuine `0`.

The captured fixtures in `testdata/` predate the host-max fix and carry `max_host_link_gen` `0`, so the passing case is hand-built — the convention that package already documents for the JPEG/OFA and encoder/FBC checks. The `a100` capture still drives a test, which pins every other element path against a real document; recapturing the fixtures after this merges would let it tighten to "no problems".

## Why

Fixes #638.

A `Board ID` of `0x0` on every device leaves the GPUs of a multi-GPU node indistinguishable to anything keying on it. A `Gen0` host maximum beside a Gen4 device maximum describes a topology that cannot exist, since a link never negotiates higher than either endpoint supports. Both undercut the point of the mock: consumers up the stack should see what they would see on real hardware.

## Checklist

- [x] Tests pass — `make test` (which runs with `-race`) and `make test-mocknvml-bridge` at 43/43
- [x] Linter passes (`make lint-fix`) — 0 issues, govulncheck clean
- [ ] Documentation updated (if applicable) — n/a

### Verification

- `Host Max` matches each profile's configured generation against built images: `t4` Gen3, `a100`/`l40s` Gen4, `h100` Gen5, `b200`/`gb200` Gen6.
- Diffing full output against a pre-fix image on `a100`: the only structural change is `Host Max` `0` → `4` on all eight GPUs plus the corresponding XML element. `topo -m` is byte-identical, and `-L`, `-q`, `-q -x`, `-q -d MEMORY`, `--list-gpus`, `--query-compute-apps` all still exit 0.
- Running the real `PCIeIdentityProblems` check over live captures from an image built from this branch: clean on `a100` (Gen4) and `gb200` (Gen6), eight GPUs each, and a Gen5 expectation fails on the same document as a negative control.
- The new bridge test was checked for vacuity by pointing the implementation at a neighbouring slot: it reports `slot 230 reported Gen0, want the configured Gen4`, and the blanket guard independently catches the stray non-zero write.
- CI green, including the full e2e matrix across `a100`, `h100`, `b200`, `gb200`, `gb300` and `t4`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `FORCE_RECREATE=true`.
 
 ### Fixed
+- mocknvml: `nvidia-smi` no longer reports impossible per-GPU PCIe identity
+  values. `Board ID` is derived from the device's PCI address the way NVML does
+  — `(domain << 16) | (bus << 8) | (device << 3)`, so a GPU at `0000:07:00.0`
+  reports `0x700` — instead of `0x0` for every GPU, which left an eight-GPU node
+  indistinguishable by board ID. `nvmlDeviceGetGpuMaxPcieLinkGeneration` is now
+  hand-written in the bridge, so the `Device Max` PCIe generation reads Gen3 on
+  `t4` through Gen6 on Blackwell instead of `N/A`. `Host Max` now matches the
+  device maximum instead of an impossible Gen0: no public NVML API exposes a
+  host-side maximum, and nvidia-smi reads it through a slot of the internal
+  export table whose catch-all stub wrote a zero count over the caller's
+  reading — the same class of bug as the phantom processes above. All three
+  maxima now agree in `nvidia-smi -q`, `-q -x` (`<max_host_link_gen>`) and
+  `--query-gpu=pcie.link.gen.hostmax`. (#638)
 - Wire eight NVML device exports that already had engine implementations but
   were still generated stubs (`GetEncoderStats`, `GetFBCStats`,
   `GetAccountingBufferSize`, `GetEncoderCapacity`, `GetEncoderSessions`,
--- a/pkg/gpu/mocknvml/bridge/device.go
+++ b/pkg/gpu/mocknvml/bridge/device.go
@@ -78,6 +78,7 @@
 // - nvmlDeviceGetCurrPcieLinkGeneration
 // - nvmlDeviceGetCurrPcieLinkWidth
 // - nvmlDeviceGetMaxPcieLinkGeneration
+// - nvmlDeviceGetGpuMaxPcieLinkGeneration
 // - nvmlDeviceGetMaxPcieLinkWidth
 // - nvmlDeviceGetPcieReplayCounter
 // - nvmlDeviceGetPcieThroughput
@@ -1887,6 +1888,27 @@ func nvmlDeviceGetMaxPcieLinkGeneration(device C.nvmlDevice_t, maxLinkGen *C.uin
 		return toReturn(ret)
 	}
 	*maxLinkGen = C.uint(val)
+	return C.NVML_SUCCESS
+}
+
+//export nvmlDeviceGetGpuMaxPcieLinkGeneration
+func nvmlDeviceGetGpuMaxPcieLinkGeneration(device C.nvmlDevice_t, maxLinkGenDevice *C.uint) C.nvmlReturn_t {
+	if ret, ok := bridgeVersionCheck("nvmlDeviceGetGpuMaxPcieLinkGeneration"); !ok {
+		return ret
+	}
+	if maxLinkGenDevice == nil {
+		return C.NVML_ERROR_INVALID_ARGUMENT
+	}
+	handle := unsafe.Pointer(device.handle)
+	dev := engine.GetEngine().LookupConfigurableDevice(handle)
+	if dev == nil {
+		return C.NVML_ERROR_INVALID_ARGUMENT
+	}
+	val, ret := dev.GetGpuMaxPcieLinkGeneration()
+	if ret != nvml.SUCCESS {
+		return toReturn(ret)
+	}
+	*maxLinkGenDevice = C.uint(val)
 	return C.NVML_SUCCESS
 }
 
--- a/pkg/gpu/mocknvml/bridge/internal.go
+++ b/pkg/gpu/mocknvml/bridge/internal.go
@@ -37,6 +37,10 @@ extern int mockInternalIsDeviceHandle(void* handle);
 // written. The entry layout is documented on the Go side.
 extern unsigned int mockInternalFillProcessList(void* handle, void* buf, unsigned int capacity);
 
+// Forward declaration of the host-side max PCIe link generation lookup (defined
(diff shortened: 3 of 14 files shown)


roma-glushko (approved):
Looks good to me 👍

PR #654, opened 2026-08-14
chore: bump Go pins 1.26.5 -> 1.26.6

giuliocalzo:
## Summary

`golang / check` currently fails on **every** branch, including `main`, at the `govulncheck` step in `make lint`. go1.26.5 is affected by seven standard-library advisories that govulncheck reports as reachable from this code, all fixed in 1.26.6:

| Advisory | Package |
| --- | --- |
| GO-2026-6218 | `net/url` |
| GO-2026-6091 | `html/template` |
| GO-2026-6090 | `crypto/tls` |
| GO-2026-6089 | `net/http` |
| GO-2026-5026 | `net/http` |
| GO-2026-6088 | `encoding/xml` |
| GO-2026-5972 | `encoding/asn1` |

The traces are ordinary pre-existing code, e.g. `pkg/diagnostics/kubernetes.go` reaching `http.Client.Do` and `pkg/gpu/mockcuda/engine/cuda.go` reaching `asn1.Unmarshal`.

CI resolves its toolchain from `deployments/devel/Dockerfile` via `hack/golang-version.sh`, so the stale pin breaks lint regardless of what a PR changes. `main`'s own `basic / golang / check` passed on Aug 13 and only started failing once the advisories were published — nothing in the repo changed.

This bumps every place the version is pinned, so the deployment images, demo and test Dockerfiles, and the `mocknvml`/`mockcuda` Makefile defaults stay in step. Same approach as the 1.26.4 -> 1.26.5 bump.

## Test plan

- [x] `hack/golang-version.sh` reports `1.26.6` (the value CI feeds every job)
- [x] `govulncheck -tags=e2e,integration ./...` under go1.26.6: **0 vulnerabilities**, down from 7
- [x] `make test` — all 30 packages pass
- [x] `golang:1.26.6` and `golang:1.26.6-bookworm` tags exist upstream
- [x] `docker build -f deployments/nvml-mock/Dockerfile` succeeds on 1.26.6, so the CGO NVML bridge still compiles
- [x] `nvidia-smi -L` in the resulting image lists the profile GPUs, confirming the built `libnvidia-ml.so` works
- [ ] Full e2e suite (CI)

Note: the `nvml-mock-e2e` jobs are failing on `main` for a separate, unrelated reason — all 24 legs fail there. This PR does not address that.


roma-glushko (approved):
Thanks for being on top of this!

Issue #566, opened 2026-07-31
ci(e2e): a single ttl.sh 502 fails the whole e2e matrix — no retry on push or pull


ArangoGutierrez:
## Summary

The e2e image pipeline depends on ttl.sh, a free anonymous registry, with no retry. A single 502 from it fails the whole run.

## Observed

2026-07-31, PR #563 attempt 1. Every job passed except `build-kind-node-image`, which failed with ttl.sh returning **502 on every blob upload**. Attempt 2 succeeded with no code change. The failure had nothing to do with the PR.

## Why the dependency exists

ttl.sh was chosen deliberately and the reasoning is still sound: it is auth-free, so **fork PRs can push to it**. GHCR cannot serve that case without credentials that fork PRs must not have. `E2E_IMAGE` is `ttl.sh/nvml-mock-${{ github.sha }}:24h`, built once and pulled by digest across every leg (#465, issue #460).

So this is not an argument for dropping ttl.sh. It is an argument that a best-effort free service is currently a hard, single-attempt dependency for the entire e2e matrix.

## What to consider

1. **Retry with backoff on push and pull.** Cheapest change, addresses the observed failure directly. A 502 on blob upload is transient by nature.
2. **Fall back to a second registry** when ttl.sh is unavailable. More work, and the fork-PR constraint limits the options.
3. **Split the paths**: GHCR for same-repo PRs and pushes to `main` where credentials exist, ttl.sh only for fork PRs. Reduces exposure without giving up fork coverage.

Option 1 is the obvious first move and may be sufficient. Option 3 is worth costing if ttl.sh flakes again.

## What NOT to conclude

Do not read this as "ttl.sh is unreliable, replace it." One 502 is one data point. Before doing structural work, check how often e2e runs have failed on registry errors — if this is the first in months, a retry is the whole fix and anything more is over-engineering.

## Acceptance

- A transient registry error does not fail the e2e matrix on the first attempt.
- If a fallback or split is implemented, fork PRs still work without credentials.

## References

- #465 / #460 — established build-once-share-via-ttl.sh
- PR #563 — attempt 1 failed solely on this; attempt 2 passed unchanged

roma-glushko:
I have noticed this interim failures as well. I think it might be better to go with image local registry like tilt suggests:
https://docs.tilt.dev/choosing_clusters.html

Slack reference: https://nvidia.slack.com/archives/C0AM75KBXV1/p1785499619107829

[...]

PR #655, opened 2026-08-14
ci(e2e): hand off prebuilt e2e images as artifacts instead of ttl.sh

giuliocalzo:
## What This PR Does

Distributes the two prebuilt e2e images through run-scoped GitHub Actions artifacts instead of pushing them to, and pulling them from, `ttl.sh`.

- `build-nvmlmock-image` and `build-kind-node-image` build with `outputs: type=docker,dest=...` and upload the tarball. The six nvml-mock legs and the standalone `e2e` leg download it and stage it with `make image-load TARBALL=<tar> IMAGE=<repo:tag>`, a new root Makefile target that loads the tarball, asserts the expected ref reached the daemon, and deletes the tarball so a `kind load` right after has disk headroom.
- `E2E_IMAGE` becomes `nvml-mock:e2e`. The digest-pinning plumbing is gone: the tarball carries its own tag, and an artifact is immutable and scoped to its run, so there is no tag-overwrite surface left to defend against.
- Tarball basenames live in workflow `env`, so the producing job and its seven consumers cannot drift apart on a filename.
- `deployments/kind-nvidia-cdi/Makefile` drops its `ttl.sh` registry default. `make build` now tags `kind-nvidia-cdi:local` in the local Docker daemon, which is all `kind create cluster --image` reads, and `make push` requires a registry-qualified `IMAGE` — checked before the long build rather than after.
- Comments and docs updated to match: `tests/e2e/README.md`, root `Makefile`, `Tiltfile`, `local/nvml_mock.tiltfile`, `CHANGELOG.md`. `ttl.sh` now appears only in the changelog, where the historical entry stays as-is.

Build-once is preserved: each image is still built exactly once per run and shared with every consumer.

## Why

`ttl.sh` is an anonymous third-party registry sitting in the path of every e2e leg, so when it is slow or unavailable the whole matrix fails for reasons unrelated to the change under test (see #566). Artifacts are first-party and need no credentials, which also means fork PRs keep working without secrets. Retention is one day — the shortest GitHub allows, and still long enough to re-run a single failed leg.

Keeping the load logic in a Makefile target rather than a composite action follows this repo's CI convention of driving pipelines through `make`, so the same step is reproducible on a laptop.

## Validation

- This PR's own CI is green across the full matrix: both image build jobs, plus the `e2e`, `e2e-dra`, `e2e-gpu-operator`, `e2e-nri` legs on all six GPU profiles and `e2e-nfd` / `e2e-multi-node`. The `e2e` legs cover the kind-node artifact path, since they load that image and then run `make cluster-create` with `KIND_NODE_IMAGE_PREBUILT`.
- An earlier `workflow_dispatch` run on a fork confirmed the hand-off end to end from the logs: six `nvml-mock` artifact loads plus the `kind-node` load, and zero `ttl.sh` references. https://github.com/giuliocalzo/k8s-test-infra/actions/runs/31786985126
- `make image-load` was also exercised against Docker locally: the happy path loads the image and removes the tarball; a mismatched expected ref fails with a diagnostic and keeps the tarball; missing arguments print usage; a missing file fails from `docker load`.

## Checklist

- [x] Tests pass (`go test -v -race ./...`) — no Go code touched
- [x] Linter passes (`make lint-fix`) — no Go code touched
- [ ] New code has SPDX license headers — n/a


faganihajizada (approved):
Looks good

PR #657, opened 2026-08-14
fix(mocknvml): report GPU C2C Mode in nvidia-smi -q

### LINKED ISSUE #639: [Bug]: GPU C2C Mode is N/A on the gb200 profile despite c2c_enabled: true
## Component / version

## What happened

The `gb200` profile explicitly declares NVLink-C2C to the Grace CPU, and the config struct has a field for it, but `nvmlDeviceGetC2cModeInfoV` is a generated stub — so `nvidia-smi -q` reports `GPU C2C Mode : N/A` on a Grace-Blackwell profile whose defining characteristic is the C2C link.

This is a configured value that is accepted and then silently dropped. On `a100` the same `N/A` is correct, since Ampere has no C2C link; the defect is specific to profiles that declare C2C.

### Example output

`profiles/gb200.yaml` sets:

```yaml
  c2c_enabled: true                 # NVLink-C2C to Grace CPU
```

and, under NVLink:

```yaml
    nvlink_c2c: true                # NVLink Chip-to-Chip (Grace connection)
    coherent_memory: true           # NVLink-C2C coherent memory
```

`nvidia-smi -q` on `gb200` nevertheless reports:

```
    Product Name                          : NVIDIA GB200
    Product Architecture                  : Blackwell
    ...
    GPU C2C Mode                          : N/A
```

`MOCK_NVML_DEBUG=1` confirms the unanswered call:

```
   8 nvmlDeviceGetC2cModeInfoV called (NOT IMPLEMENTED)
```

`C2CEnabled` is present in `pkg/gpu/mocknvml/engine/config_types.go` and parsed from the profile, but nothing reads it for this call.

## Expected behaviour

- On profiles declaring `c2c_enabled: true`, `nvidia-smi -q` reports `GPU C2C Mode : Enabled`.
- On profiles that do not declare it, or on architectures with no C2C link, the field stays `N/A` — which is the correct result for `a100`, `t4`, `h100`, and `l40s`.

## Steps to reproduce

```bash
  -v "$PWD/deployments/nvml-mock/helm/nvml-mock/profiles/gb200.yaml:/config/config.yaml:ro" \
  nvml-mock:latest nvidia-smi -q | grep -E "Product Name|C2C Mode"
```

## Environment

```yaml
```

## Notes on implementation

`nvmlDeviceGetC2cModeInfoV` uses NVML's versioned-struct calling convention, where the caller passes a struct whose first field carries a version constant. The implementation needs the correct struct layout and version handling, which is why this is more involved than a plain scalar getter. The existing versioned calls in the bridge — for example the `GpuFabricInfoV` path, which is already implemented — are the pattern to follow.

## Acceptance criteria

- [ ] `nvmlDeviceGetC2cModeInfoV` is hand-written in the bridge, reads `c2c_enabled` from the device config, and is removed from `stubs_generated.go`.
- [ ] The versioned-struct layout and version field are pinned by a unit test, in the same spirit as the existing process-entry layout test, so a wrong stride or version cannot fail silently.
- [ ] `nvidia-smi -q` reports `GPU C2C Mode : Enabled` on `gb200` and `gb300`, and leaves the field `N/A` on `a100`, `t4`, `h100`, and `l40s`.
- [ ] **Dedicated e2e test.** A new Ginkgo spec under `tests/e2e/go/` with `Label("nvidia-smi")` that asserts `nvidia-smi -q` reports `GPU C2C Mode : Enabled` on a profile with `c2c_enabled: true` (`gb200`) and `N/A` on a profile without it (`a100`). Covering both directions is required so the fix cannot be a hardcoded `Enabled`. The test must fail against the current `main`.

## Out of scope

The rest of the `gb200` platform-identity gap — `Platform Info` (chassis serial, slot, tray, host ID, peer type, module ID) — which needs a new config surface and is tracked separately. Also out of scope: the roughly 60 `-q` fields that are faithfully `N/A` on this hardware.

giuliocalzo:
## Summary

`nvidia-smi -q` reported `GPU C2C Mode : N/A` on the Grace-Blackwell profiles even though `gb200`/`gb300` declare `nvlink.c2c_enabled: true`, because `nvmlDeviceGetC2cModeInfoV` was an auto-generated "not implemented" stub. The engine now resolves the flag into node-level fabric facts and a hand-written cgo export returns it.

Node-level rather than per-device: the C2C link is a property of the board, so every GPU on a profile that has it reports it.

| Profile | `c2c_enabled` | `GPU C2C Mode` |
|---|---|---|
| `gb200`, `gb300` | `true` | `Enabled` |
| `a100`, `h100`, `b200` | `false` | `N/A` |
| `l40s`, `t4` | absent | `N/A` |

`false` and an absent key are deliberately identical: both answer `NVML_ERROR_NOT_SUPPORTED`, which renders as `N/A`. Returning `SUCCESS` with a false reading would render `Disabled`, which no shipped profile wants — `a100`, `h100` and `b200` set `false` explicitly and must stay `N/A`.

### One correction to the issue's premise

The issue assumed this entry point takes a versioned struct like `nvmlDeviceGetGpuFabricInfoV`. It does not: despite the `V` suffix, the vendored header defines `nvmlC2cModeInfo_v1_t` as a single `unsigned int` with no version member, and the caller passes no version. So there is deliberately **no** version dispatch here, unlike `bridge/fabric.go`. The struct size is pinned by a `_Static_assert` next to the definition in `nvml_types.h`, so adding a field fails the cgo compile rather than silently making the bridge write past a caller's allocation. A Go layout test mirroring it was dropped on review: it only asserted a property of the vendored go-nvml struct, exercising none of our code.

The e2e expectation is derived from `nvlink.c2c_enabled` in the same chart profile YAML the chart installs, so it cannot drift from configuration or be satisfied by a hardcoded `Enabled`.

### Rebased onto the XML assertion refactor (#658)

`main` now decodes `nvidia-smi -q -x` instead of scraping the text output, so the check moved into the `nvidiasmi` package as `C2CModeProblems` reading the new `c2c_mode` element, and the three `gb200` fixtures move from `N/A` to `Enabled` to match a capture taken from a build carrying this fix.

A failed GPU is skipped rather than compared. `c2c_mode` is answered from a handle-lookup path that deliberately does not tick the failure injector, so a lost device reports either its board's real state or an NVML error body depending on which element `nvidia-smi` asked for first — confirmed in the container, where a GPU injected with `--mode lost` renders `gpu_temp = GPU is lost` while `c2c_mode` still reads `Enabled`. Skipping every GPU is itself reported, so an all-failed document cannot pass for want of anything to compare.

`board_id`/`pci` (#652) and `c2c_mode` now sit together next to `uuid` in `gpuElement`, in the order `nvidia-smi` emits them.

## Test plan

- [x] `make test` — exit 0 across 32 packages. Unit tests pin all three semantic cases (`true`, explicit `false`, absent) plus the nil-fabric path.
- [x] `golangci-lint run ./...` — 0 issues. `make lint-fix` additionally runs `govulncheck`, which flags only stdlib advisories fixed in go1.26.6 (local toolchain is one patch behind).
- [x] `make gen-check` — clean; the C2C stub is gone from `stubs_generated.go` by regeneration, not by hand.
- [x] `make helm-tests` — passing.
- [x] **Verified against the real `nvidia-smi` in the container across all seven shipped profiles** (table above). Every profile prints exactly one `GPU C2C Mode` row per attached GPU; the `NOT IMPLEMENTED` debug line is gone while 177 stub lines from 19 other entry points still appear, so that absence is meaningful rather than logging being off.
- [x] Re-verified after the rebase onto #658: `go vet -tags e2e ./tests/e2e/...` clean, and the `c2c_mode` readings re-captured from the container (`gb200` → `Enabled` on all 8 GPUs, `a100` → `N/A`, `gb200` with GPU 1 lost → `Enabled` / `GPU is lost`).
- [x] The `_Static_assert` was confirmed to fire: adding a field to `nvmlC2cModeInfo_v1_t` fails the cgo compile.
- [ ] e2e suite on a live cluster (CI). The spec derives its expectation from the profile, so the matrix covers both directions — `Enabled` on `gb200`/`gb300`, `N/A` on `a100`/`b200`/`h100`/`t4`.

Fixes #639


docs/configuration.md
```diff
@@ -370,6 +370,14 @@ nvlink:
       remote_pci_bus_id: "0000:0F:00.0"
 ```
 
+`c2c_enabled` is node-level and drives the `GPU C2C Mode` row of `nvidia-smi -q`
```
roma-glushko, line 373: A good resource for other reviewers about C2C:

Image

CLinks between Graces are also C2C-based.

pkg/gpu/mocknvml/bridge/c2c_layout_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+// nvmlC2cModeInfo_v1_t is a single unsigned int — no version tag, despite the
+// V suffix on nvmlDeviceGetC2cModeInfoV. If a future go-nvml bump grows the
+// struct to carry the version field that suffix implies, this size moves and
+// the bridge implementation needs revisiting: it deliberately does no version
+// dispatch today because there is nowhere for a caller to pass a version.
+const expectedC2cModeInfoV1Size uintptr = 4
+
+func TestC2cModeInfoStructLayout(t *testing.T) {
+	require.Equal(t, expectedC2cModeInfoV1Size, unsafe.Sizeof(nvml.C2cModeInfo_v1{}),
```
roma-glushko: Is this a meaningful test?

roma-glushko (dismissed):
Looks good to me 👍

pkg/gpu/mocknvml/bridge/c2c_layout_test.go
giuliocalzo: Not meaningful enough to keep — removed in cf34355c.

It asserted a property of a vendored third-party struct (`nvml.C2cModeInfo_v1` is just `struct { IsC2cEnabled uint32 }`), so it exercised none of our code and could only ever fail on a go-nvml bump. Nothing in this repo calls the C2C API either, so the mismatch it guarded against — go-nvml resizing the struct while our C header stays at 4 bytes, making the bridge write a different number of bytes than a Go caller allocated — would only affect external go-nvml consumers.

The `_Static_assert` next to `nvmlC2cModeInfo_v1_t` in `nvml_types.h` still guards the side that governs what the bridge actually writes, and it fires on real drift (adding a field to the struct fails the cgo compile with `static assertion failed ... must stay a single unsigned int to match the go-nvml ABI`). Its message no longer points at the deleted test, and the note about a struct grown to carry the version field the `V` suffix implies moved there.

PR #658, opened 2026-08-14
test(e2e): assert nvidia-smi state from -q -x instead of scraping text

giuliocalzo:
## What This PR Does

Converts every e2e assertion that reads GPU state to decode the `nvidia-smi -q -x` document instead of matching substrings, regexes or CSV cells, and gives those assertions their own package at `tests/e2e/go/assertions/nvidiasmi`. `-q -x` is nvidia-smi's machine-readable contract, so the checks key off DTD element names (`nvsmi_device_v13.dtd`) rather than the human table's column layout.

The load-bearing piece is a tri-state `reading` type. nvidia-smi renders four different things in the same element — an element the architecture never emits, a supported-but-`N/A` query, an NVML error body like `GPU is lost`, and a value — and the old code collapsed them.

`nvidiasmi.Snapshot` is one decoded document with typed readings hanging off it, replacing a dozen separate `--query-gpu` execs: a scenario that read four fields made four execs and parsed four cells, and now makes one. Nesting also resolves element names that repeat across blocks, which text parsing could not — `clocks.sm` is `<clocks><sm_clock>` and not the `max_clocks` sibling, `memory.used` is the framebuffer and not `bar1`, and the power readings come from `<gpu_power_readings>` rather than the `N/A` module and memory blocks.

The package is layered, one concern per file:

| File | Owns |
| --- | --- |
| `schema.go` | decoding, and the only place element names appear |
| `snapshot.go` | `Snapshot` and `GPU` — the decoded readings |
| `checks.go` | inventory, phantom processes, JPEG/OFA (#637), encoder/FBC (#636) |
| `temperature.go` | the architecture-dependent threshold presentation (#635) |
| `exec.go` | runs nvidia-smi in a pod and asserts through Gomega |

Only `exec.go` carries `//go:build e2e`, so everything else unit-tests without a cluster against real captured documents in `testdata/` (gb200 healthy, gb200 lost, gb200 with injected ECC, a100 healthy), with small inline XML for targeted defect cases. That is what lets the pre-Ada absolute-threshold branch be verified on `make test`.

The pure checks are named `<subject>Problems` because each returns what it found rather than failing: a stubbed run reports every wrong reading in one pass instead of one per fix-and-rerun cycle, and a caller can poll the check inside an `Eventually` until the mock converges. `exec.go` is the layer that turns them into Gomega assertions, for specs that want to fail on the spot.

Scope notes:

- `nvlink -s/-c/-gt/-e` and `topo -m` keep their text parsing, so `assertions/nvlink.go` stays where it is. `-q -x` carries a `<fabric>` block but no per-link NVLink counters, so there is nothing to convert to.
- `countLinesWithPrefix` and `countMatches` stay in `assertions.go`: `ib.go` still parses `ibstat` with them.
- The bare `nvidia-smi` invocation stays as a crash smoke test for the human renderer, exit status only.
- Two helpers with no callers are dropped: `ProcessesXML`, superseded by `GPU.Processes`, and the `EncoderFBCAccounting` Ginkgo wrapper — the standalone spec polls `EncoderFBCProblems` directly, and nothing called the wrapper on `main` either.

One deliberate coverage trade-off: the process scenario's default-table and `nvidia-smi -q` text checks are gone. Both bugs they guarded are still caught, because `<process_info>` carries pid, `process_name` and `used_memory` — the export-table stride bug (which renders every process after the first as PID 0) and the dropped inline name (the name reaches nvidia-smi only through that buffer). What is lost is the human-renderer regression coverage itself.

## Why

The text assertions were quietly unsound, and one was actively wrong. `hasFailureMarker` decided a GPU had failed if the output contained `[n/a]`, `[unknown error]`, `gpu is lost` or `err`; on a passively-cooled profile `fan_speed` is legitimately `N/A`, so a healthy board could be reported as failed, and `err` matches any error word anywhere in the output. Parsing `[GPU is lost]` out of a CSV cell had the mirror-image problem: a failed reading could be read as a clean `0`.

Beyond correctness, substring and regex checks are pinned to nvidia-smi's human layout — a column shift or a renamed row label breaks them for reasons unrelated to what they assert. The XML element names are the documented contract, and they now appear in exactly one file, so a driver bump that renames one is a single-file change.

The separate package follows from that: the schema is shared by nothing else in `assertions`, and while the code sat beside the `ibstat` and DCGM helpers every name needed a `NvidiaSMI` or `XML` affix to stay distinct from its neighbours. Scoping it turns `assertions.DiffTemperatureXML` into `nvidiasmi.TemperatureProblems`, and the build-tag split that keeps the pure checks unit-testable is now visible in the layout rather than repeated as a comment in each file.

## Checklist

- [x] Commits are signed off (`git commit -s`) and signed — DCO passes; the branch is rebased on `main`, so there is no inherited merge commit left
- [x] Tests pass (`go test -v -race ./...`) — via `make test`; `pkg/gpu/allocwatch` and `pkg/network/mockib/daemon` need to run outside a filesystem sandbox, since they bind unix sockets
- [x] Linter passes (`make lint-fix`) — 0 issues. `govulncheck` reports 7 pre-existing standard-library findings via `pkg/diagnostics`, identical on `main`
- [x] Documentation updated (if applicable) — `testdata/README.md` records what each fixture captures and how to recapture it
- [ ] CHANGELOG.md updated (if user-facing change) — n/a, test-side only, no mock behaviour changes

Also verified on a cluster: `make e2e E2E_PROFILES=gb200 E2E_GINKGO_FLAGS='--label-filter="nvidia-smi || failure-injection || runtime-control"'` ran 28 specs in 7m24s with 0 failures, covering the inventory check, temperature thresholds, JPEG/OFA, encoder/FBC, all three failure modes, and the full runtime-control family down to fan, pstate, throttle and the process list. That run predates the package move and the review round, both of which change names, locations and comments only; those are covered by `go vet -tags e2e ./tests/e2e/...` plus the unit suite (47 tests, 91% of the package).


tests/e2e/go/assertions/nvidiasmi/exec.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/NVIDIA/k8s-test-infra/tests/e2e/go/framework/kube"
+	"github.com/NVIDIA/k8s-test-infra/tests/e2e/go/profile"
+)
+
+// Running nvidia-smi in a pod and asserting through Gomega. Every helper is a
+// Ginkgo helper so failures point at the calling spec line. ExecQuiet keeps the
+// ~90 KB document out of the Ginkgo log; only the decoded problems are
+// attached to a failure.
+
+// Inventory ports validate-nvidia-smi.sh through `kubectl exec`: nvidia-smi
```
roma-glushko: We need to make sure we generate comments that are useful for long-term code readers and not based on the current conversation. For example, I'm not sure I know what's `validate-nvidia-smi.sh` and why it's important to mention it here. 

I put an instruction about this to AGENTS.md but apparently it keeps popping up.

tests/e2e/go/assertions/nvidiasmi/diff.go
```diff
@@ -0,0 +1,138 @@
...
+// The checks over a decoded document. Each returns the problems it found rather
+// than failing, so one call reports every wrong reading instead of stopping at
+// the first, and the pure comparison unit-tests without a cluster. The
+// architecture-dependent threshold check is large enough to live in
+// temperature.go.
+
+// DiffInventory checks the document describes exactly wantGPUs devices, all
+// named wantProductName. product_name carries the profile's DisplayName
+// verbatim, so this is an equality check rather than a substring search.
+func DiffInventory(out, wantProductName string, wantGPUs int) []string {
```
roma-glushko: When I look at `Diff*()` methods I expect them to return a difference between two states. But the current implementation of diff methods looks like they are rather custom assertions than that. 

So maybe it makes sense to turn them into ExpectDeviceCount(), ExpectDeviceName(), etc. methods? What would take a snapshot and some additional params and then assert/require/gomega under the hood?

I feel like it might be cleaner responsibility split.

roma-glushko (dismissed):
Great changeset directionally 👍 

I would think a bit more about those Diff methods. Also, all NVIDIA SMI parsing could be as well moved to internal/* or even pkg/*.

tests/e2e/go/assertions/nvidiasmi/exec.go
giuliocalzo: Fixed in 030f59b5, and you were right to pull the thread — I found five more of the same kind and rewrote them all to state the invariant instead of the history: `snapshot.go` contrasted its readings with "the CSV cell" and "the substring scan it replaces", and two tests explained themselves by reference to the heuristic they superseded. None of that means anything to someone reading the file cold.

Two things for context on `validate-nvidia-smi.sh` specifically. It does exist, at `tests/e2e/validate-nvidia-smi.sh`, and `X ports validate-Y.sh` is the current style across the package — `ib.go`, `nvlink.go`, `iblinkinfo.go`, `ibping.go` and `pcisysfs.go` all carry it on `main`, from when the Go harness replaced the shell scripts. So the affordance was there to copy. But your point stands independently of that, and worse, the sentence was misleading here: this assertion no longer does what that script does, since the script scrapes the human table and this reads the XML. It now says what the check guarantees.

Happy to strip the script references out of the sibling files too if you want that consistent, though it touches assertions unrelated to this PR so I'd rather do it separately.

tests/e2e/go/assertions/nvidiasmi/diff.go
giuliocalzo: Agreed on the name, renamed in 030f59b5: `DiffInventory` → `InventoryProblems`, `DiffNoProcesses` → `PhantomProcessProblems`, `DiffJpgOfaUtilization` → `JpgOfaUtilizationProblems`, `DiffEncoderFBC` → `EncoderFBCProblems`, `DiffTemperature` → `TemperatureProblems`, and `diff.go` → `checks.go`. "Diff" promised a comparison of two states when what these do is take one document and answer "what is wrong with this".

I kept the signature returning findings rather than asserting inside, though, because that is what makes them usable in both modes:

- The JPEG/OFA and encoder/FBC checks are polled inside an `Eventually` (`scenario_runtime_control.go:544`, `:975`) while the mock converges on the runtime override. A `gomega.Expect` inside the check would fail the spec on the first poll instead of retrying. It could take a `Gomega` parameter to survive that, but then every caller has to pass one.
- Returning all the problems means a stubbed run reports every wrong reading in one pass rather than one per fix-and-rerun cycle, which is how #636 and #637 were actually diagnosed.

The asserting layer you are describing already exists, it just wasn't advertised by the naming: `exec.go` is exactly that — `Inventory`, `JpgOfaUtilization` and `TemperatureThresholds` take a pod, run nvidia-smi and assert through Gomega, and they are the only things carrying `//go:build e2e`. The split is checks are pure and unit-tested without a cluster, `exec.go` does the cluster and the failing. I've made that explicit in the `checks.go` header.

One inconsistency this creates, so you can tell me if you'd rather I went the other way: `Diff*`-returning-`[]string` is the existing convention in the parent package. `DiffGFDLabels` (#542) and `DiffFGOProfileConfigMap` have the identical contract, "reports every problem rather than stopping at the first" and all, which is what I was copying. Renaming only mine makes `nvidiasmi` the odd one out. I'd suggest renaming those two in a follow-up rather than growing this PR into them, but say the word if you'd prefer it all in one go — or if consistency with the siblings matters more than the name, I'll revert this commit.

tests/e2e/go/assertions/nvidiasmi/exec.go
roma-glushko: > Happy to strip the script references out of the sibling files too if you want that consistent, though it touches assertions unrelated to this PR so I'd rather do it separately.

All unneeded comments/information should be removed. Source code is for human's and they don't want to read tooo much.

tests/e2e/go/assertions/nvidiasmi/diff.go
roma-glushko: With this update I was just hoping we could avoid having this complicated string building/checks like

```
assert.Contains(t, strings.Join(problems, "; "), "2 <gpu> elements")
assert.Contains(t, strings.Join(problems, "; "), "pid 0")
...
```

I would rather check a field on some struct that has the information in the ready-to-check form and shape. Otherwise, we may argue this may be brittle.

giuliocalzo:
Both inline points are addressed in 030f59b5 and d8db5399.

On `internal/*` or `pkg/*`: I'd keep it in the e2e tree. The mock sits below nvidia-smi and never sees the document, so there's no non-test consumer and nothing in `pkg/` or `internal/` parses this XML today. The pure half (`schema.go`, `snapshot.go`, `checks.go`, `temperature.go`) is stdlib-only, so if a consumer does appear the move is mechanical and the split is already drawn along the build tag. Happy to reconsider if you have one in mind.

Rebased on `42e4dd0`. The force-push dismissed your approval, so it needs another look.

roma-glushko (approved):
Green light from my side ✅

PR #661, opened 2026-08-17
feat(mokka): add rack API

### LINKED ISSUE #669: MEP0001: Introduce SGPURack for rack-level state keeping
We have decided to introduce SGPURack in order to keep rack-level runtime stable state and some additional status information that might be useful for cluster admins.

tjrasche:
## What This PR Does

- Adds the cluster-scoped `SGPURack` API and status subresource.
- Represents rendered rack identity, pinned inventory/profile revisions, bounded logical Nodes, and exact Kubernetes Node name/UID assignments.
- Keeps reusable fabric and network capabilities in the referenced Rack Profile; the Rack stores only generated rack/GPU identities and bindings.
- Renames the unused `SGPUProfile` API to `SGPURackProfile` so the desired template and rendered Rack relationship is explicit.
- Adds Kubernetes list-map and validation semantics for deterministic server-side updates.
- Registers the types, generates deep-copy code, packages both CRDs in the existing Mokka CRD Helm chart, and updates MEP0001 terminology.
- Adds focused scheme, ownership, deep-copy, and generated-schema tests.

This PR intentionally contains no controller, allocator, Node projection, REST endpoint, agent, RBAC, or deployment implementation.

### Rack Profile API rename

The existing `v1alpha1` Profile API has no users yet, so this PR renames it before consumers depend on the old identity:

- Kind: `SGPUProfile` → `SGPURackProfile`
- Resource: `sgpuprofiles` → `sgpurackprofiles`
- Short name: `sprof` → `srprof`
- Go types, generated CRD filename, scheme registration, tests, and MEP terminology follow the new name.

The JSON relationship field remains `profileRef`; only the referenced API identity changes.

## Why

`SGPUInventory` describes the desired rack composition, but the control plane also needs a durable, rendered rack-level representation of assignments. Grouping exact Node bindings in Rack resources avoids a resource per Node and provides the persisted source that later controller and agent-delivery steps can consume. Static fabric and network capabilities stay normalized in the pinned Rack Profile rather than being copied into every Rack.

`SGPURackProfile` now names the reusable desired template, while `SGPURack` names one materialized instance. This is an incremental API-only step for MEP0001 (#536).

## Testing

- `go test -race ./...`
- `make lint-fix`
- `make gen-check`
- `make helm-crds-tests`
- `make build`
- `make vendor-check`
- `make modules-check`
- Repeated generation produced an identical tree.

## Checklist

- [x] Documentation updated
- [x] CHANGELOG.md updated (not applicable; these unused APIs are still `v1alpha1`)


internal/controlplane/api/v1alpha1/sgpurack_types.go
```diff
@@ -0,0 +1,247 @@
...
+	FabricUUID string `json:"fabricUUID"`
+
+	// CliqueID remains zero while each rack represents one fabric clique.
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=0
+	CliqueID int32 `json:"cliqueID"`
+}
+
+// SGPURackSlot is one durable allocation coordinate in a rack.
+type SGPURackSlot struct {
```
roma-glushko: Basically, SGPURackSlot represents the state of SGPU Node. We should probably name it correspondingly not to diverge from the current terminology. 

I know it intersects with K8s vanilla Node which is not a confidence as SGPU nodes are represented as regular K8s nodes in reality.

internal/controlplane/api/v1alpha1/sgpurack_types.go
```diff
@@ -0,0 +1,247 @@
...
+
+	// +kubebuilder:validation:Minimum=0
+	NUMANode int32 `json:"numaNode"`
+
+	// +kubebuilder:validation:Minimum=0
+	HostProcessorIndex int32 `json:"hostProcessorIndex"`
+}
+
+// SGPUGPUFabric is the rendered rack-local GPU fabric topology.
+type SGPUGPUFabric struct {
```
roma-glushko: Do we need this information?

internal/controlplane/api/v1alpha1/sgpurack_types.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+// SGPUGPUFabricSwitches contains rendered fabric-switch visibility.
+type SGPUGPUFabricSwitches struct {
+	// +kubebuilder:validation:Minimum=1
+	VisiblePerNode int32 `json:"visiblePerNode"`
+}
+
+// SGPUNetwork is the rendered rack-local network topology.
+type SGPUNetwork struct {
```
roma-glushko: Do we need this information?

roma-glushko (dismissed):
Looks great to me 👍

internal/controlplane/api/v1alpha1/sgpuprofile_types.go
```diff
@@ -32,7 +32,8 @@ type SGPUProfileList struct {
 
 // SGPUProfileSpec is one logical rack profile.
 type SGPUProfileSpec struct {
-	Rack SGPURack `json:"rack"`
+	// Rack is the logical rack shape described by this profile.
+	Rack SGPUProfileRack `json:"rack"`
```
roma-glushko: Should not it be SGPURackProfile? I think the whole SGPUProfile then should be renamed to SGPURackProfile

internal/controlplane/api/v1alpha1/sgpurack_types.go
tjrasche: Nope.

I even spent a lot of time arguing that we should not duplicate derived state. Mea culpa.

internal/controlplane/api/v1alpha1/sgpurack_types.go
tjrasche: Nope.

I even spent a lot of time arguing that we should not duplicate derived state. Mea culpa.

tjrasche: Removed. The current network fields were a direct copy of `SGPUProfile.spec.node.topology.network`. Rack-specific network data can be added later if materialization produces identities or mappings that cannot be derived from the Profile.

internal/controlplane/api/v1alpha1/sgpurack_types.go
tjrasche: Updated. The API now uses `SGPURackNode` and `.spec.nodes[]`, with `status.assignedNodes`. Each entry represents a stable logical sGPU Node and carries an optional exact Kubernetes Node name/UID binding.

internal/controlplane/api/v1alpha1/sgpurack_types.go
tjrasche: Removed. Static fabric capabilities remain in the pinned `SGPUProfile`; `SGPURack` retains only the generated rack-specific `fabricUUID` and `cliqueID` plus its logical Node/GPU identities.

internal/controlplane/api/v1alpha1/sgpuprofile_types.go
tjrasche: Agreed and updated. The root API is now `SGPURackProfile`, including `SGPURackProfileList`, the `sgpurackprofiles` resource, `srprof` short name, Go types/file, generated CRD filename, scheme registration, tests, and MEP terminology. The JSON relationship remains `profileRef`.

tjrasche: yeah naming the other CR SGPURackProfile seems like the most sensible choice to me.

Done :)

roma-glushko (approved):
Great thank you addressing my comments 👍

PR #662, opened 2026-08-18
fix(mocknvml): report bare-metal virtualization mode

### LINKED ISSUE #640: [Bug]: nvmlDeviceGetVirtualizationMode is unimplemented, so nvidia-smi reports N/A where bare metal reports None
## Component / version

## What happened

`nvmlDeviceGetVirtualizationMode` is a generated stub, so `nvidia-smi -q` reports `Virtualization Mode : N/A`. A real GPU on bare metal reports `None`, which is a meaningful answer rather than an absence — the mock is claiming it cannot tell whether it is virtualized.

This is the most frequently called unimplemented function in a single `nvidia-smi -q` run: 25 calls across 8 GPUs, roughly three per device, more than any other stub.

### Example output

`profiles/a100.yaml`:

```
    GPU Virtualization Mode
        Virtualization Mode               : N/A
        Host VGPU Mode                    : N/A
        vGPU Heterogeneous Mode           : N/A
```

A real bare-metal A100 reports:

```
    GPU Virtualization Mode
        Virtualization Mode               : None
        Host VGPU Mode                    : N/A
```

Only the first row is wrong. `Host VGPU Mode` and `vGPU Heterogeneous Mode` are correctly `N/A`, since those require vGPU.

`MOCK_NVML_DEBUG=1`, sorted by call count, shows it at the top:

```
  25 nvmlDeviceGetVirtualizationMode called (NOT IMPLEMENTED)
  16 nvmlDeviceGetDriverModel_v2 called (NOT IMPLEMENTED)
   8 nvmlDeviceWorkloadPowerProfileGetCurrentProfiles called (NOT IMPLEMENTED)
   ...
```

## Expected behaviour

- `nvmlDeviceGetVirtualizationMode` returns `NVML_GPU_VIRTUALIZATION_MODE_NONE` for the bare-metal profiles the mock ships, and `nvidia-smi -q` reports `Virtualization Mode : None`.
- `Host VGPU Mode` and `vGPU Heterogeneous Mode` remain `N/A`.
- `nvidia-smi pmon`, `topo -m`, and the default table are unaffected.

## Known risk — read before implementing

An implementation of this function was attempted and reverted during the investigation behind PR #630. Returning a virtualization mode changed the code path `nvidia-smi pmon` takes and produced a deeper crash than the one being fixed at the time. `pmon` is unusually sensitive here: it reaches the mock through the reverse-engineered internal export table, and on `main` it exits `0` with no output, while small changes in what neighbouring calls return can turn that into exit `255` or a `SIGSEGV` (exit `139`).

Any fix must therefore verify `pmon` explicitly rather than assuming it is unaffected, which is why this is filed separately from the straightforward stub-wiring gaps rather than bundled with them.

## Steps to reproduce

```bash
  nvml-mock:latest nvidia-smi -q | grep -A 3 "GPU Virtualization Mode"

# call frequency
```

## Environment

```yaml
```

## Acceptance criteria

- [ ] `nvmlDeviceGetVirtualizationMode` is hand-written in the bridge, returns `NONE` for bare-metal profiles, and is removed from `stubs_generated.go`.
- [ ] `nvidia-smi -q` reports `Virtualization Mode : None`, with `Host VGPU Mode` and `vGPU Heterogeneous Mode` still `N/A`.
- [ ] `nvidia-smi pmon -c 1` behaviour is compared against `main` before and after the change and is no worse: it must not segfault, and its exit code must not regress.
- [ ] `nvidia-smi topo -m` still succeeds, guarding the neighbouring internal-export-table path.
- [ ] **Dedicated e2e test.** A new Ginkgo spec under `tests/e2e/go/` with `Label("nvidia-smi")` that asserts `nvidia-smi -q` reports `Virtualization Mode : None`, and — in the same spec or an adjacent one — that `nvidia-smi pmon -c 1` does not exit with a signal and that `nvidia-smi topo -m` exits `0`. The `pmon` and `topo` assertions are the point: they turn the regression that was hit manually during #630 into something CI catches. The test must fail against the current `main`.

## Out of scope

Actual vGPU support — `Host VGPU Mode`, `vGPU Heterogeneous Mode`, and GRID licensable features stay `N/A`, matching bare-metal hardware. Also out of scope: the roughly 60 other `-q` fields that are faithfully `N/A`.

giuliocalzo:
## What This PR Does

`nvmlDeviceGetVirtualizationMode` was a generated stub, so `nvidia-smi -q` printed `Virtualization Mode : N/A`. It is now hand-written and resolves the `virtualization.mode` every profile already carried, so `-q` reports `None`:

```
    GPU Virtualization Mode
        Virtualization Mode               : None      # was N/A
        Host VGPU Mode                    : N/A
        vGPU Heterogeneous Mode           : N/A
```

vGPU stays out of scope. `Host VGPU Mode` and `vGPU Heterogeneous Mode` keep reading `N/A` the way bare-metal hardware reports them, and `nvmlDeviceGetVgpuHeterogeneousMode` remains a stub.

A new Ginkgo spec under `Label("nvidia-smi")` asserts the reading from `-q -x`, and — the point of the second commit — pins `pmon -c 1` and `topo -m` alongside it.

## Why

Closes #640.

`N/A` claims the driver cannot tell whether the GPU is virtualized. No real device answers that: bare metal reports `None`. It was also the most frequently called unimplemented export in a single `-q` run, at 25 calls across 8 GPUs.

**The reason this was filed separately from the other stub-wiring gaps:** an implementation was written and reverted during the investigation behind #630, because returning a mode moved `nvidia-smi pmon` onto a different code path and crashed it. `pmon` reaches the mock through the reverse-engineered internal export table, which is unusually sensitive to what neighbouring calls return, so that regression was only ever caught by hand. This PR turns it into something CI catches.

### Verification against `main`

Baseline captured from the pre-fix image before anything was touched, then compared:

| | `main` | this PR |
| --- | --- | --- |
| `-q` virtualization mode | `N/A` | `None` |
| `-q` host vGPU / heterogeneous | `N/A` | `N/A` |
| `pmon -c 1` | 255, `Not supported on the device(s)` | 255, identical |
| `topo -m` | 0 | 0 |
| default table, `-q`, `-q -x` | 0 | 0 |

Checked across the `a100`, `gb200`, `t4` and `h100` profiles. No segfault, no exit-code regression. A full diff of `nvidia-smi -q` between the two images moves only three things: the timestamp, the virtualization mode itself, and the deliberately dynamic utilization percentages.

Note the issue text says `pmon` exits 0 with no output on `main`. That is no longer accurate — #630 changed it to refuse gracefully with 255, as the current `[Unreleased]` changelog records — so the spec accepts 0 or 255 and fails on a signal death instead of pinning today's number as correct.

The document check is proved against the committed `-q -x` fixtures, which are verbatim captures of `main`, so it fails against `main` as the issue requires. The e2e spec was run against a real Kind cluster: 5 specs passed, and the log confirms `kubectl exec` surfaces pmon's 255 the way the assertion reads it.

## Checklist

- [x] Tests pass (`make test`, plus `make e2e E2E_PROFILES=a100 E2E_GINKGO_FLAGS='--label-filter="nvidia-smi && !nri"'`)
- [x] Linter passes (`make lint-fix`) — `golangci-lint` and `gen-check` are clean. `govulncheck` fails locally on seven Go standard-library CVEs fixed in go1.26.6; my toolchain is 1.26.5 and the repo pins 1.26.6 since #654, so this is a local-toolchain artefact rather than anything in the diff.
- [x] New code has SPDX license headers — the new e2e code follows the SPDX form used in `tests/e2e/go`; the new engine test uses the full Apache header that every other file in `pkg/gpu/mocknvml/engine` carries.
- [ ] Documentation updated (if applicable) — no doc change needed; the profile field was already documented.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   reading — the same class of bug as the phantom processes above. All three
   maxima now agree in `nvidia-smi -q`, `-q -x` (`<max_host_link_gen>`) and
   `--query-gpu=pcie.link.gen.hostmax`. (#638)
+- mocknvml: `nvidia-smi -q` reports `Virtualization Mode : None` instead of
+  `N/A`. `nvmlDeviceGetVirtualizationMode` was a generated stub — the most
+  frequently called one in a single `-q` run — so the mock claimed it could not
+  tell whether it was virtualized, where bare-metal hardware always answers
+  `None`. The export is now hand-written and resolves the `virtualization.mode`
+  the profiles already carried (`none`, `passthrough`, `vgpu`; anything
+  unrecognised reads as bare metal). vGPU stays out of scope: `Host VGPU Mode`
+  and `vGPU Heterogeneous Mode` still read `N/A`, matching bare metal. An
+  earlier attempt at this fix was reverted during PR #630 because it moved
+  `nvidia-smi pmon` onto a different code path and segfaulted it, so an e2e spec
+  now pins `pmon -c 1` to a graceful refusal or success and requires `topo -m`
+  to succeed, guarding the neighbouring internal-export-table path. (#640)
 - Wire eight NVML device exports that already had engine implementations but
   were still generated stubs (`GetEncoderStats`, `GetFBCStats`,
   `GetAccountingBufferSize`, `GetEncoderCapacity`, `GetEncoderSessions`,
--- a/pkg/gpu/mocknvml/bridge/device.go
+++ b/pkg/gpu/mocknvml/bridge/device.go
@@ -54,6 +54,7 @@
 // - nvmlDeviceGetCurrentClocksThrottleReasons
 // - nvmlDeviceGetUtilizationRates
 // - nvmlDeviceGetComputeMode
+// - nvmlDeviceGetVirtualizationMode
 // - nvmlDeviceGetEccMode
 // - nvmlDeviceGetDisplayMode
 // - nvmlDeviceGetAccountingMode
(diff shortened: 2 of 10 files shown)


roma-glushko (approved):
LGTM ✅

Issue #597, opened 2026-08-03
Observability for scale runs — Prometheus and Grafana across the simulated cluster


ArangoGutierrez:
## Problem

A 5,000-node run produces a verdict nobody can explain without telemetry. The goal of the run is a component-by-component *what breaks at scale* report, and today there is no way to collect the evidence for one.

## Why now

Carlos Arango Gutierrez, 2026-07-30: *"to do this, we are adding more observability and monitoring to the project. And we are working towards like Prometheus, Grafana, the entire suite. And we want to be able to get metrics when you are running at scale so we can check what is breaking and why is breaking."*

In the 2026-07-20 sprint planning this was named a prerequisite for the scale work rather than a parallel nice-to-have: *"What he has been doing on observability and monitoring is very important, mostly for what we are planning to do with Sai, because once we go thousands of nodes, we are going to need a lot of data coming out of that experiment."*

It is also a second-party ask. In the 2026-07-20 walkthrough for the Network Operator team, Giulio Calzolari noted Prometheus/Grafana end-to-end integration was among the hottest requests they had received.

## Scope

- Prometheus scrape targets across the Mokka components, with a Grafana dashboard shipped in-tree
- Metrics that answer *which component broke and when* during a scale run, not only host vitals
- Retain the series as a CI artifact so a run stays diagnosable afterwards
- Cross-check against the DCGM exporter surface already integrated

## Constraint

Whatever this adds counts against the per-node footprint budget. Size it together with that work.

## Related

#252 (resource monitoring dashboard, the CI-artifact half of this).
---
*Raised by the 2026-08-03 roadmap review of team meeting transcripts, Slack and mail. No tracking issue existed for this.*

roma-glushko:
> Carlos Arango Gutierrez, 2026-07-30: "to do this, we are adding more observability and monitoring to the project. And we are working towards like Prometheus, Grafana, the entire suite. And we want to be able to get metrics when you are running at scale so we can check what is breaking and why is breaking."

How is this connected to Mokka though? In the upcoming 5k test, we are going to test not Mokka but our consumers. So we will take a Grafana stack, connect our consumer to the stack so they can emit o11y signals. Then, we will use that information for analysis. I'm not sure I see what Mokka is supposed to be doing here. That's outside of Mokka's responsibility. 

@avasilevskii do you have a different opinion here?

avasilevskii:
[...]

PR #663, opened 2026-08-18
feat(local): --observability Tilt consumer with Prometheus and Grafana

giuliocalzo:
Refs #597.

Prometheus scrapes the real, unmodified NVIDIA `dcgm-exporter` reading Mokka's mock `libnvidia-ml.so`, and Grafana renders it — on a cluster with no GPUs. Two manual triggers then inject a GPU fault and **fail if it never reaches Prometheus**, so the scrape path is asserted rather than eyeballed. That gives a scale run a "which component broke, and when" view.

```bash
make cluster-create
tilt up -- --observability     # Grafana: http://localhost:3000/d/mokka-gpu (admin/mokka)
```

## What this adds

- `local/observability/` — a Tilt consumer in the shape of `local/compute-domain/`: kube-prometheus-stack (trimmed: Alertmanager off, default rules off, 2h retention), the Mokka dashboard in-tree, and two `auto_init=False` triggers `inject-thermal` / `inject-xid`.
- `--observability` on the root Tiltfile. It implies `--gpu-operator`, since `dcgm-exporter` is one of the Operator's operands.
- `gpu_operator.tiltfile` grows `extra_values` / `extra_resource_deps` (following `dra.tiltfile`'s precedent) so the consumer can layer an overlay re-enabling `dcgmExporter`. Default behavior unchanged.
- Replaces the standalone `docs/demo/observability/run.sh` from the earlier revision of this PR. Tilt composes with the GPU Operator instead of replacing it, and the script duplicated per-worker toolkit provisioning and CDI wiring that the shared kind-node image already bakes in.

## Design notes

**Install order is load-bearing.** kube-prometheus-stack goes in *before* the GPU Operator: it ships the `ServiceMonitor` CRD that the Operator's chart needs. Expressed via `resource_deps`, not call order, so reordering the root Tiltfile can't silently break it.

**The `release` label trap.** kube-prometheus-stack defaults to `serviceMonitorSelectorNilUsesHelmValues: true`, so Prometheus only discovers ServiceMonitors labelled `release: kube-prometheus-stack`. A mismatch is silent — every panel is empty with no error anywhere. Pinned in the overlay with a comment naming the coupling, and the verification below asserts the target is actually `up`.

**Assertions can't pass on residue.** The thermal run resets every GPU on the node and waits for a reading that *differs* before asserting exact equality, so a re-run can't pass on the previous pin. The Xid run rotates between codes 79 and 48 and injects whichever is provably absent, because DCGM latches the last Xid per device and never retracts it — a fixed code would be satisfied by the previous run's residue, and would pass even with dcgm-exporter dead.

**The temperature clamp.** The mock clamps an injected temperature to the profile's `shutdown_threshold_c`, and the assertion checks exact equality, so too high a value burns the poll budget and reports a timeout that never mentions clamping. Default is 90 °C, the one band that works for every `--gpu-profile`: above the simulator's 73 °C ceiling (so a sibling can't wander onto it and fake a scope leak) and at or below the lowest threshold of any profile (92 °C on a100/h100). Guarded up front with messages that name both failure modes, plus a clamp hint on timeout.

**Grafana access** is a `local_resource` with `serve_cmd` port-forwarding the Grafana Service by name, gated on a readiness probe against Grafana's `/api/health`. Not `helm_resource(port_forwards=...)`: that extension deploys a whole release as *one* Tilt resource owning every pod, so Tilt would forward to whichever of Prometheus, node-exporter, kube-state-metrics or Grafana it picked.

## Verification

`tilt ci -- --observability` → `SUCCESS. All workloads are healthy.` in 51s (warm cluster).

dcgm-exporter discovered through the ServiceMonitor and scraped on both workers:

```
pool=serviceMonitor/gpu-operator/nvidia-dcgm-exporter/0 job=nvidia-dcgm-exporter health=up node=mokka-worker
pool=serviceMonitor/gpu-operator/nvidia-dcgm-exporter/0 job=nvidia-dcgm-exporter health=up node=mokka-worker2
count(DCGM_FI_DEV_GPU_TEMP) = 16
```

Grafana imported the dashboard, via Grafana's own API rather than the ConfigMap existing:

```
uid=mokka-gpu title=Mokka — Mock GPU Fleet type=dash-db
```

Both triggers run twice, each showing the non-vacuity property holding:

```
inject-thermal #1: gpu 0 on mokka-worker stepped 63C -> 90C   (siblings 59-65C)
inject-thermal #2: gpu 0 on mokka-worker stepped 70C -> 90C   (different baseline: the reset works)
inject-xid #1: series absent -> 79   (did not exist before, so the whole path was travelled)
inject-xid #2: 79 -> 48              (rotated: a fresh delivery, not a latched re-read)
```

No regression with the flag off: `tilt alpha tiltfile-result` yields the identical resource graph as before (`nvml-mock`, `gpu-validator`), and the GPU Operator gets exactly its original single `--values` file; the overlay is layered second only when `--observability` is on.

Static: `shellcheck` clean, `bash -n` clean, `jq empty` on the dashboard, SPDX headers on all new YAML/shell.

## Not verified

- **A cold-cluster run.** The ServiceMonitor CRD survives a Helm uninstall, so the CRD-ordering dependency wasn't exercised on this pass; it's correct by construction via `resource_deps`.
- **Dashboard hot-reload.** The ConfigMap is built with `read_file` + `encode_yaml` so Tilt watches the JSON and should re-apply on edit, but I didn't confirm the sidecar re-imports it.


giuliocalzo:
view from grafana


image

local/observability/scenarios/inject-thermal.sh
```diff
@@ -0,0 +1,188 @@
...
+  || fail "no sibling GPU series on ${target_node} to compare against, so the scope check would be vacuous"
+hot_siblings=$(jq -r --argjson hot "${HOT_TEMP_C}" \
+  '[.[] | select((.value[1] | tonumber) == $hot) | .metric.gpu] | sort | join(",")' <<<"${siblings}")
+[[ -z "${hot_siblings}" ]] \
+  || fail "gpu(s) ${hot_siblings} on ${target_node} also read ${HOT_TEMP_C}C; the override is not scoped to gpu ${TARGET_GPU}"
+printf 'OK: the other %s GPUs on %s kept simulator-driven temperatures (%s)\n' \
+  "${sibling_count}" "${target_node}" \
+  "$(jq -r '[.[] | "gpu" + .metric.gpu + "=" + .value[1] + "C"] | sort | join(" ")' <<<"${siblings}")"
+
+printf '\n==> inject-thermal passed: gpu %s on %s stepped %sC -> %sC and Prometheus recorded it.\n' \
```
ArangoGutierrez, line 185: The port drops the pod-churn check run.sh carried: it fingerprinted the nvml-mock and dcgm-exporter pods (name, startTime, per-container restartCount) before and after injection and failed if either set changed, so "no pod was recycled, so the series has no gap" was asserted rather than stated. Both script headers and the README still make that claim, but nothing verifies it now. If the Operator replaces dcgm-exporter mid-run, the pinned value still reaches Prometheus on the new pod and the scenario prints passed over exactly the gap it advertises against.

ArangoGutierrez (approved):
Nothing here blocks merge.

The part I like here is that both scenarios are real assertions rather than eyeball demos, and the non-vacuity arguments hold up: the thermal reset-then-wait-for-a-differing-reading genuinely stops a re-run passing on the previous pin, and rotating Xid codes is the right answer to DCGM latching the last one. I checked the HOT_TEMP_C guard band against the source rather than the comment, and it is correct: 73 is exactly base 55 + ramp 15 + variance 3 from the dynamic-metrics simulator, 92 is the lowest shutdown_threshold_c (a100/h100), and no profile overrides the temperature defaults. The err_code ambiguity hard-error and the local-dev Grafana password both read as deliberate calls, so I left them alone. What follows is one assertion I think went missing in the port, plus some comment drift.

1. This comment describes two split-out port constants, but it sits above `_GRAFANA_URL` (a dashboard URL) and the port-forward hardcodes `3000:80` inline in `_serve_grafana`. Reads like leftover from a refactor that removed the constants. (local/observability/observability.tiltfile:55)
2. The "Ported from docs/demo/observability/run.sh" header points at a path this branch deletes, so it resolves to nothing for anyone reading after the merge. Citing #597 or the commit would survive. (local/observability/scenarios/inject-thermal.sh:18; also local/observability/scenarios/inject-xid.sh:21)
3. `_install_dashboard()` hardcodes `grafana_dashboard: "1"` rather than reading these values, so "observability.tiltfile builds the ConfigMap's label from them" overstates it. The sync is manual, which is what the tiltfile's own comment already says correctly. (The `enabled`/`label`/`labelValue`/`searchNamespace` values themselves do match the 88.3.0 chart defaults.) (local/observability/kube-prometheus-stack.values.yaml:37)

local/observability/gpu-operator.values.yaml
```diff
@@ -0,0 +1,39 @@
...
+#
+# Standalone DCGM (nv-hostengine) stays disabled — dcgm-exporter embeds its own
+# hostengine and nothing here polls one directly.
+
+dcgmExporter:
+  enabled: true
+  env:
+    # The exporter loads the mock libnvidia-ml.so from the driver root the
+    # nvml-mock DaemonSet stages on each node.
+    - name: NVIDIA_DRIVER_ROOT
```
roma-glushko, line 19: We need to be careful with what we put into these YAML overrides because that's our end user experience. To me, env vars like this leaks what NRI was promised to help us with - hiding how we plug in mokka into consumer workloads. 

Can we make NRI do this injection for us instead of asking our users to apply overrides themselves?

local/observability/scenarios/inject-thermal.sh
roma-glushko, line 1: These two sh files make me curious what's the best way to organize that.
Right now they look like one off tests that should not be even committed.

@giuliocalzo Let's think together why we are doing that & what we might be missing to do it in cleaner way if at all.

local/observability/observability.tiltfile
```diff
@@ -0,0 +1,190 @@
...
+    k8s_resource(port_forwards=...) on a resource which — because
+    helm_resource deploys the whole release as ONE Tilt resource — owns every
+    pod in the stack, so Tilt forwards to whichever of Prometheus,
+    node-exporter, kube-state-metrics or Grafana it picked. Forwarding the
+    Grafana Service by name is deterministic.
+
+    The shared Kind config carries no extraPortMappings, and adding them for
+    one optional consumer would change the cluster shape for every other one.
+    """
+    local_resource(
```
roma-glushko, line 147: Tilt allows to do portforwarding natively not need to hack it:

https://docs.tilt.dev/accessing_resource_endpoints.html#creating-a-kubectl-port-forward-tunnel

local/observability/observability.tiltfile
```diff
@@ -0,0 +1,190 @@
...
+
+
+def _install_dashboard():
+    """Provision dashboards/mokka-gpu.json for the Grafana sidecar.
+
+    Built from read_file + encode_yaml rather than shelled out to `kubectl
+    create configmap` so Tilt watches the JSON: editing a panel and saving
+    re-applies the ConfigMap and the sidecar re-imports it, with no re-install.
+    """
+    k8s_yaml(encode_yaml({
```
roma-glushko, line 109: This is fine to create a CM/Secret from tilt but I would even consider having a custom, slim helm chart that depends on grafana chart and adds this configmap + required values overrides to it. 

This might be even a cleaner solution. Worth trying out.

Then tilt would just deploy that helm chart. The only inconvenience could be the need to do helm build before tilt up (but tilt can handle that as well).

roma-glushko (approved):
Looks good to me 🥇

local/observability/gpu-operator.values.yaml
roma-glushko: @ArangoGutierrez what's your opinion here?

Issue #602, opened 2026-08-05
[Bug]: Flaky Test scenario_gpu_operator_test.go:109 fails with "[FAIL] nvml-mock GPU Operator profile h100 [It] surfaces an injected Xid through dcgm-exporter [gpu-operator, h100, dcgm, xid]"


roma-glushko:
### Component / version

main

### What happened

```shell
scenario_gpu_operator_test.go:109 test fails random CI pipelines with:

[FAIL] nvml-mock GPU Operator profile h100 [It] surfaces an injected Xid through dcgm-exporter [gpu-operator, h100, dcgm, xid]
```

Example: https://github.com/NVIDIA/k8s-test-infra/actions/runs/30935047138/job/92095889751

### Expected behaviour

The test runs deterministically.

### Steps to reproduce

```bash
make e2e E2E_GINKGO_FLAGS='--label-filter=gpu-operator'
```

### Environment

```yaml
Github Action (E2E tests), but likely it's reproducible locally too
```

### Logs / artefacts

[flaky-gpu-operator-test.output.txt](https://github.com/user-attachments/files/30745402/flaky-gpu-operator-test.output.txt)

### Pre-submission checks

- [x] I searched existing issues and Discussions and this is not a duplicate.
- [x] This is a defect, not a usage question (questions belong in Discussions).

PR #674, opened 2026-08-18
feat(nvml-mock): serve the rendered PCI sysfs tree and a DMI identity to Go consumers

### LINKED ISSUE #673: [Bug]: rendered PCI sysfs tree never reaches Go consumers, so GPU Operator labels gpu.mode and gpu.machine unknown
## What happened

On a mock node the rendered PCI sysfs tree is staged only inside the nvml-mock pod's own overlay (`/var/lib/nvml-mock/sys`). Consumers reach it through `MOCK_PCI_ROOT` and the `libpcimocksys.so` LD_PRELOAD redirector, which works for C tooling (`lspci`) but not for Go binaries: Go's `os` package issues `openat` directly, so the shim never sees the open and the process reads the node's real `/sys`, where the mock GPUs do not exist.

The visible symptom is in GPU Feature Discovery, which resolves each GPU's BDF from NVML and then reads its PCI class from sysfs:

```
Failed to create GPU mode labeler: failed to get device classes: unable to read PCI device vendor id
for 0000:0a:00.0: open /sys/bus/pci/devices/0000:0a:00.0/vendor: no such file or directory
```

The node is then labelled:

```
nvidia.com/gpu.mode=unknown
nvidia.com/gpu.machine=unknown
```

`gpu.machine` has a second, independent cause: nothing mocks the node's DMI identity, so GFD's `--machine-type-file` (`/sys/class/dmi/id/product_name`) is missing or belongs to the host.

The same gap keeps `dra.k8s.io/pcieRoot` absent from the DRA driver's ResourceSlices — the reason the profiles' `pcie_topology:` comments still say the driver "does NOT" resolve the root complex (#265).

## Expected

A GPU Operator or DRA consumer on a mock node resolves PCI attributes for the mock GPUs, labelling `nvidia.com/gpu.mode=compute` (or `graphics`) and a machine type from the profile, with no per-consumer configuration.

## Reproduction

Bring up `kind` with the nvml-mock chart and the GPU Operator (`gpu-operator` scenario), then:

```
kubectl logs -n gpu-operator -l app=gpu-feature-discovery | grep -i "gpu mode labeler"
kubectl get node <node> -o json | jq '.metadata.labels | with_entries(select(.key|startswith("nvidia.com/gpu.m")))'
```

## Investigation notes

Three constraints shape any fix; all were verified on `kind` (containerd 2.x, `kindest/node`):

1. **The tree must appear at the kernel paths.** `nvpci` (via go-nvlib) hard-codes `/sys/bus/pci/devices`, and `k8s.io/dynamic-resource-allocation/deviceattribute` requires the entry to be a symlink whose target starts with `devices/pci` — so the canonical layout the renderer already writes has to be preserved as-is.
2. **Two mounts are needed, not one.** `sys/bus/pci/devices` holds relative symlinks into `../../../devices/pciDDDD:BB`, so mounting it alone yields dangling entries: they list, but every attribute read returns ENOENT. A probe pod with both `sys/bus/pci/devices` and `sys/devices` mounted reads `0x10de` from `/sys/bus/pci/devices/0000:0a:00.0/vendor`.
3. **Per-root-complex mounts are impossible.** Mounting only the root complexes a profile declares would leave the rest of `/sys/devices` intact, but the runtime has to create the mountpoint and sysfs is read-only in the container:

```
error mounting "/var/lib/nvml-mock/sys/devices/pci0000:00" to rootfs at "/sys/devices/pci0000:00":
make mountpoint "/sys/devices/pci0000:00": ... read-only file system
```

Profiles routinely declare root complexes the host does not have (the `gb200` profile renders `pci0000:00`, `:40`, `:80`, `:c0` on a node whose `/sys/devices` has no `pci*` at all), so this path fails container creation rather than degrading.

Mounting `sys/devices` therefore hides the host's other device classes (CPU topology among them) from any container that is served the mock tree. That is the accepted trade-off for consumers that resolve GPUs through sysfs.

For DMI, `/sys/class/dmi` does not exist at all on some hosts (Docker Desktop's linuxkit VM), so it cannot be a mount target either. `/sys/class/dmi/id` is however only a symlink to `/sys/devices/virtual/dmi/id`, which is inside the subtree above — rendering the product name there makes it resolvable through both paths wherever DMI exists, and reachable by explicit `--machine-type-file` / `GFD_MACHINE_TYPE_FILE` everywhere else.

## Delivery

Both of the mock's injection channels need the mounts, since either may serve a given container:

- the CDI spec `setup.sh` generates (`/var/run/cdi/nvidia.yaml`) — the channel that reaches GPU Operator operands today, which is how GFD already receives the mock `libnvidia-ml.so.1`;
- the NRI plugin's container adjustment, for containers it serves.

Related: #671 (which fixed the malformed `busId` that made this failure read `:0a:00.0` instead of a well-formed BDF), #265.

giuliocalzo:
## Summary

The mock renders a PCI sysfs tree, but only libc consumers could read it: `MOCK_PCI_ROOT` works through the `libpcimocksys.so` `LD_PRELOAD` shim, and Go's `os` package issues `openat` directly, so the shim never sees the open and the process reads the node's real `/sys`. GPU Feature Discovery and the NVIDIA DRA driver are both Go. GFD therefore resolved each mock GPU's BDF from NVML, failed to read its `vendor`/`class`, and labelled the node `nvidia.com/gpu.mode=unknown`.

* `sys/devices` and `sys/bus/pci/devices` are bind-mounted read-only onto the kernel paths, in both channels that serve containers: the CDI spec `setup.sh` generates (which is how GFD already receives the mock `libnvidia-ml.so.1`) and the NRI plugin's adjustment. They go together — the PCI entries are relative symlinks into `../../../devices/pciDDDD:BB`, so mounting one alone yields entries that list but whose every attribute read returns `ENOENT`, which looks identical to no mount at all.
* The renderer mirrors the node's own `product_name` and `product_uuid` into `sys/devices/virtual/dmi/id`. This is mount-target compatibility, not a machine-type mock: shadowing `/sys/devices` replaces the directory `/sys/class/dmi/id` resolves into, and kind's `mount-product-files.sh` `createContainer` hook bind-mounts the node's product files there for *every* container. `mount(8)` cannot create a target on a read-only sysfs, so without the mirror every served pod died with `mount: .../sys/class/dmi/id/product_uuid: mount point does not exist` — on Linux only, since Docker Desktop's linuxkit VM exposes no DMI and kind's hook guards on the host kernel having it.
* The PCI render moves ahead of CDI spec generation, because a bind mount whose source is missing fails container creation for the whole pod.

There is no opt-out. A node running nvml-mock is simulating GPU hardware, and serving the tree is what makes that simulation visible to sysfs consumers; a switch that turns it half off only adds a way to end up with `gpu.mode=unknown` for a reason nothing reports. Workloads that need the host's real device tree on an NRI-injected node belong in `nri.excludedNamespaces`.

One read-only-sysfs constraint shapes the rest of the design, and it is worth stating because it rules out the obvious alternative. `/sys/devices` has to be mounted whole: narrowing it to the profile's root complexes needs a mountpoint the runtime cannot create, and it fails hard rather than degrading —

```
error mounting "/var/lib/nvml-mock/sys/devices/pci0000:00" to rootfs at "/sys/devices/pci0000:00":
make mountpoint "/sys/devices/pci0000:00": ... read-only file system
```

which profiles routinely trigger (`gb200` renders four root complexes on a node whose `/sys/devices` has none). That mount hides the host's other device classes, CPU topology among them, from served containers.

Fixes #673.

**Not in scope:** `nvidia.com/gpu.machine`. Mocking the machine type turned out to have a cause of its own — kind serves its own `product_name` at GFD's default machine-type file in every container, so no value the renderer writes there survives — and it is tracked separately in #681.

## Test plan

* `TestRender_MirrorsKernelDMI`, `TestRender_StandsInForUnreadableDMI`, `TestRender_NoDMIWithoutKernelDMI` — the mirrored attributes, the empty stand-in for `product_uuid` (mode `0400`, unreadable to a non-root renderer but still needed as a mount target), and no DMI directory at all on a kernel that exposes none.
* `TestAdjustMountsPCISysfsWhenStaged` / `...SkipsPCISysfsMountsWhenNotStaged` / `...SkipsPCIDevicesMountWithoutSysDevices` — the pair is emitted together, and skipped fail-open on a node the DaemonSet has not staged yet (a bind mount with a missing source would fail container creation for the whole pod).
* e2e: the gpu-operator scenario reads the tree from inside the GFD container at the kernel paths — device count equal to the profile's GPUs, and a `vendor` read that only resolves through the symlink into the mounted `/sys/devices`.
* `make test` — 33 packages pass. `make helm-tests` — 169 tests, 15 snapshots. `make lint-fix` — no violations (`govulncheck` reports pre-existing Go stdlib advisories from the local toolchain, unrelated).
* Live on a kind cluster (gb200 profile, GPU Operator in CDI mode, rebased branch): every operator pod `Running` with 0 restarts, `nvidia.com/gpu.mode=compute` where it previously read `unknown`, `gpu.product=NVIDIA-GB200` with 8 allocatable GPUs. Inside the `gpu-feature-discovery` container, `/sys/bus/pci/devices` lists 8 entries, the first symlinking to `../../../devices/pci0000:00/0000:0a:00.0`, which reads `vendor 0x10de`, `class 0x030200`, `numa_node 0` — the class being what GFD turns into `mode=compute`, so the label is derived from the mounted tree rather than coincidence. That host is linuxkit, so the mirror correctly writes nothing and `gpu.machine` stays `unknown` (#681); the Linux legs in CI cover the kind-hook path that the mirror exists for.


pkg/nri/nvmlmock/adjust.go
```diff
@@ -207,6 +221,7 @@ func Adjust(cfg Config, container Container) (Adjustment, bool, error) {
 		},
 		Env: buildEnv(cfg, container.Env, topologyInjectable(cfg)),
 	}
+	adjustment.Mounts = append(adjustment.Mounts, pciSysfsMounts(cfg)...)
```
ArangoGutierrez, line 221: This append lands before the DeviceAnnotation branch, so every container outside the release namespace and kube-system gets /sys/devices replaced by the mock tree, including containers that never opted into GPU injection. The CDI path is opt-in by construction (a container only gets the edits if it requests the device), so the same mount arrives with very different blast radii through the two channels. The rendered tree holds only sys/devices/<rootComplex> and sys/devices/virtual/dmi/id, so an unrelated pod loses /sys/devices/system/cpu and also its own namespaced sysfs: /sys/class/net/<iface> is a symlink into devices/virtual/net and dangles after the mount. I read 853cea23 and I take the argument that half-simulating a node is its own trap, but the knob it removed conceded this case in its own comment, and excludedNamespaces replaces an opt-out with a manual allowlist you have to know to populate. Gating on the same cfg.DeviceAnnotation that already gates device injection would keep the simulation whole for containers that asked for GPUs without touching the rest of the node. Was that considered and rejected?

pkg/nri/nvmlmock/adjust.go
```diff
@@ -338,6 +353,59 @@ func topologyInjectable(cfg Config) bool {
...
+// whole pod. Silence rather than a warning because this runs for every
+// container on the node, staged or not.
+func pciSysfsMounts(cfg Config) []Mount {
+	if cfg.HostOverlayPath == "" {
+		return nil
+	}
+	sysDevices := filepath.Join(cfg.HostOverlayPath, sysDevicesRelPath)
+	pciDevices := filepath.Join(cfg.HostOverlayPath, pciDevicesRelPath)
+	for _, dir := range []string{sysDevices, pciDevices} {
+		info, err := os.Stat(dir)
```
ArangoGutierrez, line 391: The staged-tree guard stats only sys/devices and sys/bus/pci/devices, but renderTopology creates both directories at its very start and renderDMI writes product_name/product_uuid last. A tree caught mid-render passes this check, gets mounted, and kind's createContainer hook then fails container creation on the missing bind target, which is the failure the guard exists to prevent. Since this plugin has no ordering against the staging DaemonSet (as the comment above notes), statting one of the mirrored DMI attributes too would make the guard match what the mount actually needs.

pkg/system/mockpcisysfs/render/render.go
```diff
@@ -90,6 +108,49 @@ func Render(o Options) error {
...
+			// it either and a stand-in would only invent an identity.
+			continue
+		}
+		// product_uuid is mode 0400 on most kernels: unreadable to a
+		// non-root renderer, yet still needed as a mount target.
+		contents, err := os.ReadFile(src)
+		if err != nil {
+			contents = nil
+		}
+		if err := writeFile(root, filepath.Join(dmiIDDir, attr), string(contents)); err != nil {
```
ArangoGutierrez: writeFile hardcodes 0644, so the node's product_uuid, which the kernel deliberately exposes 0400 root-only, gets republished world-readable into every served container (the DaemonSet is privileged, so the read does succeed). Low stakes on a kind node, but the value isn't needed at all: the comment above says kind bind-mounts the node's own copy over this path, so only the file's existence matters as a mount target. Writing an empty stand-in for product_uuid, the way the unreadable branch already does, drops the exposure without changing behavior.

tests/e2e/go/scenario_gpu_operator_test.go
```diff
@@ -74,6 +75,18 @@ var _ = Describe("nvml-mock GPU Operator", Label("gpu-operator"), Ordered, func(
 				assertions.WaitAllocatableGPU(ctx, h.Kube, node, p.ExpectedGPUs(), config.ReadyTimeout(), config.PollInterval())
 			})
 
+			It("serves the rendered PCI tree to the GFD container at the kernel paths", Label("device-plugin"), func(ctx SpecContext) {
+				// The NVML-derived labels above cannot distinguish "GFD read
+				// the mock tree" from "GFD read the host's sysfs and happened
+				// to agree". Reading the tree from inside the container pins
+				// the delivery itself, independent of what GFD makes of it.
+				pod, err := h.Kube.FirstPodName(ctx, gpuOperatorNamespace, "app=gpu-feature-discovery")
```
ArangoGutierrez: FirstPodName returns Items[0] with no phase filter, so a Terminating or Pending GFD pod matches the selector as well as the Running one and the ExecSh below fails. There's also no guarantee Items[0] is the pod on the node the surrounding specs target. RunningPodNames is the filtered sibling; picking the Running pod on `node` would make this spec assert about the node it means to.

ArangoGutierrez (commented):
Nice piece of debugging, and the writeup makes the root cause easy to follow: the LD_PRELOAD shim can't see Go's openat, so only a real mount at the kernel path works. The renderer split into renderTopology/renderDMI is clean, the fail-open guards on both the CDI and NRI sides are the right instinct, and the e2e that reads vendor through the symlink from inside the GFD container is exactly the assertion that distinguishes delivery from coincidence. I also like that renderDMI mirrors rather than mocks, and says so. The one thing I'd want settled before merge is the blast radius of the NRI mount: it is node-wide where the CDI path is opt-in by construction, and dropping nri.pciSysfsMounts in 853cea23 removed the only switch for it.

1. The gate checks that the two directories exist, not that this run rendered them, and the renderer only adds entries. Re-profiling a node (a100 and h100 share no BDFs) leaves both sets under $HOST/sys, the gate still flips on, and the union is what gets bind-mounted at /sys/bus/pci/devices. The missing prune predates this PR, but mounting the tree at the kernel path is what makes it visible, and the exact-count assertion in PCISysfsAtKernelPath would fail on such a node. (deployments/nvml-mock/scripts/setup.sh:140)
2. While you're rewriting the row above it, this one is now stale too: the renderer writes numa_node for every device, and it reaches served containers through the same mount, so the device plugin does get NUMA hints there. Still accurate for containers the mock doesn't serve, so it probably wants the same qualifier the row above gained. (deployments/nvml-mock/helm/nvml-mock/README.md:1409)

deployments/nvml-mock/helm/nvml-mock/README.md
```diff
@@ -708,6 +709,48 @@ DaemonSet under `set -e` if it finds a typo:
 If a profile omits `pcie_topology:` entirely the renderer falls back to
 a flat single-root layout (every device under `pci0000:00`, NUMA 0).
 
+### PCI sysfs in containers
+
+Reaching the tree through `MOCK_PCI_ROOT` requires the `libpcimocksys.so`
```
roma-glushko, line 714: So since we are doing mounting of sys directories, we don't need to use the previous way of mocking and we can remove it? Mounting should work for both type of consumers right?

deployments/nvml-mock/helm/nvml-mock/README.md
```diff
@@ -708,6 +709,48 @@ DaemonSet under `set -e` if it finds a typo:
...
+|---|---|
+| `/var/lib/nvml-mock/sys/devices` | `/sys/devices` |
+| `/var/lib/nvml-mock/sys/bus/pci/devices` | `/sys/bus/pci/devices` |
+
+Both are needed together: the entries under `/sys/bus/pci/devices` are
+relative symlinks into `../../../devices/pciDDDD:BB`, so mounting only that
+directory yields entries that list but whose every attribute read fails with
+`ENOENT`.
+
+**Trade-off:** `/sys/devices` is mounted whole, which hides the host's other
```
roma-glushko, line 735: This means that we should take the real node's devices file and "expand" it with simGPU devices in order to produce `/var/lib/nvml-mock/sys/devices`. This way all devices should be there.

This should work even if host devices file is not static. We could fnotify watch it and mirror in our produced devices file.

pkg/nri/nvmlmock/adjust.go
giuliocalzo: Agreed, and fixed in ffeef779 — though not by statting a mirrored DMI attribute. On a kernel that exposes no DMI at all (Docker Desktop's linuxkit VM) `renderDMI` legitimately writes nothing, so requiring `product_name` there would suppress the mounts on exactly the host where they work today.

The renderer now writes `sys/.rendered` last, once topology and DMI are both on disk, and this guard plus `setup.sh`'s gate key on that. It sits outside both mounted subtrees, so it is not visible to a served container.

One thing the marker cannot cover, which the prune in the same commit made explicit: a CDI-served container has no way to wait for it, because the runtime applies the spec's mounts unconditionally. So `pruneTree` empties the two directories rather than removing them, and keeps `sys/devices/virtual` — a container created mid-render finds every mount's source and every one of kind's targets in place, and may only see fewer devices than the profile declares.

pkg/system/mockpcisysfs/render/render.go
giuliocalzo: Agreed — fixed in ffeef779. `product_uuid` is now an empty stand-in: `dmiMirroredAttrs` carries a `byValue` flag, set for `product_name` only, and the file is still written whenever the kernel exposes the attribute so kind's hook keeps its target.

`TestRender_StandsInForUnreadableDMI` moved to `product_name` to keep covering the unreadable-source branch, which no longer has a caller that hits it in practice but still has to leave the target behind.

tests/e2e/go/scenario_gpu_operator_test.go
giuliocalzo: Right on both counts — fixed in 2c4b3dd7. Added `RunningPodOnNode`, which filters on `spec.nodeName` and takes the first Running pod, and the spec now passes the `node` the surrounding specs assert labels for. `RunningPodNames` alone would have fixed the phase but not the node.

pkg/nri/nvmlmock/adjust.go
giuliocalzo: Considered, and I'd rather not gate on `cfg.DeviceAnnotation` — the container that needs this mount most never sets it. GFD reads sysfs and requests no device: it resolves each GPU's BDF from NVML and reads `vendor`/`class` under `/sys/bus/pci/devices`. That is the whole reason for this PR, and on a stock `kindest/node` with no toolkit, NRI is the only channel that reaches it. Gating on the annotation would put `gpu.mode=unknown` back for every NRI-only deployment unless someone annotates the Operator's operands, which are not ours to edit. The same holds for the DRA driver.

So the difference in blast radius is real but it tracks a difference in purpose: the CDI path serves containers that ask for a mock GPU, and the NRI path is ambient injection for consumers that cannot ask — which is also why the overlay mount and `LD_PRELOAD` are already unconditional there.

On the missing switch: `nri.pciSysfsMounts` was not the only one. `nri.optOutAnnotation` (`nvml-mock.nvidia.com/inject: "false"`) exempts a single pod from injection entirely, and `nri.excludedNamespaces` exempts a namespace — the first is a per-workload opt-out, not an allowlist you have to populate up front. The chart README's trade-off paragraph now names both, since it pointed only at `excludedNamespaces` (ffeef779). What I don't want back is a knob that leaves the mock injecting NVML while sysfs disagrees with it, which is the state `gpu.mode=unknown` came from.

Happy to reopen this if you'd rather have the flag; it is a small revert. But I'd want it to be an all-or-nothing "inject on this node" switch rather than one that half-serves a container.

giuliocalzo:
Thanks — all five actionable points are addressed in ffeef779 and 2c4b3dd7, and I answered the blast-radius question in its thread (short version: gating on `nri.deviceAnnotation` would drop the mount for GFD, which reads sysfs and requests no device, so NRI-only deployments would go back to `gpu.mode=unknown`; the per-pod `nvml-mock.nvidia.com/inject: "false"` opt-out already covers the workload that wants the host's tree, and the README now says so).

On the two numbered points:

1. Fixed, in the renderer rather than in the gate. `render.Render` prunes before it writes, so `cmd/render-pci-sysfs` callers get it too and not just `setup.sh`. `TestRender_PrunesStaleDevices` renders one profile over another and asserts the previous root complex and symlink are gone. The gate itself now keys on the completion marker, which also answers "did *this* run render it" rather than "does a directory exist".
2. Fixed — the `numa_node` row gained the same qualifier, and says the served device plugin does get the hints.

`make test` (33 packages), `make helm-tests` (169 tests, 15 snapshots) and `golangci-lint run ./...` (0 issues) all pass locally; `make lint-fix` still reports the pre-existing Go stdlib advisories from my local toolchain (1.26.5 vs 1.26.6), unrelated to this branch.

pkg/system/mockpcisysfs/render/render.go
```diff
@@ -57,28 +57,115 @@ type Options struct {
...
-		return nil
+	if !o.hasTopology() {
+		// Nothing to render — the caller passed a profile with no declared
+		// topology and no devices, which setup.sh does unconditionally. A tree
+		// left here by a previous profile would still describe the node, so it
+		// is emptied rather than kept.
+		if o.Output == "" {
+			return nil
+		}
+		return pruneTree(o.Output)
```
ArangoGutierrez, line 104: `main.go` returns at its `topo == nil` check before `Render` is ever called, and `EffectiveTopology` is nil whenever no device declares a `pci.bus_id`, which `gpu.customConfig` makes reachable, so this prune never runs on the production path. Rendering a100 and then such a config into the same output directory leaves all eight A100 symlinks and `sys/.rendered` in place, so the gate in setup.sh flips `PCI_SYSFS_RENDERED=on` and both channels mount the previous profile's tree at the kernel paths; calling Render unconditionally from main.go, or clearing the marker on that early return, closes it.

Also at deployments/nvml-mock/scripts/setup.sh:146.

tests/e2e/go/framework/kube/kube.go
```diff
@@ -285,6 +285,25 @@ func (c *Client) RunningPodNames(ctx context.Context, ns, selector string) ([]st
...
+// fails, and a pod on another node answers about hardware the assertion is not
+// about.
+func (c *Client) RunningPodOnNode(ctx context.Context, ns, selector, node string) (string, error) {
+	var pl podList
+	if err := c.getJSON(ctx, &pl, "pods", "-n", ns, "-l", selector,
+		"--field-selector", "spec.nodeName="+node); err != nil {
+		return "", err
+	}
+	for _, p := range pl.Items {
+		if p.Status.Phase == "Running" {
```
ArangoGutierrez: A terminating pod keeps `status.phase: Running` until its containers exit, `deletionTimestamp` is what marks it, and the `objectMeta` this decodes into carries only name, labels and annotations, so the Terminating half of the doc above is not enforced and a GFD rollout can still hand back the pod that is going away. The call site resolves it once with no `Eventually` either, unlike `waitOperatorValidatorRunning` and the dcgm assertion, which poll because the operator replaces its operands a reconcile after nvml-mock rolls (#602).

ArangoGutierrez (commented):
Nothing here blocks merge.

The marker protocol is the right shape. Writing `sys/.rendered` last and gating both channels on it makes "this render finished" distinguishable from "the directories exist", and `pruneTree` keeping the two mounted directories and `virtual/dmi/id` in place is what lets a CDI-served container survive a re-render it has no way to wait for. `product_uuid` as an empty stand-in and `product_name` mirrored rather than mocked both land, and the new render tests each name a bug they would catch: the prune test goes red on the old additive renderer, the DMI tests on a mount target that stops existing. Moving the render to step 3b checks out too, since nothing between there and the old step 10 writes under `$HOST/sys`, and the renderer reads the ConfigMap at `/etc/nvml-mock/config.yaml` rather than the copy step 5 makes. Where the protocol still has a hole is the one path that never reaches the renderer at all.

1. The `Output` godoc still says Render is a no-op when Topology is nil or has no root complexes; that path now prunes the tree and deletes the marker whenever Output is set. (pkg/system/mockpcisysfs/render/render.go:55)
2. `MarkerRelPath` is exported from render for exactly this seam, but these two path constants are re-declared here and staged with their own literals in adjust_test.go, so a rename on the renderer side would leave `pciSysfsMounts` statting a path that no longer exists and failing open silently, with nothing red. (pkg/nri/nvmlmock/adjust.go:50)

pkg/system/mockpcisysfs/render/render.go
giuliocalzo: Correct, and this was the hole worth having found: the prune shipped in a commit whose own entry point could not reach it on that path. Fixed in 6b827e39.

`render-pci-sysfs` now calls `Render` unconditionally — a nil topology is already "empty the tree" to the renderer, so the command has no reason to decide that for itself — and `--dry-run` is the only short-circuit left. `main` is split into a `run()` so the case that regressed has a test: `TestRun_ClearsTreeWhenProfileDeclaresNoDevices` renders a profile with a `bus_id` and then one without, and asserts both the devices and the marker are gone. Against the early return it fails with `Should be empty, but was [L 0000:07:00.0]`, which is the a100-then-customConfig sequence you described.

`setup.sh`'s marker comment now also says the renderer removes the marker for a profile that renders nothing, since that is what makes the gate at :146 correct rather than the gate itself doing anything new.

tests/e2e/go/framework/kube/kube.go
giuliocalzo: Both true — fixed in 9974d6f2. My doc claimed a filter the type could not express: `objectMeta` now decodes `deletionTimestamp`, and the pod has to be phase `Running` *and* not have one, so the pod on its way out is skipped rather than exec'd into.

The call site polls too, with `ReadyTimeout`/`PollInterval`, in the same shape as `waitOperatorValidatorRunning`. Resolving once assumed a steady state that #602 says does not hold here: the operator replaces its operands a reconcile after nvml-mock rolls, so there is a window with no candidate at all, and skipping the terminating pod turns that window from a failed exec into an empty result — which only helps if the caller waits it out.

deployments/nvml-mock/helm/nvml-mock/README.md
giuliocalzo: Sorry for the slow reply — these landed while I was answering the other thread.

For the consumers this PR is about, yes: a Go consumer in a container the mock serves now reads the real thing at the kernel path, and the shim adds nothing there. But the mount cannot replace the shim in general, because a bind mount needs a target that already exists — `mount(8)` cannot create one on a read-only sysfs, which is why this PR mounts `/sys/devices` whole instead of just the profile's root complexes.

That rules the mount out wherever the path is absent on the node:

- `/sys/class/infiniband` and `/sys/class/infiniband_verbs` do not exist on a node with no IB hardware and no `ib_core`, so the IB tree is reachable only through `libibmocksys`. Same for `/dev/infiniband`.
- `/sys/class/dmi` is absent on Docker Desktop's linuxkit VM, which is why this PR mirrors DMI into the tree rather than relying on the kernel's copy.

And it only reaches containers one of the two channels serves. The nvml-mock DaemonSet is never self-injected, so `lspci` in its own pod works through `LD_PRELOAD` + `MOCK_PCI_ROOT` (set in `daemonset.yaml`), as does anything in a pod that opted out with `nvml-mock.nvidia.com/inject: "false"` or lives in an excluded namespace.

So the two mechanisms answer different questions: the shim covers any libc consumer anywhere on the node, at any path; the mount covers the paths Go consumers read with raw syscalls, in the containers we serve.

deployments/nvml-mock/helm/nvml-mock/README.md
giuliocalzo: I dug into this and filed #689 for it, because it deserves more than a paragraph here — but I do not think mirroring is the shape that works, for two reasons.

Copying gives frozen values. Sysfs attributes are kernel-backed reads, not files: `scaling_cur_freq`, thermal zones and every counter change between the read and the consumer's. A mirrored `/sys/devices/system/cpu` would look right and be stale, which is worse than a missing directory that fails loudly. Live mirroring would have to bind-mount each host entry into our tree rather than copy it — doable from a privileged DaemonSet, but then it is mounts all the way down, and inotify on sysfs does not reliably report attribute changes, so the watch would not be the part that works.

The harder problem is that no mirror of the *node's* tree can be right. `/sys/class/net/<iface>` resolves into `devices/virtual/net`, and what belongs there is the pod's own net namespace view — the kernel populates it per netns at mount time. The node's copy holds the host's interfaces, so mirroring it hands the container someone else's network.

So the way out is to stop shadowing `/sys/devices` at all. If the renderer emitted symlink targets that escape sysfs into the overlay we already mount — `/sys/bus/pci/devices/0000:07:00.0 -> ../../../../opt/nvml-mock/sys/devices/pci0000:00/0000:07:00.0` — then only `/sys/bus/pci/devices` needs mounting, a path every node has, and CPU topology, DMI and the pod's namespaced sysfs are all left alone. That is #689; the open question is whether the DRA driver's `deviceattribute` still finds the PCIe root in a path that does not start with `/sys/devices`.

giuliocalzo:
All four addressed. The two you numbered:

1. `Options.Output` no longer claims Render is a no-op without a topology; it says that path empties the tree and drops the marker, which is what it has done since the prune landed (047d6e83).
2. `PCIDevicesRelPath` and `SysDevicesRelPath` are exported from `render` now and the NRI guard takes both from there, as it already did for `MarkerRelPath`. `adjust_test.go` stages from the same constants while the mount assertions keep their literals, so renaming a path in the renderer goes red on the assertion instead of silently failing open (047d6e83).

And the two inline: the `topo == nil` early return is gone, with a test that reproduces the a100-then-`customConfig` sequence (6b827e39), and the GFD pod lookup now skips terminating pods and polls (9974d6f2).

I also answered @roma-glushko's two questions from yesterday, which I had missed. The second one — mirroring the node's `/sys/devices` into ours — turned into #689: mirroring cannot be made correct (a container's `devices/virtual/net` is netns-scoped, so the node's copy is the wrong network), but pointing the PCI symlinks outside sysfs into the overlay would mean never mounting `/sys/devices` at all, which dissolves the blast-radius question in your first review rather than trading it off. Not for this PR — it needs verifying against `deviceattribute` first.

`make test`, `golangci-lint run ./...` (0 issues) and `make helm-tests` (169 tests, 15 snapshots) pass locally.

roma-glushko:
> I also answered @roma-glushko's two questions from yesterday, which I had missed. The second one — mirroring the node's /sys/devices into ours — turned into https://github.com/NVIDIA/k8s-test-infra/issues/689: mirroring cannot be made correct (a container's devices/virtual/net is netns-scoped, so the node's copy is the wrong network), but pointing the PCI symlinks outside sysfs into the overlay would mean never mounting /sys/devices at all, which dissolves the blast-radius question in your first review rather than trading it off. Not for this PR — it needs verifying against deviceattribute first.

@giuliocalzo make sense to move this question out of scope, but I would think a bit more about what we can do there.

dharmjit:
Hi @giuliocalzo, I wanted to test the changes in PR https://github.com/NVIDIA/dcgm-exporter/pull/716 for fixing the issue [DCGM_EXPORTER_ENABLE_GPU_BIND_UNBIND_WATCH feature does not work as intended](https://github.com/NVIDIA/dcgm-exporter/issues/706), and I thought `NVIDIA/k8s-test-infra` would be a good candidate to validate this locally. I came across some current limitations around the PCI sysfs tree, which is how I landed on this PR.

For the scenario above, I’d create a separate GH issue to work on adding `writable, refcount-gated` driver-binding nodes (bind/unbind/driver_override) so that a consumer’s “drain handles on topology change” behavior can be exercised without a GPU. However, I think a couple of small improvements in this PR could make that follow-up work easier:

1. Render the driver-binding nodes (read-only) as part of the tree: In addition to `devices/` and `bus/pci/devices/`, render `bus/pci/drivers/{nvidia,vfio-pci}/` and, per device, the `driver` symlink + `driver_override`. This just completes the tree (still read-only); the follow-on then only makes these existing, already-mounted nodes writable rather than reshaping the renderer and the mount plumbing.
2. One source-of-truth list of mount specs, shared by the CDI and NRI paths: The `/sys/devices` + `/sys/bus/pci/devices` pair is currently expressed in both `setup.sh`(CDI) and `adjust.go`(`pciSysfsMounts()`). If they read from a single list where each entry carries `ro/rw` + propagation, the follow-on appends one `rw` entry in one place instead of editing two paths that could drift.

PR #682, opened 2026-08-18
feat(mocknvml): report Platform Info for Grace-Blackwell profiles

### LINKED ISSUE #642: [Feature]: gb200 profile reports no platform identity (chassis, slot, tray, host ID, peer type, module ID)
## Problem statement

On the `gb200` profile — a Grace-Blackwell NVL72 rack — `nvidia-smi -q` reports nothing for the entire `Platform Info` block. Chassis serial number, slot number, tray index, host ID, peer type, and module ID are all `N/A`.

These are precisely the fields that rack-scale fault correlation depends on. On NVL72, knowing that a failing GPU is module 2 of the tray in slot 9 of a given chassis is what turns a GPU fault into an actionable physical location. Software that consumes this topology cannot be tested against the mock today, and the `gb200`/`gb300` profiles are the reason those consumers exist.

Found during a full audit of `nvidia-smi -q` (all 195 leaf fields, profiles `t4`, `a100`, `h100`, `gb200`) on `k8s-test-infra` @ `a4cf377a`, `nvml-mock` chart `0.3.0`.

### Example output

`profiles/gb200.yaml`, GPU 0:

```
GPU 0000:0A:00.0
    Product Name                          : NVIDIA GB200
    Product Brand                         : NVIDIA
    Product Architecture                  : Blackwell
    ...
    Platform Info
        Chassis Serial Number             : N/A
        Slot Number                       : N/A
        Tray Index                        : N/A
        Host ID                           : N/A
        Peer Type                         : N/A
        Module Id                         : N/A
        GPU Fabric GUID                   : N/A
```

For contrast, the fabric state on the same profile **is** modeled correctly, which shows the gap is specific to platform identity rather than to NVL72 support in general:

```
    Fabric
        State                             : Completed
        Status                            : Success
        CliqueId                          : 0
        ClusterUUID                       : 00000000-0000-0000-0000-000000000001
        Health
            Summary                       : N/A
            Bandwidth                     : N/A
            Route Recovery in progress    : N/A
            Route Unhealthy               : N/A
            Access Timeout Recovery       : N/A
            Incorrect Configuration       : N/A
```

### Root cause

`nvmlDeviceGetPlatformInfo` is a generated stub returning `NVML_ERROR_NOT_SUPPORTED`, and there is no configuration surface for any of these fields — unlike `c2c_enabled`, which at least exists in config even though it is dropped. `nvmlDeviceGetModuleId` is separately stubbed. `MOCK_NVML_DEBUG=1` shows both being asked for once per GPU:

```
   8 nvmlDeviceGetModuleId called (NOT IMPLEMENTED)
   8 nvmlDeviceGetPdi called (NOT IMPLEMENTED)
```

On `a100` and `t4` the same `N/A` rows are correct, since those are not Grace platforms. The defect is that a Grace-Blackwell profile is indistinguishable from a PCIe workstation card by platform identity.

## Proposed solution

Add a platform-identity configuration surface and implement the backing calls:

- Extend the device config with chassis serial number, slot number, tray index, host ID, peer type, and module ID, settable per device so an eight-GPU profile can describe a realistic tray-and-slot layout rather than repeating one value.
- Implement `nvmlDeviceGetPlatformInfo` using NVML's versioned-struct convention, following the already-implemented `GpuFabricInfoV` path as the model.
- Implement `nvmlDeviceGetModuleId`.
- Populate the `gb200` and `gb300` profiles with a coherent NVL72 layout, so the shipped profiles are useful without hand-editing.
- Leave the fields absent, and therefore `N/A`, on non-Grace profiles.

Alternatives considered: hardcoding plausible constants inside the Blackwell code path would remove the `N/A` but would make every mock GPU report the same physical location, which defeats the fault-correlation use case.

Separately worth considering in the same area, though not required here: `Fabric Health` rows are `N/A` while the rest of the `Fabric` block is populated.

## Scope / non-goals

In scope: the six `Platform Info` fields, `nvmlDeviceGetPlatformInfo`, `nvmlDeviceGetModuleId`, and realistic NVL72 values in the `gb200` and `gb300` profiles.

Non-goals: `GPU Fabric GUID` and the `Fabric Health` rows, which can be tracked separately; `GPU C2C Mode`, tracked in its own issue; `nvmlDeviceGetPdi`; and the roughly 60 `-q` fields that are faithfully `N/A` on this hardware.

## Audience

Internal NVIDIA teams.

## Acceptance criteria

- [ ] Chassis serial number, slot number, tray index, host ID, peer type, and module ID are configurable per device.
- [ ] `nvmlDeviceGetPlatformInfo` and `nvmlDeviceGetModuleId` are implemented and removed from `stubs_generated.go`, with the versioned-struct layout pinned by a unit test so a wrong layout cannot fail silently.
- [ ] `nvidia-smi -q` on `gb200` and `gb300` reports all six fields with the configured values.
- [ ] The shipped `gb200` and `gb300` profiles describe a coherent NVL72 layout in which GPUs differ by module and tray rather than sharing one identity.
- [ ] Non-Grace profiles (`a100`, `t4`, `h100`, `l40s`) continue to report `N/A`.
- [ ] **Dedicated e2e test.** A new Ginkgo spec under `tests/e2e/go/` with `Label("nvidia-smi")` that, on `gb200`, asserts every `Platform Info` field is populated rather than `N/A` and matches the profile's configured values; asserts the module ID and tray/slot values are **distinct across GPUs** so a physical location can actually be derived per device; and asserts the same fields remain `N/A` on `a100`. The distinctness and the negative `a100` case are both required, so the fix cannot be a shared constant. The test must fail against the current `main`.

giuliocalzo:
Fixes #642

## Summary

`nvidia-smi -q` read `N/A` for every row of the `Platform Info` block on `gb200` and `gb300`, so an NVL72 rack was indistinguishable from a PCIe workstation card by physical location — the fields rack-scale fault correlation reads to turn a GPU fault into "module 2 of the tray in slot 9 of this chassis".

- `nvmlDeviceGetPlatformInfo` and `nvmlDeviceGetModuleId` are implemented and removed from `stubs_generated.go`. The former dispatches on the caller's struct version tag and serves both v1 and v2 from one payload, since the rename (`rackGuid` -> `chassisSerialNumber`) kept the same 44 bytes. The layout is pinned by `_Static_assert` in `nvml_types.h` and by a unit test, so a wrong layout cannot fail silently.
- A `device_defaults.platform` block configures the identity. Only `module_id` is set per device: NVML scopes the other five fields to the node, which occupies exactly one tray of one chassis, so the shipped profiles vary GPUs by module while sharing the node's location. Per-device merging is copy-on-write, or every GPU would inherit the last override's module id.
- `gb200` and `gb300` describe a coherent NVL72 layout. `slot_number` counts switch trays as well as compute trays while `tray_index` counts compute trays only, so slot runs ahead of tray for a tray above the rack's nine switch trays: compute tray 11 lands in slot 21.
- Profiles declaring no block keep reporting `N/A`, which stays correct for every board outside such a rack.

`gb200`, GPU 0 and 1:

```
    Platform Info
        Chassis Serial Number             : 1822725100200
        Slot Number                       : 21
        Tray Index                        : 11
        Host ID                           : 1
        Peer Type                         : Switch Connected
        Module Id                         : 1
        GPU Fabric GUID                   : 0x0000000000000000
    Platform Info
        ...
        Module Id                         : 2
```

## Notes for review

`GPU Fabric GUID` is a non-goal of the issue and is not modelled, but it sits in the same struct and now renders `0x0000000000000000` where it used to read `N/A`. The bridge zeroes that buffer deliberately, so the row cannot render the caller's uninitialised memory; the changelog records it as unmodelled.

`b200` turned out to be the negative control the acceptance criteria were really after — Blackwell, driver past the 560 version gate, but in no rack — so its `N/A` proves the reading is driven by configuration rather than by the version gate. It is covered alongside `a100`.

## Test plan

- [x] New Ginkgo spec `reports the profile's platform identity via nvidia-smi` under `Label("nvidia-smi")`, deriving its expectation from the profile so it asserts both directions: populated, node-consistent and module-distinct on `gb200`; `N/A` on `a100` and `b200`. Passes against a freshly built image on all three.
- [x] Red against `main`: the profile guard fails on `main`'s profiles, which declare no platform block, and the assertion rejects an all-`N/A` document against the `gb200` expectation with 12 problems — `main`'s exact output.
- [x] `make test`, `make helm-tests` (snapshots updated for the two profiles), `make gen-check`, `go vet`, `golangci-lint`.
- [x] `docs/configuration.md` documents the configuration surface; `CHANGELOG.md` entry under `[Unreleased]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- mocknvml: `nvidia-smi -q` now reports the `Platform Info` block on the
+  `gb200`/`gb300` profiles — chassis serial number, slot number, tray index,
+  host ID, peer type and module ID, where every row previously read `N/A`. These
+  are the fields rack-scale fault correlation reads to turn a GPU fault into a
+  physical location, so a Grace-Blackwell rack was indistinguishable from a PCIe
+  workstation card by location. `nvmlDeviceGetPlatformInfo` (both struct
+  versions) and `nvmlDeviceGetModuleId` are implemented and no longer generated
+  stubs, and a new `device_defaults.platform` block configures the identity —
+  per device, though only `module_id` varies between the GPUs of a node, since
+  NVML scopes the rest to the node, which sits in exactly one tray of one
+  chassis. Profiles that declare no block, and any profile on a driver older
+  than 560, keep reporting `N/A`. The `GPU Fabric GUID` row of the same block is
+  not modelled and now renders `0x0000000000000000` where it used to read `N/A`.
+  (#642)
 - mocknvml: configured `processes:` now surface in nvidia-smi — the default
   table's Processes box, `-q`, and `--query-compute-apps` all report the
   configured PIDs, names and GPU memory instead of always reporting none.
--- a/deployments/nvml-mock/helm/nvml-mock/profiles/gb200.yaml
+++ b/deployments/nvml-mock/helm/nvml-mock/profiles/gb200.yaml
@@ -58,6 +58,26 @@ device_defaults:
     health_mask: 0
 
   # ---------------------------------------------------------------------------
+  # Platform identity — where this node sits in the NVL72 rack, which
+  # nvidia-smi -q renders as its "Platform Info" block and rack-scale fault
+  # correlation reads to turn a GPU fault into a physical location.
+  #
+  # Every field but module_id describes the node, so all its GPUs share them:
+  # a node occupies one tray, in one slot, of one chassis. Only module_id
+  # ("ID of this GPU within the node") is set per device below.
+  #
+  # slot_number counts switch trays as well as compute trays while tray_index
+  # counts compute trays only, so slot runs ahead of tray for a tray sitting
+  # above the rack's nine switch trays: compute tray 11 lands in slot 21.
+  # ---------------------------------------------------------------------------
+  platform:
+    chassis_serial_number: "1822725100200"
+    slot_number: 21
+    tray_index: 11
+    host_id: 1                      # first of the tray's two OS domains
+    peer_type: "switch_connected"   # peers reached through the NVSwitch trays
+
+  # ---------------------------------------------------------------------------
   # InfoROM versions
   # ---------------------------------------------------------------------------
   inforom:
@@ -375,6 +395,8 @@ devices:
       bus_id: "0000:0A:00.0"
     minor_number: 0
     grace_cpu_pair: 0               # Paired with Grace CPU 0
+    platform:
+      module_id: 1                  # ID of this GPU within the node
 
   - index: 1
     uuid: "GPU-b200b200-0000-0000-0000-000000000001"
@@ -383,6 +405,8 @@ devices:
       bus_id: "0000:0B:00.0"
     minor_number: 1
     grace_cpu_pair: 0               # Same superchip as GPU 0
+    platform:
+      module_id: 2
 
   - index: 2
     uuid: "GPU-b200b200-0000-0000-0000-000000000002"
@@ -391,6 +415,8 @@ devices:
       bus_id: "0000:4A:00.0"
     minor_number: 2
     grace_cpu_pair: 1
+    platform:
+      module_id: 3
 
   - index: 3
     uuid: "GPU-b200b200-0000-0000-0000-000000000003"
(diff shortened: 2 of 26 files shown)


ArangoGutierrez (commented):
No issues found — checked for bugs and CLAUDE.md compliance.

ArangoGutierrez (approved):
Approving. The v1/v2 struct layout matches the vendored NVML header field-for-field and the 44-byte static assert lines up with go-nvml's PlatformInfo_v2, so the shared-payload dispatch holds. The copy-on-write in mergePlatformOverride is covered by a test that goes red if you drop it, and the N/A profiles are a real negative control.

PR #683, opened 2026-08-19
fix(local): pin kind worker node names and select the fleet by hostname

giuliocalzo:
## Summary

- Pins the worker node names in `local/kind/default.kind.yaml` to `worker-0` / `worker-1` via `JoinConfiguration`, instead of letting them be derived from the cluster name (`mokka-worker`, `mokka-worker2`). Stable names make node-targeted commands and docs reproducible across cluster recreations.
- Switches `install_fleet()` in `local/nvml_mock.tiltfile` to pin each per-profile release with `nodeSelector.kubernetes.io/hostname=<worker>`, and drops the now-redundant `nvml-mock/profile` worker labels. The mapping is positional: the Nth entry of `FLEET_PROFILES` lands on `worker-N`.
- Fixes the e2e harness to discover node names from the Kubernetes API. `Cluster.Nodes()` previously enumerated `kind get nodes`, which reports **docker container** names, and passed them straight to `kubectl` — that works only while the two happen to be identical.

## Why the harness change is required

Pinning `nodeRegistration.name` decouples the Kubernetes node name from the Kind container name. Without the harness fix, every scenario that resolves a node by name fails with `Error from server (NotFound): nodes "mokka-worker" not found`.

`cluster.Node` now carries both identities: `Name` (Kubernetes, for `kubectl`) and `Container` (docker, for `docker exec`). The container name is read from Kind's provider ID, `kind://docker/<cluster>/<container>`, which keeps pointing at the container after a rename. The `docker exec` call sites in the GPU Operator and NRI scenarios were updated to use `Container`. Roles now come from the `node-role.kubernetes.io/control-plane` label rather than a `"worker"` substring match, since a pinned name need not describe its role.

The control-plane node deliberately keeps its Kind-derived name: kubeadm names the bootstrap node from `InitConfiguration`, so a `JoinConfiguration` patch there is silently ignored.

## Test plan

- [x] `make test-e2e-framework` and `make test` pass; `golangci-lint --build-tags e2e ./tests/e2e/go/...` reports 0 issues
- [x] New unit tests in `cluster_test.go` cover renamed nodes, label-derived roles, non-Kind provider IDs, ordering, and malformed input
- [x] Verified against a live Kind cluster with a renamed worker: node `worker-0` resolves to container `rename-probe-worker`
- [x] Verified unchanged behaviour on a cluster without renames: names and containers match, roles correct
- [x] `helm template` renders a flat `kubernetes.io/hostname: worker-0` under `nodeSelector`
- [x] `tilt alpha tiltfile-result -- --multi-gpu-profile` pins `a100` to `worker-0` and `t4` to `worker-1`
- [ ] Full e2e suite in CI (gpu-operator / nri / nfd / multi-node were the jobs broken by the rename)

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -50,7 +50,7 @@ config.define_string('k8s-context', args=False,
 # --dra` parses as three separate flags without positional-value
 # ambiguity that Tilt's string-flag parser can exhibit for `=X` forms.
 config.define_bool('multi-gpu-profile', args=False,
-    usage='Install one nvml-mock release per GPU profile, node-pinned via nodeSelector.nvml-mock/profile=<profile>. Simulates a heterogeneous fleet on the default cluster (a100 + t4 workers). Without this flag, a single nvml-mock release covers all nodes with the profile from --gpu-profile.')
+    usage='Install one nvml-mock release per GPU profile, node-pinned via nodeSelector.kubernetes.io/hostname=<worker>. Simulates a heterogeneous fleet on the default cluster (a100 on worker-0 + t4 on worker-1). Without this flag, a single nvml-mock release covers all nodes with the profile from --gpu-profile.')
 config.define_bool('compute-domain', args=False,
     usage='ComputeDomain scenario: 4-worker cluster with GB200 profile + NVLink topology overlay (requires PROFILE=compute-domain cluster)')
 config.define_bool('gpu-operator', args=False,
--- a/local/README.md
+++ b/local/README.md
@@ -54,7 +54,7 @@ Supported `--gpu-profile` values: `a100`, `h100`, `b200`, `gb200`, `gb300`, `l40
 
 ### Heterogeneous fleet (per-GPU-profile releases)
 
-Installs one nvml-mock release per worker, pinned by node selector `nvml-mock/profile=<profile>`. Profiles are fixed to `a100` and `t4` (matching the worker labels in `local/kind/default.kind.yaml`). `--gpu-profile` is ignored in this mode.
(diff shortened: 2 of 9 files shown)


ArangoGutierrez (commented):
Nothing here blocks merge.

The Name/Container split lands cleanly. Every docker exec path now takes Container and the kube-side lookups still take Name, including the one place it would have been easy to miss (nriPluginPodOnNode gets victim.Name, wedgeNRIPlugin gets victim.Container). Reading the role off node-role.kubernetes.io/control-plane rather than a substring of the name is the right move once names are pinned, and containerName's fallback keeps a non-Kind providerID from producing garbage. The helm escaping checks out too: flags reach helm_resource as argv, so the backslash in kubernetes\.io is not re-read by a shell. I ran the new package tests and they pass under make test-e2e-framework, which CI does invoke.

1. All three inputs here are already decided by the regex, so the length bound is untested: deleting len(name) == 0 || len(name) > 40 leaves the test green. A 41-character lowercase name would pin the bound the doc comment justifies. (tests/e2e/go/framework/cluster/cluster_test.go:99)

giuliocalzo:
Good catch on the untested length bound — you're right that every existing case fails on the regex alone, so `len(name) == 0 || len(name) > 40` could have been deleted with the test staying green.

Fixed in 5ac2068a with cases either side of the bound: a 40-character name passes, a 41-character one fails. I verified it actually pins the check by deleting the length guard and confirming the test goes red at the 41-character assertion, then restoring it.

roma-glushko (approved):
I like the idea to keep consistent. I think the other cluster already has stable worker names.

PR #684, opened 2026-08-19
feat(mocknvml): model SRAM ECC errors and row-remap availability

### LINKED ISSUE #641: [Feature]: model SRAM ECC errors and row-remap availability so SRAM fault handling can be tested
## Problem statement

`nvml-mock` exists so GPU fault-handling software can be tested without physical hardware, and SRAM ECC errors plus row-remap exhaustion are among the most important failure modes that software has to handle. Today the mock cannot simulate them: every SRAM-related field in `nvidia-smi -q` reports `N/A`, so a consumer reading SRAM counters sees an unsupported feature rather than a healthy or failing GPU.

DRAM error counters are modeled and report `0`, which makes the gap easy to miss — the ECC section looks half-populated and plausible until you notice that only the DRAM half is real.

Found during a full audit of `nvidia-smi -q` (all 195 leaf fields, profiles `t4`, `a100`, `h100`, `gb200`) on `k8s-test-infra` @ `a4cf377a`, `nvml-mock` chart `0.3.0`.

### Example output

`profiles/a100.yaml` with `ECC Mode: Enabled`. Every SRAM row is `N/A` while the DRAM rows report real counts:

```
    ECC Mode
        Current                           : Enabled
        Pending                           : Enabled
    ECC Errors
        Volatile
            SRAM Correctable              : N/A
            SRAM Uncorrectable Parity     : N/A
            SRAM Uncorrectable SEC-DED    : N/A
            DRAM Correctable              : 0
            DRAM Uncorrectable            : 0
        Aggregate
            SRAM Correctable              : N/A
            SRAM Uncorrectable Parity     : N/A
            SRAM Uncorrectable SEC-DED    : N/A
            DRAM Correctable              : 0
            DRAM Uncorrectable            : 0
            SRAM Threshold Exceeded       : N/A
        Aggregate Uncorrectable SRAM Sources
            SRAM L2                       : N/A
            SRAM SM                       : N/A
            SRAM Microcontroller          : N/A
            SRAM PCIE                     : N/A
            SRAM Other                    : N/A
        Channel Repair Pending            : N/A
        TPC Repair Pending                : N/A
```

and, in the row-remapping section:

```
    Remapped Rows
        Correctable Error                 : 0
        Uncorrectable Error               : 0
        Pending                           : No
        Remapping Failure Occurred        : No
        Bank Remap Availability Histogram : N/A
```

A real ECC-enabled A100 reports numbers for the SRAM counters, a yes/no for `SRAM Threshold Exceeded`, and a populated availability histogram.

### Root cause

Three separate gaps combine:

1. Only ECC field IDs 3–6 (`SBE`/`DBE`, volatile and aggregate DRAM totals) are modeled in `pkg/gpu/mocknvml/engine/field_values.go`. There are no SRAM-specific field IDs, which is where nvidia-smi reads the SRAM rows from.
2. `nvmlDeviceGetSramEccErrorStatus` is a generated stub, so `SRAM Threshold Exceeded` cannot be answered.
3. `nvmlDeviceGetRowRemapperHistogram` is a generated stub, so `Bank Remap Availability Histogram` cannot be answered.

`MOCK_NVML_DEBUG=1` confirms nvidia-smi asks for the last two once per GPU:

```
   8 nvmlDeviceGetSramEccErrorStatus called (NOT IMPLEMENTED)
   8 nvmlDeviceGetRowRemapperHistogram called (NOT IMPLEMENTED)
```

Note also that `nvmlDeviceGetMemoryErrorCounter` is implemented but ignores its `locationType` argument, returning the same counter for DRAM and SRAM alike. Any fix should make that argument meaningful.

## Proposed solution

Add a configuration surface for SRAM ECC state and row-remap availability, then wire it through the field-value path and the two stubbed exports:

- Extend the ECC config with SRAM counters — correctable, uncorrectable parity, uncorrectable SEC-DED — for both volatile and aggregate scopes, plus the per-source aggregate breakdown (L2, SM, microcontroller, PCIe, other) and an `SRAM Threshold Exceeded` flag.
- Add the SRAM ECC field IDs to `getDeviceFieldValue` so nvidia-smi and DCGM read the same values.
- Implement `nvmlDeviceGetSramEccErrorStatus` and `nvmlDeviceGetRowRemapperHistogram`.
- Make `nvmlDeviceGetMemoryErrorCounter` honour `locationType`, so DRAM and SRAM counters are distinguishable.
- Extend the failure injector so SRAM ECC errors can be injected at runtime through `nvml-mock-ctl`, the way ECC uncorrectable already can, since injecting a fault mid-test is the main use case.

Alternatives considered: reporting hardcoded zeros for the SRAM rows would remove the `N/A` and make the ECC section look complete, but it would not let anyone test SRAM fault handling, which is the actual goal — so it is rejected.

## Scope / non-goals

In scope: SRAM ECC counters and status, the aggregate per-source breakdown, the row-remap availability histogram, honouring `locationType`, and runtime injection.

Non-goals: `Channel Repair Pending` and `TPC Repair Pending`, which are driven by `nvmlDeviceGetRepairStatus` and can be tracked separately; and the roughly 60 `-q` fields that are faithfully `N/A` on this hardware.

## Audience

Internal NVIDIA teams.

## Acceptance criteria

- [ ] SRAM ECC state is configurable per device, and the values appear in `nvidia-smi -q` under both `Volatile` and `Aggregate` rather than `N/A`.
- [ ] `SRAM Threshold Exceeded` and `Bank Remap Availability Histogram` report configured values.
- [ ] `nvmlDeviceGetMemoryErrorCounter` returns different counters for DRAM and SRAM locations.
- [ ] The same values are visible through the field-value path used by DCGM, not only through `nvidia-smi`.
- [ ] SRAM ECC errors can be injected at runtime via `nvml-mock-ctl` without restarting the workload.
- [ ] Unit tests cover the new field IDs, the two new exports, and `locationType` handling.
- [ ] **Dedicated e2e test.** A new Ginkgo spec under `tests/e2e/go/` with `Label("failure-injection")` and `Label("runtime-control")` that: (1) asserts a healthy profile reports `0` rather than `N/A` for the SRAM counters; (2) injects SRAM ECC errors at runtime via `nvml-mock-ctl`; (3) asserts `nvidia-smi -q` reflects the injected counts, that `SRAM Threshold Exceeded` flips as configured, and that the per-source breakdown attributes them to the expected source; and (4) clears the injection and asserts the counters return to baseline. The inject-and-clear cycle is the point, since simulating a fault mid-test is the capability being added. The test must fail against the current `main`.

giuliocalzo:
## Summary

`nvmlDeviceGetSramEccErrorStatus` and `nvmlDeviceGetRowRemapperHistogram` were generated stubs, so every SRAM row of `nvidia-smi -q` and the Bank Remap Availability Histogram read `N/A`. `N/A` means "this GPU cannot tell you", which is the one answer real ECC-enabled hardware never gives — a healthy A100 reports `0` — so SRAM fault handling could not be tested at all, and the half-populated ECC section hid the gap behind the real DRAM counters beside it.

- New `ecc.sram` config block: correctable, uncorrectable-parity and uncorrectable-SEC-DED counters for the volatile and aggregate scopes, the aggregate per-unit source breakdown (L2, SM, microcontroller, PCIe, other) and the threshold-exceeded flag. Both stubbed exports are now answered from it.
- `nvmlDeviceGetMemoryErrorCounter` honours `locationType` and `counterType` instead of returning the DRAM counter for every location, and the per-location ECC field ids (7–28) resolve through it, so DCGM's field-value path reads what `nvidia-smi` reads.
- `remapped_rows.availability_histogram` carries bank remap capacity. The Ampere-and-later profiles ship it (16 banks per GiB, anchored on the 640 a real 40 GiB A100 reports); `t4` leaves it out and keeps reporting the histogram unsupported, as Turing does. An unconfigured device answers `NOT_SUPPORTED` rather than inventing a bank population, and a GPU with ECC off reports no SRAM counters at all.
- `nvml-mock-ctl sram-ecc --gpu <idx> [--type correctable|parity|secded] [--source l2|sm|microcontroller|pcie|other] [--threshold-exceeded] <count>` injects into a running workload; `0` heals. It writes both scopes (hardware that just took a fault reports it in both) and never rewrites the ECC mode, since a GPU with ECC off reports nothing and the injection would erase itself.

Fixes #641

## Test plan

- [x] `make lint-fix` — 0 golangci-lint issues (`govulncheck` reports only pre-existing Go 1.26.5 stdlib advisories)
- [x] `make test` — new unit tests cover the two exports, the C struct layouts and version tag, `locationType`/`counterType` handling, the per-location field ids, the `sram-ecc` patch builder and its CLI validation
- [x] `make helm-tests` — chart snapshots updated for the profile change
- [x] `make gen-check`
- [x] New Ginkgo specs: a `Label("nvidia-smi")` baseline asserting `0` rather than `N/A` plus the profile's configured histogram, and a `Label("failure-injection")`/`Label("runtime-control")` spec that injects at runtime, asserts the counts, the threshold flip and the source attribution, then clears and asserts the return to baseline. The `nvidia-smi` XML checks are unit-tested against the captured pre-fix fixtures, so they fail against current `main`.
- [x] Verified against the real `nvidia-smi` 580.65.06 binary over a freshly built mock (a100 profile): healthy GPUs report `0` and `SRAM Threshold Exceeded : No` with a populated `Max : 640 bank(s)`; after injection the counters read `4` in both scopes with `SRAM SM : 4` and `Yes`; healing and `reset` return to baseline with `ECC Mode : Enabled` intact; `t4` still reports the histogram `N/A`, and an ECC-disabled device reports the SRAM counters `N/A`.

### DIFF
--- a/pkg/gpu/mocknvml/bridge/ecc_layout_test.go
+++ b/pkg/gpu/mocknvml/bridge/ecc_layout_test.go
@@ -1,0 +1,75 @@
+// Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Layout tests for the ECC structs the bridge writes into. See
+// fabric_layout_test.go for why the expected sizes are constants rather than
+// C.sizeof_* reads.
+//
+// The SRAM status size matters twice over: the bridge writes twelve fields into
+// a caller-allocated buffer, and the size is half of the version tag the caller
+// must present, so drift would turn every SRAM query into a version mismatch.
+
+package main
+
+import (
+	"testing"
+	"unsafe"
+
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+	"github.com/stretchr/testify/require"
+)
+
+// Expected byte sizes derived from nvml_types.h field-by-field. Update in
+// lockstep with any C struct change.
+//
+//	EccSramErrorStatus: version(4) + 4 pad + 11 * 8 + bThresholdExceeded(4)
+//	                    + 4 trailing pad = 104
+//	RowRemapperHistogramValues: 5 * 4 = 20
+const (
+	expectedEccSramErrorStatusSize         uintptr = 104
+	expectedRowRemapperHistogramValuesSize uintptr = 20
+)
+
+func TestECCStructLayouts(t *testing.T) {
+	cases := []struct {
+		name     string
+		goSize   uintptr
+		expected uintptr
+	}{
+		{"EccSramErrorStatus", unsafe.Sizeof(nvml.EccSramErrorStatus{}), expectedEccSramErrorStatusSize},
+		{"EccSramErrorStatus_v1", unsafe.Sizeof(nvml.EccSramErrorStatus_v1{}), expectedEccSramErrorStatusSize},
+		{
+			"RowRemapperHistogramValues",
+			unsafe.Sizeof(nvml.RowRemapperHistogramValues{}),
+			expectedRowRemapperHistogramValuesSize,
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			require.Equal(t, tc.expected, tc.goSize,
+				"%s: go-nvml = %d bytes, C layout expects %d bytes — ABI drift; update either go-nvml vendor or nvml_types.h",
+				tc.name, tc.goSize, tc.expected)
+		})
+	}
+}
+
+// TestEccSramErrorStatusVersion_MatchesGoNvml pins the version tag the bridge
+// demands against the one a caller built from go-nvml's STRUCT_VERSION helper
+// produces. nvidia-smi stamps this field before calling; if the two ever
+// disagree, every SRAM query answers ARGUMENT_VERSION_MISMATCH and the SRAM
+// rows silently go back to reporting N/A.
+func TestEccSramErrorStatusVersion_MatchesGoNvml(t *testing.T) {
+	theirs := nvml.STRUCT_VERSION(nvml.EccSramErrorStatus_v1{}, 1)
+	ours := FabricStructVersion(expectedEccSramErrorStatusSize, 1)
+	require.Equal(t, theirs, ours)
+}
--- a/pkg/gpu/mocknvml/bridge/ecc_layout_test.go
+++ b/pkg/gpu/mocknvml/bridge/ecc_layout_test.go
@@ -68,8 +68,19 @@ func TestECCStructLayouts(t *testing.T) {
 // produces. nvidia-smi stamps this field before calling; if the two ever
 // disagree, every SRAM query answers ARGUMENT_VERSION_MISMATCH and the SRAM
 // rows silently go back to reporting N/A.
+//
+// Calling the production function is what makes this a real pin: it reads
+// sizeof(C.nvmlEccSramErrorStatus_t), so the C struct sits on one side of the
+// comparison and go-nvml's on the other. Recomputing the tag from
+// expectedEccSramErrorStatusSize instead would compare two Go-derived values and
+// stay green through exactly the C-side drift described above. The size constant
+// still earns its keep in TestECCStructLayouts, where it documents the
+// hand-written layout that nvml_types.h is supposed to have.
 func TestEccSramErrorStatusVersion_MatchesGoNvml(t *testing.T) {
 	theirs := nvml.STRUCT_VERSION(nvml.EccSramErrorStatus_v1{}, 1)
-	ours := FabricStructVersion(expectedEccSramErrorStatusSize, 1)
-	require.Equal(t, theirs, ours)
+	ours := sramEccErrorStatusVersion()
+	require.Equal(t, theirs, ours,
+		"bridge demands 0x%x, a go-nvml caller stamps 0x%x — every SRAM query would answer "+
+			"ARGUMENT_VERSION_MISMATCH; reconcile nvml_types.h with the vendored go-nvml struct",
+		ours, theirs)
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   than 560, keep reporting `N/A`. The `GPU Fabric GUID` row of the same block is
   not modelled and now renders `0x0000000000000000` where it used to read `N/A`.
   (#642)
+- mocknvml: SRAM ECC errors and row-remap availability are now modelled, so SRAM
+  fault handling can be tested. Every SRAM row of `nvidia-smi -q` read `N/A`
+  while `nvmlDeviceGetSramEccErrorStatus` and
+  `nvmlDeviceGetRowRemapperHistogram` were generated stubs, which told a consumer
+  the feature was unsupported rather than that the GPU was healthy. A new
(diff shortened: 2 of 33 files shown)


pkg/gpu/mocknvml/bridge/ecc_layout_test.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// TestEccSramErrorStatusVersion_MatchesGoNvml pins the version tag the bridge
+// demands against the one a caller built from go-nvml's STRUCT_VERSION helper
+// produces. nvidia-smi stamps this field before calling; if the two ever
+// disagree, every SRAM query answers ARGUMENT_VERSION_MISMATCH and the SRAM
+// rows silently go back to reporting N/A.
+func TestEccSramErrorStatusVersion_MatchesGoNvml(t *testing.T) {
+	theirs := nvml.STRUCT_VERSION(nvml.EccSramErrorStatus_v1{}, 1)
+	ours := FabricStructVersion(expectedEccSramErrorStatusSize, 1)
```
ArangoGutierrez: `ours` is built from the `expectedEccSramErrorStatusSize` constant, so both sides of this comparison are Go-derived and the C struct never gets read. Adding a field to `nvmlEccSramErrorStatus_t` leaves this test and `TestECCStructLayouts` green while `sramEccErrorStatusVersion()` moves from 16777320 to 16777328, which is exactly the ARGUMENT_VERSION_MISMATCH the comment above says this pins. Calling `sramEccErrorStatusVersion()` here compiles fine and does catch it.

ArangoGutierrez (commented):
Nothing here blocks merge.

Checked the parts the description leans on. The v1 struct layout matches the vendored go-nvml field for field (104 and 20 bytes), the ECC field ids 7 through 28 line up with nvml.h, and the pre-Ampere path does what the comment claims: no availability_histogram in t4, ERROR_NOT_SUPPORTED out of GetRowRemapperHistogram, and the t4 profile used as the e2e negative control. GetMemoryErrorCounter honouring all three arguments is a real fix rather than a refactor, since the old one took counterType as _ and returned 0 for every location. Helm snapshots were regenerated, and dropping the two generated stubs is safe under make generate because the generator scans the bridge for hand-written exports.

1. The bank count is matched with `strings.Contains` over the whole `<row_remapper_histogram>` body, so it never pins which bucket the count landed in. Moving all 640 banks from `max` to `none`, the exhausted GPU this feature exists to model, still reports no problems, and expecting 64 against an actual 640 passes as well. Comparing per bucket would close both. (tests/e2e/go/assertions/nvidiasmi/checks.go:506)
2. `sramLocationErrors` reads `ecc.sram` without checking ECC mode, while `GetSramEccErrorStatus` just above returns ERROR_NOT_SUPPORTED when it is off. After the documented `set ecc.mode_current=disabled`, the same GPU answers NOT_SUPPORTED through one SRAM API and SUCCESS with a live count through the other. (pkg/gpu/mocknvml/engine/device.go:2069)
3. The old `GetMemoryErrorCounter` took `counterType` as `_` and returned 0 for every location, so it never returned the DRAM counter for the other locations. The part about now honouring both arguments is right, it is just the before state that is off. (CHANGELOG.md:41)

pkg/gpu/mocknvml/bridge/ecc_layout_test.go
giuliocalzo: Confirmed and fixed in 61094d98 — `ours` now calls `sramEccErrorStatusVersion()`, so `sizeof(C.nvmlEccSramErrorStatus_t)` is on one side and go-nvml's struct on the other.

I checked the claim by adding an `unsigned long long` to `nvmlEccSramErrorStatus_t` in a scratch checkout: the constant-derived form passed with the drift in place, and the new form fails with `expected: 0x1000068, actual: 0x1000070` — the 16777320 → 16777328 move you predicted.

Worth noting for the compile question: the "no cgo in test files for packages with //export directives" caveat in `fabric_layout_test.go` doesn't apply here, since the test calls a production function that reads the size internally rather than importing `C` itself. The size constant stays for `TestECCStructLayouts`, where it documents the intended hand-written layout.

The same shape exists in two places outside this PR — `fabric_layout_test.go` computes its tags from `unsafe.Sizeof(nvml.GpuFabricInfo_v2{})` while `fabric.go` dispatches on `sizeof(C.nvmlGpuFabricInfo_v2_t)`, and `platform_dispatch_test.go` uses the `platformStructSize = 44` constant against `platform.go`'s C reads. Both are pinned only Go-side today. Happy to fix them here or in a follow-up, whichever you prefer.

PR #685, opened 2026-08-19
feat(mocknvml): decode GPU fabric health for nvidia-smi -q

### LINKED ISSUE #677: [Feature]: decode GPU fabric health so nvidia-smi -q reports Summary, Bandwidth, and route state instead of N/A
## Problem statement

`nvidia-smi -q` reports NVLink fabric health as seven separate fields — an overall summary, whether bandwidth is degraded, and five specific fault conditions. On an NVL72 rack these are the first thing an operator or a fault-handling controller reads to decide whether a GPU's fabric attachment is usable. The mock reports the fabric's *identity* (state, clique, cluster UUID) but every health field is `N/A`, so a consumer cannot distinguish a healthy fabric from an unknown one, and degraded-fabric handling cannot be exercised at all.

The mock is one field away from being able to do this: `FabricInfo` already carries a `HealthSummary`, and the bridge already copies it into the v3 struct. Nothing ever sets it, and there is no config key to set it from — so it is always `0`, which `nvidia-smi` renders as `N/A`.

Found by diffing `nvidia-smi -q` from a real 4-GPU GB300 tray (driver `580.173.02`, CUDA 13.0) against the `gb300` profile running in kind, extending the audit behind #644 to Grace-Blackwell.

### Example output

`profiles/gb300.yaml` — identity is populated, health is not:

```
    Fabric
        State                             : Completed
        Status                            : Success
        CliqueId                          : 0
        ClusterUUID                       : 00000000-0000-0000-0000-000000000001
        Health
            Summary                       : N/A
            Bandwidth                     : N/A
            Route Recovery in progress    : N/A
            Route Unhealthy               : N/A
            Access Timeout Recovery       : N/A
            Incorrect Configuration       : N/A
```

A real GB300 in a healthy rack:

```
    Fabric
        State                                              : Completed
        Status                                             : Success
        CliqueId                                           : 32766
        ClusterUUID                                        : dc66362f-6973-449b-9299-8640b2d2badf
        Health
            Summary                                        : Healthy
            Bandwidth                                      : Full
            Route Recovery in progress                     : False
            Route Unhealthy                                : False
            Access Timeout Recovery                        : False
            Incorrect Configuration                        : None
            Partition Assigned                             : N/A
```

`Partition Assigned` is `N/A` on the reference tray as well, but note that the mock does not print the row at all — the driver never answers it, so `nvidia-smi` omits it. A healthy mock should print it as `N/A`, matching hardware, rather than dropping it.

### Root cause

`buildFabricInfo` in `pkg/gpu/mocknvml/engine/fabric.go` populates clique, state, cluster UUID, and health *mask*, but not the health *summary*:

```go
func buildFabricInfo(cfg *FabricConfig) FabricInfo {
    info := FabricInfo{
        CliqueID:   cfg.CliqueID,
        HealthMask: cfg.HealthMask,
        State:      resolveFabricState(cfg.State),
    }
    info.ClusterUUID = parseClusterUUID(cfg.ClusterUUID)
    return info
}
```

`FabricInfo` declares `HealthSummary uint8` and `pkg/gpu/mocknvml/bridge/fabric.go` assigns it into the v3 struct (`gpuFabricInfo.healthSummary = C.uchar(info.HealthSummary)`), so the plumbing is complete end to end — but `FabricConfig` has no `health_summary` key and `buildFabricInfo` never sets the field, so it is always zero. `nvidia-smi` treats a zero summary as "not reported" and prints `N/A` for the whole `Health` block, including the per-condition rows it would otherwise decode out of `healthMask`.

`HealthMask` is configurable (`fabric.health_mask`, default `0` = healthy) but has no effect on the rendered output today, which makes it the silently-dropped-configuration failure mode #644 calls out: a profile can set a health mask, the mock accepts it, and `nvidia-smi` reports `N/A`.

## Proposed solution

Make fabric health a first-class, configurable state and decode it:

- Add a `health_summary` key to `FabricConfig` and set it in `buildFabricInfo`, defaulting to the healthy value so existing profiles start reporting `Healthy` / `Full` instead of `N/A`.
- Express the health conditions symbolically in config — bandwidth degraded, route recovery in progress, route unhealthy, access timeout recovery, incorrect configuration — and derive `healthMask` from them, rather than asking a profile author to hand-assemble a bitmask. Keep raw `health_mask` accepted as an escape hatch.
- Verify the v2/v3 struct-tail handling stays correct once the summary is non-zero: the bridge comment notes v2 callers must not touch `healthSummary`, and that boundary is currently never exercised with a non-zero value.
- Report `Partition Assigned` as `N/A` rather than omitting the row.
- Extend runtime failure injection so fabric health can be degraded mid-test through `nvml-mock-ctl`, since a fabric that degrades under a running workload is the case this is worth simulating.

Alternatives considered: hardcoding `Healthy` for every profile would clear the `N/A` and cost almost nothing, but it would make degraded-fabric handling permanently untestable, which is the reason to do this at all — rejected.

## Scope / non-goals

In scope: the health summary, the per-condition decoding, config and injection surfaces, and the v2/v3 struct boundary under a non-zero summary.

Non-goals: fabric identity (state, clique, cluster UUID), which already works; IMEX channel behaviour; and multi-rack clique topology, tracked in #598.

## Audience

Internal NVIDIA teams.

## Acceptance criteria

- [ ] Fabric health summary is configurable, defaults to healthy, and `nvidia-smi -q` reports `Summary : Healthy` and `Bandwidth : Full` on the shipped Grace-Blackwell profiles rather than `N/A`.
- [ ] Each health condition — route recovery, route unhealthy, access timeout recovery, incorrect configuration, degraded bandwidth — can be set individually in config and is reflected in the matching `-q` row.
- [ ] `Partition Assigned` is printed as `N/A` rather than omitted.
- [ ] A raw `health_mask` still works, and setting it is no longer silently dropped.
- [ ] `nvmlDeviceGetGpuFabricInfo` (v1/v2) callers are unaffected by a non-zero health summary, with a unit test pinning the struct-tail boundary.
- [ ] Fabric health can be degraded at runtime via `nvml-mock-ctl` and restored, without restarting the workload.
- [ ] **Dedicated e2e test.** A new Ginkgo spec under `tests/e2e/go/` with `Label("nvidia-smi")` and `Label("failure-injection")` that: (1) asserts a healthy `gb300` deployment reports `Summary : Healthy` and `Bandwidth : Full`; (2) injects a specific degradation — route unhealthy, not merely "some" fault — and asserts only the matching row flips while the others stay `False`; and (3) clears it and asserts the summary returns to healthy. Asserting one condition flips while its neighbours do not is the point, since a mask applied wholesale would pass a weaker test. The test must fail against the current `main`.

giuliocalzo:
Fixes #677

## Summary

Every row of the `nvidia-smi -q` `Fabric` → `Health` block read `N/A`, so a consumer could not tell a healthy fabric from an unknown one, and degraded-fabric handling could not be exercised at all. `FabricInfo` already carried a `HealthSummary` that the bridge copied into the v3 struct, but nothing set it — and a zero summary makes `nvidia-smi` treat the whole block as unreported, which also made the existing `fabric.health_mask` key silently ineffective.

- **Health is configured and decoded.** A `fabric:` block with no health keys means a healthy fabric, so the shipped Grace-Blackwell profiles report `Summary : Healthy` / `Bandwidth : Full`. `fabric.health` faults one named condition at a time (`degraded_bandwidth`, `route_recovery`, `route_unhealthy`, `access_timeout_recovery`, `incorrect_configuration: <name>`), the `Summary` row is derived from the conditions unless pinned with `fabric.health_summary`, and a raw `fabric.health_mask` stays available as an escape hatch — an explicit `0` still means "the driver reported no health".
- **Runtime injection.** `nvml-mock-ctl fabric-health --gpu <idx> route_unhealthy` degrades a fabric under a running workload and `... fabric-health --gpu <idx> healthy` restores it without a `reset`, so the device's other overrides survive. It is authoritative like `throttle`: the named conditions are set and the rest cleared, so one condition flips its own row and only its own row. Injecting a fault also releases a profile-pinned summary, since a `Healthy` summary next to a faulted condition is not a state hardware can be in.
- **v1/v2 callers are unaffected** now that the summary is non-zero: `healthSummary` starts exactly where a v2 buffer ends, pinned by a unit test alongside the shared-field offsets the v2 branch's reinterpret cast depends on, plus a check that a v2 tag never takes the v3 path.
- **e2e assertions read the `-q -x` document** (following #658) rather than the human table. The fabric health elements are decoded in `schema.go`, and the checks compare a whole block per GPU, so one observation covers both the degraded device and its untouched neighbours — a fault applied to the wrong device fails rather than passes.

## Verification against a real `nvidia-smi`

Built the mock library from this branch, layered it onto the mock image (which bundles the real `nvidia-smi` 580.65.06) and ran the binary against the `gb200` profile config.

Before:

```
        Health
            Summary                       : N/A
            Bandwidth                     : N/A
            Route Recovery in progress    : N/A
            Route Unhealthy               : N/A
            Access Timeout Recovery       : N/A
            Incorrect Configuration       : N/A
```

After, with `nvml-mock-ctl fabric-health --gpu 1 route_unhealthy` applied — GPU 0 untouched, GPU 1 degraded, from the same `-q -x` document (this capture is the new `testdata/qx-gb200-fabric-degraded.xml` fixture):

```xml
<health>                                    <health>
  <summary>Healthy</summary>                  <summary>Unhealthy</summary>
  <bandwidth>Full</bandwidth>                 <bandwidth>Full</bandwidth>
  <route_recovery_in_progress>False</..>      <route_recovery_in_progress>False</..>
  <route_unhealthy>False</route_unhealthy>    <route_unhealthy>True</route_unhealthy>
  <access_timeout_recovery>False</..>         <access_timeout_recovery>False</..>
  <incorrect_configuration>None</..>          <incorrect_configuration>None</..>
</health>                                   </health>
```

`fabric-health healthy` restored it; `no_partition` rendered `<incorrect_configuration>No Partition</incorrect_configuration>` with `Summary : Unhealthy` on the target GPU only; `degraded_bandwidth` rendered `Bandwidth : Degraded` with `Summary : Limited Capacity`.

### Note on the `Partition Assigned` acceptance criterion

There is no row or element to assert. The string `Partition Assigned` does not exist in the 580.65.06 `nvidia-smi` the image bundles, and neither `PARTITION_ASSIGNED_TRUE` nor `_FALSE` in the health mask makes it appear; the reference tray prints it because it runs 580.173.02. The mock reports the field as `NOT_SUPPORTED`, which is what that healthy tray reports and what a newer `nvidia-smi` renders as `N/A`, and `schema.go` deliberately decodes no element whose name this package cannot pin against a captured document. Documented in `docs/configuration.md`.

## Test plan

- [x] `make test` — engine health resolution written TDD; mask field shifts/widths and summary constants pinned against `go-nvml`; `mockctl.FabricHealthPatch` and the `fabric-health` CLI covered; the XML check covered against both the new degraded fixture and the pre-#677 capture already in `testdata`, which it must reject
- [x] `pkg/gpu/mocknvml/bridge/fabric_tail_test.go` — struct-tail boundary, shared-field offsets, version dispatch
- [x] `make test-mocknvml-bridge` — C ABI harness exercises v1/v2/v3 against a healthy and a faulted device (51 passed)
- [x] `make lint-fix` — 0 lint issues (`govulncheck` reports pre-existing Go stdlib advisories; unmodified `main` fails identically)
- [x] `make helm-tests` — 169 tests, 15 snapshots (profile checksums refreshed)
- [x] Real `nvidia-smi` 580.65.06 against the built library, as above
- [ ] New Ginkgo spec in CI: healthy baseline per profile (`nvidiasmi.FabricHealth`), then `route_unhealthy` injected on one GPU with its neighbours asserted unchanged, cleared, and a named misconfiguration. Gated on `p.HasFabric()`; selected by `--label-filter="nvidia-smi"` or `"failure-injection"`.

## Docs

`docs/configuration.md` gains a Fabric Health section (conditions table, summary derivation, raw mask, the `Partition Assigned` caveat); `docs/nvml-mock-ctl.md` documents `fabric-health` with a worked example and its reset semantics; `CHANGELOG.md` entry under `[Unreleased]`.

### DIFF
--- a/pkg/gpu/mockctl/config_override.go
+++ b/pkg/gpu/mockctl/config_override.go
@@ -327,6 +327,83 @@ func SramECCPatch(count uint64, errorType, source string, thresholdExceeded bool
 	}, nil
 }
 
+// fabricHealthConditionKeys maps CLI-friendly fabric health condition names to
+// the fabric.health boolean they set. The canonical JSON keys are accepted
+// directly; the aliases cover how operators say them out loud.
+var fabricHealthConditionKeys = map[string]string{
+	"degraded_bandwidth":      "degraded_bandwidth",
+	"bandwidth":               "degraded_bandwidth",
+	"degraded_bw":             "degraded_bandwidth",
+	"route_recovery":          "route_recovery",
+	"route_unhealthy":         "route_unhealthy",
+	"access_timeout_recovery": "access_timeout_recovery",
+	"access_timeout":          "access_timeout_recovery",
+}
+
+// allFabricHealthConditionKeys is the full set of fabric.health booleans, used
+// to write an authoritative all-clear baseline so a FabricHealthPatch
+// represents exactly the requested conditions.
+var allFabricHealthConditionKeys = []string{
+	"degraded_bandwidth", "route_recovery", "route_unhealthy", "access_timeout_recovery",
+}
+
+// FabricHealthPatch builds a config override patch that sets a device's NVLink
+// fabric health conditions, so a fabric can degrade under a running workload
+// (#677). conditions are CLI-friendly names (see fabricHealthConditionKeys)
+// plus the incorrect-configuration values the engine accepts
+// (engine.FabricIncorrectConfigNames, e.g. "no_partition"); the special
+// condition "healthy" clears everything and cannot be combined with others.
+//
+// Like ThrottlePatch the block is authoritative: every condition is written
+// (requested ones faulted, the rest clear) so repeated invocations replace
+// rather than accumulate. It also releases fabric.health_summary to "auto",
+// because a profile that pins a summary would otherwise keep reporting it no
+// matter which condition is injected.
+func FabricHealthPatch(conditions []string) (map[string]any, error) {
+	if len(conditions) == 0 {
+		return nil, errors.New("fabric-health requires at least one condition (or 'healthy')")
+	}
+	health := map[string]any{"incorrect_configuration": "none"}
+	for _, k := range allFabricHealthConditionKeys {
+		health[k] = false
+	}
+	for _, c := range conditions {
+		key := strings.ToLower(strings.TrimSpace(c))
+		if key == "healthy" || key == "none" {
+			if len(conditions) != 1 {
+				return nil, errors.New("fabric-health 'healthy' cannot be combined with other conditions")
+			}
+			break
+		}
+		if canonical, ok := fabricHealthConditionKeys[key]; ok {
+			health[canonical] = true
+			continue
+		}
+		if misconfig, ok := fabricIncorrectConfigName(key); ok {
+			health["incorrect_configuration"] = misconfig
+			continue
+		}
+		return nil, fmt.Errorf("unknown fabric health condition %q", c)
+	}
+	return map[string]any{"fabric": map[string]any{
+		"health":         health,
+		"health_summary": "auto",
+	}}, nil
+}
+
+// fabricIncorrectConfigName reports whether name is one of the
+// incorrect-configuration values the engine decodes. "not_supported" is
+// excluded: it means "this driver does not answer the question", which is a
+// profile-level statement rather than something to inject at runtime.
+func fabricIncorrectConfigName(name string) (string, bool) {
+	for _, known := range engine.FabricIncorrectConfigNames() {
+		if known == name && known != "none" && known != "not_supported" {
+			return known, true
+		}
+	}
+	return "", false
+}
+
 // throttleReasonKeys maps CLI-friendly throttle reason names to the
 // clocks_throttle_reasons config field they enable. The canonical JSON keys are
 // accepted directly; short aliases cover the common thermal/power cases.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,25 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `nvmlDeviceGetMemoryErrorCounter` now honours its `locationType` and
   `counterType` arguments instead of returning the DRAM counter for every
   location. (#641)
+- mocknvml: GPU fabric health is now decoded and configurable, so
+  `nvidia-smi -q` reports `Summary : Healthy` / `Bandwidth : Full` on the
+  shipped Grace-Blackwell profiles instead of `N/A` for every row of the
+  `Fabric` → `Health` block. A `fabric:` block with no health keys means a
+  healthy fabric; individual conditions (`degraded_bandwidth`,
+  `route_recovery`, `route_unhealthy`, `access_timeout_recovery`,
+  `incorrect_configuration`) can be faulted under `fabric.health`, and the
+  `Summary` row is derived from them (degraded bandwidth alone reports
+  `Limited Capacity`) unless pinned with `fabric.health_summary`. The raw
+  `fabric.health_mask` stays available as an escape hatch and is no longer
+  silently dropped — it previously had no effect on the rendered output
+  because the health summary was always zero. Fabric health can also be
+  degraded and restored while a workload runs, with
+  `nvml-mock-ctl fabric-health --gpu <idx> route_unhealthy` and
+  `... fabric-health --gpu <idx> healthy`. `Partition Assigned` is reported as
+  NOT_SUPPORTED, matching hardware, which newer `nvidia-smi` builds render as
+  `N/A`. v1/v2 `nvmlDeviceGetGpuFabricInfo` callers are unaffected by the
+  now non-zero summary: the field lives past the end of the v2 struct, pinned
+  by a unit test on the struct-tail boundary. (#677)
 - mocknvml: configured `processes:` now surface in nvidia-smi — the default
   table's Processes box, `-q`, and `--query-compute-apps` all report the
   configured PIDs, names and GPU memory instead of always reporting none.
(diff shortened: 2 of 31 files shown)


pkg/gpu/mockctl/config_override.go
```diff
@@ -327,6 +327,83 @@ func SramECCPatch(count uint64, errorType, source string, thresholdExceeded bool
...
+		}
+		if misconfig, ok := fabricIncorrectConfigName(key); ok {
+			health["incorrect_configuration"] = misconfig
+			continue
+		}
+		return nil, fmt.Errorf("unknown fabric health condition %q", c)
+	}
+	return map[string]any{"fabric": map[string]any{
+		"health":         health,
+		"health_summary": "auto",
```
ArangoGutierrez, line 390: The patch releases a pinned health_summary to "auto" but leaves health_mask alone, and resolveFabricHealthMask returns a non-nil mask outright before ever looking at the health block. On a config that still pins health_mask the injection merges cleanly and is then discarded, so the command exits 0 having done nothing; with the health_mask: 0 that every shipped profile carried until this PR removed it, the derived summary comes back NotSupported and the whole block renders N/A, which is the pre-#677 behaviour this change is fixing. Worth writing health_mask as null alongside health_summary so the same reasoning that releases the pinned summary also releases the field that outranks it.

ArangoGutierrez (approved):
Nothing here blocks merge.

The mask encoding is the part I looked at hardest and it holds up: each condition gets its own shift/width slot, deriveFabricHealthSummary keeps degraded bandwidth at limited-capacity instead of promoting it to unhealthy, and mask == 0 still reports not-supported so the raw escape hatch keeps its old meaning. fabric_tail_test.go is the test I would have asked for - pinning healthSummary at exactly the end of a v2 struct, plus a v2 tag never classifying as v3, is what stops the now non-zero summary from writing past a v2 caller's buffer, and the shared-offset table covers the reinterpret cast the v2 branch depends on. On the e2e side, comparing a whole block per GPU out of one -q -x document with the neighbours pinned to healthy means a fault applied to the wrong device fails instead of quietly passing. One thing on the runtime injection path is worth picking up, inline - it does not change the shape of the change, so I am not holding the review for it.

PR #686, opened 2026-08-19
perf(e2e): cap the NRI workload grace period to unblock pod teardown

giuliocalzo:
## What This PR Does

1. Caps `terminationGracePeriodSeconds` on the NRI scenario's workload pods, which removes ~6min of pure dead time per `e2e-nri` leg.
2. Moves pod-manifest rendering into a new `tests/e2e/go/framework/pod` package built on the typed `corev1` API objects, replacing concatenated YAML text.

## Why

The `e2e-nri` leg ran ~12min, and **more than half of it was spent waiting**. Extracting timestamps from the [baseline run](https://github.com/NVIDIA/k8s-test-infra/actions/runs/32243013101/job/96039521833) (`e2e-nri (a100)`, 12m11s) shows cluster setup takes ~2min while `make e2e-nri` takes 10m09s — and inside that step there are 13 gaps of ~30s, each landing immediately after a pod delete:

```
   31s  AFTER:   pod "nri-dp-single" deleted from default namespace
   31s  AFTER:   pod "nri-dp-b" deleted from default namespace
   30s  AFTER:   pod "nri-dp-annotated-request" deleted from default namespace
   ... 13 total, 6.7min
```

The workload pods run `command: ["/bin/sh", "-c", "sleep 3600"]` and set no `terminationGracePeriodSeconds`, so they inherit Kubernetes' 30s default. `sleep` ends up as PID 1 and installs no SIGTERM handler, and **PID 1 never receives a signal it does not handle** — so kubelet's SIGTERM is discarded and the pod survives until the post-grace SIGKILL. `kubectl delete` waits, and this `Ordered` suite deletes a pod per spec, so every teardown paid the full 30s.

Measured directly on a Kind cluster with the same image and command:

| pod | delete duration |
| --- | --- |
| default grace period | 33s |
| `terminationGracePeriodSeconds: 1` | 3s |

The second change follows from the first. The manifests were assembled as YAML text, which puts indentation and quoting on every call site and fails quietly in both directions: a fragment spliced one level too deep is still valid YAML that the API server accepts with the setting silently dropped, and an unquoted `true` arrives as a bool where an annotation value must be a string. Rendering from typed objects makes field names a compile-time concern, and capping the grace period becomes the package default rather than something each call site must remember — which is what stops a future scenario from silently reintroducing the stall.

## Results

Measured in CI, same pipeline, all six GPU profiles:

| profile | before | after |
| --- | --- | --- |
| a100 | 12m11s | 5m49s |
| h100 | 11m47s | 7m02s |
| t4 | 11m50s | 5m48s |
| b200 | 12m39s | 5m57s |
| gb200 | 11m55s | 6m05s |
| gb300 | 12m49s | 6m28s |
| **total runner time** | **73min** | **37min** |

`e2e-nri` was the pipeline's critical path; `e2e-gpu-operator` now is, so end-to-end wall clock drops from ~12m49s to ~7m29s. The other legs were checked for the same pattern and are clean — `e2e`, `e2e-dra`, `e2e-gpu-operator` and `e2e-multi-node` have zero delete stalls, so their runtime is real work. The chart's own DaemonSets already default to a 2s grace period, so the NRI rollout waits are genuine readiness time.

## Verification

The refactor is proven behaviour-preserving rather than assumed. Every pod shape was rendered from both `main` and this branch, pushed through a live API server with `--dry-run=server` so all defaults were applied, and diffed. Each shape differs in exactly one meaningful field:

```
   "terminationGracePeriodSeconds": 30,
---
>  "terminationGracePeriodSeconds": 1,
```

The only other differences were per-request random values (`uid`, the projected `kube-api-access-*` volume name). That also independently confirms the 30s default was in effect before.

`framework/pod` carries its own tests in the existing `make test-e2e-framework` gate (no cluster needed), covering the default cap, the override, and that boolean-like annotation values stay quoted. The scenario keeps a table test for the shapes it composes on top.

No new dependencies: every import was already vendored, so `go.mod`, `go.sum` and `vendor/` are untouched — the constraint `framework/kube` documents.

## Checklist

- [x] Tests pass (`make test`, `make test-e2e-framework`; e2e legs green in CI)
- [x] Linter passes (`golangci-lint` 0 issues repo-wide, `go vet -tags e2e` clean; `make lint-fix` exits non-zero only on pre-existing stdlib govulncheck findings, reproduced identically on untouched `main`)
- [x] Documentation updated (`tests/e2e/README.md` scenario layout)
- [x] CHANGELOG.md updated


tests/e2e/go/scenario_nri_unit_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+	"github.com/NVIDIA/k8s-test-infra/tests/e2e/go/framework/kube"
+	"github.com/NVIDIA/k8s-test-infra/tests/e2e/go/framework/pod"
+)
+
+// The scenario's pod shapes are asserted through the API types, so the checks
+// hold whatever the renderer emits. framework/pod covers the rendering contract
+// itself; this covers the choices this scenario makes on top of it — placement,
+// the device and IMEX opt-ins, and the GPU request.
+func TestNRIPodManifests(t *testing.T) {
```
roma-glushko: I'm not sure we need these tests really. It's a part of the e2e test which itself a test so the manifest is already being tested along the way 😌 

More tests don't mean better if they are not meaningful 🤷

tests/e2e/go/framework/pod/pod.go
```diff
@@ -0,0 +1,107 @@
...
+// as PID 1, which installs no SIGTERM handler, and PID 1 never receives a
+// signal it does not handle — so kubelet's SIGTERM is discarded and the pod only
+// dies on the post-grace SIGKILL. At Kubernetes' 30s default that makes every
+// `kubectl delete` block for the full grace period, which is pure dead time in
+// a suite that deletes a pod per spec.
+const DefaultGracePeriodSeconds int64 = 1
+
+// Spec describes a single-container test pod. The zero value renders a valid
+// pod once Name and Image are set.
+type Spec struct {
```
roma-glushko, line 50: Can we simply add that test workload to local dir and ask tilt to deploy it as everything else? I don't want multiply ginkgo codebase unnecessary. All that redundant stuff would be something to read & maintain if we keep it.

tests/e2e/go/scenario_nri_test.go
```diff
@@ -840,116 +844,78 @@ func applyNRIWorkload(ctx context.Context, h *harness.Harness, manifest []byte,
...
+// nriWorkload is the shared shape of this scenario's workloads: a long-lived
+// container on the standard image, in the workload namespace. Callers vary
+// placement, annotations and the GPU request from there.
+func nriWorkload(name string) pod.Spec {
+	return pod.Spec{
+		Name:      name,
+		Namespace: nriWorkloadNS,
+		Image:     nriWorkloadImage,
+		// Kept alive so specs can exec into it.
+		Command: []string{"/bin/sh", "-c", "sleep 3600"},
```
roma-glushko: We should consider using busybox here instead of bare sleep. Busybox should react to sigterms.

ArangoGutierrez (commented):
Nothing here blocks merge.

The diagnosis holds up. PID 1 discards a signal it installs no handler for, so this is a real stall rather than a timeout that wanted tuning, and the package doc is right that busybox behaves identically for the same reason. I rendered every shape the five helpers produce and compared each against the string it replaces: they match field for field apart from the grace period, and an empty annotations map correctly renders no block rather than an empty one. Six e2e-nri legs at ~6min back the numbers.

One thing worth fixing in the description rather than the code: "What This PR Does" still says the package is built on typed corev1 objects, and "Verification" still credits pod_test.go plus a scenario table test, but 2d593513 replaced the first and deleted the second. That text becomes the squash commit message, so it is worth trimming before merge.

1. The comment calls this a cap, but `Render` only fills the field when it is zero, so a caller asking for 30 gets 30 and no caller can ask for 0. The `*int64` in e9f56f06 drew that distinction. Either clamp to the constant or call it a default, matching the field's own doc. (tests/e2e/go/framework/pod/pod.go:43)
2. `Env` and its template branch have no caller yet. Worth leaving out until one turns up, since a branch nothing renders is where an indentation slip would sit unnoticed. (tests/e2e/go/framework/pod/pod.go:65)
3. This reads as a structural guarantee, but scenario_dra_test.go:139 still hand-rolls a `sleep 300` pod with no grace period, so the pattern can come back through a path that never touches this package. Worth scoping the sentence to the NRI scenario. (CHANGELOG.md:76)

tests/e2e/go/framework/pod/pod.go
```diff
@@ -0,0 +1,130 @@
...
+// DefaultContainerName is the container's name when Spec leaves it unset.
+const DefaultContainerName = "app"
+
+// DefaultRestartPolicy suits a pod a spec either execs into or runs once; in
+// neither case should the kubelet restart it under the suite.
+const DefaultRestartPolicy = "Never"
+
+// DefaultGracePeriodSeconds is the cap applied unless a spec asks for longer.
+// See the package doc for why it is not left at the Kubernetes default.
+const DefaultGracePeriodSeconds = 1
```
ArangoGutierrez, line 43: Of the tests dropped in 2d593513, this is the one the e2e run does not cover for you. The image, command and placement assertions really were redundant, and a lost `| quote` fails loudly when the API server rejects the pod, but a regressed grace period is silent: I dropped the `terminationGracePeriodSeconds` line from pod.tpl.yaml and `make test-e2e-framework`, `go vet -tags e2e` and the suite compile all stayed green, so the 6.7min would just come back as a slower leg nobody is watching. A two-line render check on this constant would go red.

PR #687, opened 2026-08-20
feat(control-plane): Setup Control Plane Structured Logging

### LINKED ISSUE #618: MEP0001: Init structured logging in Control Plane
Init structured logs using the standard slog.

## Acceptance Criteria

- Use slog to emit structured logs
- Allow to have debug, info, warn, error log levels. 
- Allow to pass the minimal log level as a CLI flag
- Allow to setup plain or json log format

roma-glushko:
## What This PR Does

Setting up structured logging in control plane via slog package. Support different log levels.

Renamed mokka-control-plane to just control-plane to simply naming.

## Why

In order to be properly parsed by container log aggregators.

## Checklist

### DIFF
--- a/deployments/control-plane/Dockerfile
+++ b/deployments/control-plane/Dockerfile
@@ -28,10 +28,10 @@ COPY internal/ internal/
 COPY cmd/ cmd/
 RUN CGO_ENABLED=0 go build -mod=vendor \
     -ldflags="-s -w" \
-    -o /out/mokka-control-plane ./cmd/mokka-control-plane
+    -o /dist/control-plane ./cmd/control-plane
 
 FROM gcr.io/distroless/static-debian12:nonroot
-COPY --from=builder /out/mokka-control-plane /usr/local/bin/mokka-control-plane
+COPY --from=builder /dist/control-plane /usr/local/bin/control-plane
 USER nonroot:nonroot
 EXPOSE 8080
-ENTRYPOINT ["/usr/local/bin/mokka-control-plane"]
+ENTRYPOINT ["/usr/local/bin/control-plane"]
--- a/deployments/nvml-mock/helm/nvml-mock/values.yaml
+++ b/deployments/nvml-mock/helm/nvml-mock/values.yaml
@@ -381,7 +381,7 @@ controlPlane:
   enabled: false
   replicas: 1
   image:
-    repository: ghcr.io/nvidia/control-plane
+    repository: ghcr.io/nvidia/mokka-control-plane
     tag: latest
     pullPolicy: IfNotPresent
   logLevel: info                       # debug | info | warn | error
--- a/deployments/nvml-mock/helm/nvml-mock/values.yaml
+++ b/deployments/nvml-mock/helm/nvml-mock/values.yaml
@@ -381,7 +381,7 @@ controlPlane:
   enabled: false
   replicas: 1
   image:
-    repository: ghcr.io/nvidia/mokka-control-plane
+    repository: ghcr.io/nvidia/control-plane
     tag: latest
     pullPolicy: IfNotPresent
   logLevel: info                       # debug | info | warn | error
@@ -389,7 +389,7 @@ controlPlane:
     type: ClusterIP
     port: 8080
   # Total shutdown budget on the pod: preStop sleep + in-flight request drain
-  # (mokka-control-plane --shutdown-timeout, default 5s) must fit inside this.
+  # (control-plane --shutdown-timeout, default 5s) must fit inside this.
   # 30s matches the kube default and leaves headroom over the current 10s
   # preStop + 5s drain; raise this together with --shutdown-timeout if the
   # CP starts holding long-lived connections. Setting it below preStop + drain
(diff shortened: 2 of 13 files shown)


deployments/nvml-mock/helm/nvml-mock/values.yaml
```diff
@@ -381,15 +381,15 @@ controlPlane:
   enabled: false
   replicas: 1
   image:
-    repository: ghcr.io/nvidia/mokka-control-plane
+    repository: ghcr.io/nvidia/control-plane
```
giuliocalzo: @roma-glushko not sure about this... should be

```
ghcr.io/nvidia/k8s-test-infra/control-plane
```

or

```
ghcr.io/nvidia/mokka/control-plane
```

ArangoGutierrez (commented):
The slog wiring itself is clean: NewLogger's format switch mirrors the level switch, empty falls back to json to match DefaultConfig, and rejecting "text"/"logfmt" rather than silently accepting them is the right call. The mokka- prefix drop is carried through main.go, the chart template, the helm-unittest expectations and values.yaml consistently. What it misses is the Tilt dev flow, which is the only thing in the repo that builds this image.

1. The doc comment says a bad format is a startup error "so a typo in a Helm value fails loudly", but there is no Helm value for format yet: values.yaml exposes only controlPlane.logLevel and the deployment passes only --listen-addr and --log-level, so --log-format is reachable via CLI or MOKKA_CP_LOG_FORMAT and not from the chart. Either add controlPlane.logFormat alongside logLevel, or narrow the comment to level. (internal/controlplane/logging.go:14)

deployments/control-plane/Dockerfile
```diff
@@ -28,10 +28,10 @@ COPY internal/ internal/
 COPY cmd/ cmd/
 RUN CGO_ENABLED=0 go build -mod=vendor \
     -ldflags="-s -w" \
-    -o /out/mokka-control-plane ./cmd/mokka-control-plane
+    -o /dist/control-plane ./cmd/control-plane
```
ArangoGutierrez, line 31: The directory rename doesn't reach Tilt: local/nvml_mock.tiltfile:88 still passes dockerfile='deployments/mokka-control-plane/Dockerfile' and :96 watches the same dead prefix, so `tilt up -- --control-plane` fails at build_control_plane_image() on a path this PR deletes. Nothing in CI catches it because no workflow passes --control-plane and no workflow builds this image, so Tilt is the sole consumer. While you're in there, _CONTROL_PLANE_IMAGE at local/nvml_mock.tiltfile:27 and the image_deps entry at local/compute-domain/compute_domain.tiltfile:131 still say ghcr.io/nvidia/mokka-control-plane; image_keys substitution means those don't break the rollout, but they'd leave Tilt deploying the old name against a chart that now defaults to the new one.

internal/controlplane/config_test.go
```diff
@@ -47,3 +48,31 @@ func TestNewLogger(t *testing.T) {
...
+			cfg := controlplane.DefaultConfig()
+			cfg.LogFormat = tc.format
+			logger, err := controlplane.NewLogger(cfg)
+			if tc.wantErr {
+				require.Error(t, err)
+				require.Nil(t, logger)
+				return
+			}
+			require.NoError(t, err)
+			require.NotNil(t, logger)
```
ArangoGutierrez, line 77: The accept cases only check NoError and NotNil, and slog.New never returns nil, so nothing here distinguishes plain from json: point the "plain" branch in NewLogger at slog.NewJSONHandler and every case still passes. Since that branch is the new behavior in this PR, a type assertion on logger.Handler() (*slog.TextHandler vs *slog.JSONHandler) would make the table actually pin it.

deployments/nvml-mock/helm/nvml-mock/values.yaml
roma-glushko: The current image is called `ghcr.io/nvidia/nvml-mock` so I guess we should keep it as `ghcr.io/nvidia/mokka-control-plane` (applied https://github.com/NVIDIA/k8s-test-infra/pull/687/commits/46d9832a72fb174f61d59ac4eaae292f21616a3f).

giuliocalzo (dismissed):

deployments/control-plane/Dockerfile
roma-glushko: https://github.com/NVIDIA/k8s-test-infra/pull/687/changes/7f91d6919ee332143f7c56321c174fec29586c2d should be addressed in this PR ✅

internal/controlplane/config_test.go
roma-glushko: Added type checking in https://github.com/NVIDIA/k8s-test-infra/pull/687/commits/728d58afe8846bf21a9e08c26a6c6f1d17594c35 ✅ 

Thank you!

roma-glushko:
> The doc comment says a bad format is a startup error "so a typo in a Helm value fails loudly", but there is no Helm value for format yet: values.yaml exposes only controlPlane.logLevel and the deployment passes only --listen-addr and --log-level, so --log-format is reachable via CLI or MOKKA_CP_LOG_FORMAT and not from the chart. Either add controlPlane.logFormat alongside logLevel, or narrow the comment to level. (internal/controlplane/logging.go:14)

Yes, we should expose logFormat for sure. Addressed that in https://github.com/NVIDIA/k8s-test-infra/pull/687/commits/41a4449f9e7bbc33e3781fe2600a0f36405a6cd1 ✅

PR #690, opened 2026-08-20
feat(mokka): add informer-driven rack controller

### LINKED ISSUE #620: MEP0001: Introduce SGPUInventory Controller
Introduce SGPUInventory Controller so it watches for `SGPUInventory` resources that makes SGPU node allocations.

Example of the SGPUInventory resource:

```
apiVersion: mokka.nvidia.com/v1alpha1
kind: SGPUInventory
metadata:
  name: dev
spec:
  rackGroups:
    - id: training
      count: 2
      profileRef:
        name: gb300-nvl72

      placement:
        nodeSelector:
          matchLabels:
            mokka.nvidia.com/sgpu: training

    - id: ci
      count: 3
      profileRef:
        name: custom-vera-rubin-300
        
      placement:
        nodeSelector:
          matchLabels:
            mokka.nvidia.com/sgpu: ci

# information added by Control Plane
# This is useful for troubleshooting and understanding simulated capacity usage
status:
  capacity:
    racks: 5
    nodes: 48
    gpus: 168

  usage:
    requestedNodes: 40
    allocatedNodes: 40
    availableNodes: 8
    pendingNodes: 0

  rackGroups:
    - id: training
      profileName: gb300-nvl72
      
      capacity:
        racks: 2
        nodes: 36
        gpus: 144

      usage:
        requestedNodes: 32
        allocatedNodes: 32
        availableNodes: 4
        pendingNodes: 0

    - id: ci
      profileName: a100-2gpu-test-rack
      capacity:
        racks: 3
        nodes: 12
        gpus: 24

      usage:
        requestedNodes: 8
        allocatedNodes: 8
        availableNodes: 4
        pendingNodes: 0

  conditions:
    - type: Accepted
      status: "True"
      reason: Accepted
      message: The inventory configuration is valid.
      observedGeneration: 3
      lastTransitionTime: "2026-08-05T10:40:00Z"

    - type: ResolvedRefs
      status: "True"
      reason: ProfilesResolved
      message: All referenced SGPUProfiles were resolved.
      observedGeneration: 3
      lastTransitionTime: "2026-08-05T10:40:20Z"

    - type: Programmed
      status: "True"
      reason: InventoryPublished
      message: The inventory capacity was published to the allocation backend.
      observedGeneration: 3
      lastTransitionTime: "2026-08-05T10:40:40Z"

    - type: RequestsSatisfied
      status: "False"
      reason: InsufficientCapacity
      message: >-
        5 requested sGPU nodes in rack group "training" are waiting because
        all matching capacity is allocated.
      observedGeneration: 3
      lastTransitionTime: "2026-08-05T10:45:12Z"
```

## Acceptance Criteria

- Resolve SGPUProfile references in `SGPUInventory.rackGroups[]`
- For each `SGPUInventory` create sGPU node allocations in Redis that corresponds the real number of GPU nodes the specified rack can provide in real data centers
- The controller should publish status.capacity, status.usage, status.rackGroups updates.
- Persist sGPU rack/node runtime state in Redis (e.g. clique ID)

### LINKED ISSUE #622: MEP0001: Introduce SGPUProfile Controller
Introduce a controller that watches SGPUProfile objects and makes them available to the sGPU allocation logic.                                                                             

[MEP0001] defines SGPUProfile as the static ground truth for a GPU platform - rack shape, node GPU count, PCI topology, NVLink fabric, InfiniBand, software versions, and initial runtime defaults. The inventory controller (#620) references profiles by name via profileRef; this controller makes those references resolvable.      
                                              
## Acceptance Criteria

- Controller watches SGPUProfile cluster-wide resources using controller-runtime
- Validates the profile schema on reconcile:
   - spec.rack.nodesPerRack ≥ 1 (should be done via CRD built-in validators)
   - spec.node.topology.gpuSlots length equals spec.node.gpus.count and each index is unique 
- Sets a Ready condition on the profile status reflecting validation outcome
- Exposes an in-memory cache (indexed informer) so the SGPUInventory controller can resolve profileRef by name without extra API calls

tjrasche:
## What This PR Does

- adds an informer-driven controller to the existing Mokka control-plane process;
- materializes `SGPURackProfile` and `SGPUInventory` declarations into stable, controller-owned `SGPURack` resources;
- keeps allocation and status derivation in memory from informer caches, with Kubernetes objects as the source of truth;
- projects the required assignment annotation and labels onto bound Nodes using non-forced server-side apply;
- adds leader election, bounded API client settings, RBAC, Helm/Tilt wiring, examples, operator documentation, and an opt-in KWOK scale harness.

The controller stage intentionally excludes the agent/driver integration and runtime-policy evaluation.

## Why

This provides a small end-to-end control-plane slice for validating deterministic placement and Kubernetes API behavior before adding agent-side simulation. It avoids an additional shared state store: restarts rebuild derived state from Kubernetes, while durable rack bindings preserve stable identities.

This follows the direction discussed in #536 and targets clusters up to 100k Nodes while keeping API writes limited to materialized rack/status changes and the required per-Node projection.

## Testing

- `make test`
- `make lint-fix`
- `make helm-crds-tests`
- `make helm-tests`

## Checklist


internal/mokka/allocate/allocator.go
roma-glushko, line 1: I'm actively working on purging `pkg/*` content and move all unnecessary stuff to `internal/*`. We should have a good reason to add/keep things in pkg. Most stuff should live in the internal package.

enhancements/meps/0001-mokka-control-plane/controller-implementation-plan.md
roma-glushko, line 1: We are keeping the implementation plans in this repo. So make sense to keep this one private/out of git too. 

That said, I think MEP0001 should be updated to record the SGPURack update we discussed and implemented here.

internal/mokkacontroller/nodecatalog/catalog.go
roma-glushko, line 1: What's the purpose of internal/controlplane vs internal/mokkacontroller separation? internal/mokkacontroller would contain all the core logic and internal/controlplane service machinery? I would consider to keep it under internal/controlplane.

go.mod
```diff
@@ -50,24 +51,24 @@ require (
 	github.com/opencontainers/runtime-spec v1.3.0 // indirect
 	github.com/pmezard/go-difflib v1.0.1-0.20181226105442-5d4384ee4fb2 // indirect
 	github.com/sirupsen/logrus v1.9.4 // indirect
+	github.com/spf13/pflag v1.0.10 // indirect
```
roma-glushko, line 63: wondering what's importing cobra's pflag 👀

deployments/nvml-mock/helm/nvml-mock/templates/controlplane/admission.yaml
roma-glushko, line 1: Worth having a separate directory like templates/controlplane to group related resources.

cmd/control-plane/main.go
```diff
@@ -50,33 +47,82 @@ func newCLI() *cli.Command {
...
 func run(ctx context.Context, cmd *cli.Command) error {
 	cfg := controlplane.Config{
-		ListenAddr:      cmd.String("listen-addr"),
-		LogLevel:        cmd.String("log-level"),
-		ShutdownTimeout: cmd.Duration("shutdown-timeout"),
+		ListenAddr:              cmd.String("listen-addr"),
+		LogLevel:                cmd.String("log-level"),
+		ShutdownTimeout:         cmd.Duration("shutdown-timeout"),
+		Kubeconfig:              cmd.String("kubeconfig"),
+		LeaderElectionNamespace: cmd.String("leader-election-namespace"),
```
roma-glushko: We should probably think of adding nested config structs. Otherwise, the config would become bulky.

tests/kwok/README.md
roma-glushko, line 1: Thanks for adding kwok tests 👍 
Should be a great think long term. We could even try to run them periodically at some point.

roma-glushko (commented):
The PR is big enough for looking for a good stopping point and sending it to review.

Feel free to switch it to ready state once you are done. I will do the detailed dive into the actual logic once that happens. In the mean time, I have left a few high-level comments.

internal/mokka/allocate/allocator.go
tjrasche: Moved the complete Mokka domain subtree from `pkg/mokka` to `internal/mokka` in 1f75db8e. All repository imports now use the internal path, with no compatibility shim or remaining `pkg/mokka` references.

enhancements/meps/0001-mokka-control-plane/controller-implementation-plan.md
tjrasche: Removed the private controller implementation plan and updated MEP0001 to describe the implemented `SGPURack` and Kubernetes-authoritative assignment model in 7fa79362. The same documentation commit aligns the status example and CRD packaging text and removes the obsolete allocation diagram.

internal/mokkacontroller/nodecatalog/catalog.go
tjrasche: The split is intentional: `internal/controlplane` is the service shell (HTTP lifecycle, Kubernetes clients, leader election, and readiness), while `internal/mokkacontroller` is the independently testable informer/workqueue and reconciliation core. The dependency remains one-way from the service shell to the controller core, so I kept that boundary.

go.mod
tjrasche: `pflag` is indirect rather than a Cobra dependency in this command. `go mod why` resolves it as `internal/controlplane` → `k8s.io/client-go/tools/clientcmd` → `github.com/spf13/pflag`; there is no direct Cobra or pflag import.

deployments/nvml-mock/helm/nvml-mock/templates/controlplane/admission.yaml
tjrasche: Grouped the five resources under `templates/controlplane/` in 179bb1e4 and updated every helm-unittest template path. The full chart suite passes: 12 suites, 198 tests, and 15 snapshots.

cmd/control-plane/main.go
tjrasche: Refined this in 1f75db8e: `Config` now groups `Server`, `Kubernetes`, `LeaderElection`, and the existing controller `Options`. CLI flag names, environment variables, defaults, and runtime validation remain unchanged.

tests/kwok/README.md
tjrasche: Thanks! Agreed that periodic KWOK scale runs would be useful. I would track that as a follow-up once the harness and expected runtime have settled, rather than expanding this already-large PR.

deployments/nvml-mock/helm/nvml-mock/templates/controlplane/rbac.yaml
```diff
@@ -0,0 +1,86 @@
...
+    verbs: ["get", "list", "watch", "patch"]
+  - apiGroups: ["mokka.nvidia.com"]
+    resources: ["sgpurackprofiles"]
+    verbs: ["get", "list", "watch"]
+  - apiGroups: ["mokka.nvidia.com"]
+    resources: ["sgpuinventories"]
+    verbs: ["get", "list", "watch", "update"]
+  - apiGroups: ["mokka.nvidia.com"]
+    resources: ["sgpuracks"]
+    verbs: ["get", "list", "watch", "create", "patch", "delete"]
```
ArangoGutierrez: The reconciler writes racks only with Update, a PUT that RBAC maps to the update verb, so with create/patch/delete granted and update missing the controller can create a rack and then takes a 403 on every later write: converging spec changes, clearing node slots, and removing the mokka.nvidia.com/rack-cleanup finalizer, which leaves inventory deletion hanging. Granting update and dropping the unused patch covers what Mutations actually calls.

Also at deployments/nvml-mock/helm/nvml-mock/tests/controlplane_rbac_test.yaml:60.

internal/mokkacontroller/controller.go
```diff
@@ -0,0 +1,810 @@
...
+						var projectErr error
+						outcome, projectErr = projection.ProjectFresh(
+							ctx,
+							key.cleanup.RackName,
+							key.cleanup.Binding.Coordinate.NodeIndex,
+						)
+						return projectErr
+					}
+					if !allocation.RevisionCurrent(revision) {
+						continue
```
ArangoGutierrez: This continue retries without checking ctx, unlike the RevokeCleanup path a few lines below that does, and it spins while holding the per-inventory mutex, so sustained allocation-relevant Node churn can keep one inventory's reconciles blocked and the loop will not drain on shutdown. The same ctx.Err() check on this branch, plus a bounded number of retries, would let it give up cleanly.

tests/kwok/run.sh
```diff
@@ -0,0 +1,446 @@
...
+controller_ready() {
+	local port
+	port="$(cat "${WORK_DIR}/controller.port")"
+	curl --fail --silent --max-time 2 "http://127.0.0.1:${port}/readyz" >/dev/null
+}
+
+start_controller() {
+	[[ -z "${CONTROLLER_PID}" ]] || die "controller is already running"
+	local port
+	port="$((18081 + ($$ % 1000)))"
```
ArangoGutierrez: The listen port is derived from the shell PID, so two concurrent runs can collide on it, and controller_ready only curls /readyz without checking CONTROLLER_PID is still alive. A controller that exits during startup therefore burns the full KWOK_TIMEOUT_SECONDS, 1800s by default, before the harness reports anything. A kill -0 on the PID inside controller_ready turns that into an immediate failure.

ArangoGutierrez (commented):
Rebuilding derived state from Kubernetes on restart instead of adding another state store is the right trade here, and the non-forced apply with an explicit conflict outcome keeps the Node metadata honest about other writers.

The split between the service shell and the reconciliation core reads well, and the KWOK harness gives the 100k-node target something concrete to stand on. One thing in the chart blocks a real cluster deployment.

- client-gen, lister-gen and informer-gen are pinned at v0.36.3 while k8s.io/client-go, k8s.io/api and k8s.io/apimachinery sit at v0.37.0, and gen-check regenerates with the older tool on every make lint. Tracking the generator with the k8s minor keeps the checked-in clientset from drifting away from the library it compiles against. (hack/update-mokka-codegen.sh:16)

.github/workflows/mokka-control-plane-publish.yaml
```diff
@@ -0,0 +1,137 @@
...
+    runs-on: ubuntu-latest
+    permissions: {}
+
+    steps:
+      - name: Require an approved publishing ref
+        env:
+          PUBLISH_REF: ${{ github.ref }}
+          PUBLISH_REF_PROTECTED: ${{ github.ref_protected }}
+        run: |
+          if [[ "${PUBLISH_REF_PROTECTED}" != "true" ]]; then
```
ArangoGutierrez: A release tag cannot satisfy this gate. The repo has no tag ruleset and no legacy tag protection, so github.ref_protected is false for refs/tags/v*, which makes the semver branch below it unreachable and exits every tag push here before publish runs. Either add a tag ruleset covering v* or key the gate on the ref pattern instead of ref_protected.

hack/boilerplate.go.txt
```diff
@@ -0,0 +1,2 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright YEAR NVIDIA CORPORATION
```
ArangoGutierrez: gengo replaces YEAR with the year of the run, so the committed headers under pkg/generated read 2026 while make gen after January 1 emits 2027. gen-check byte-diffs ./pkg/generated and make lint runs it in CI, so every PR in the repo goes red on a date change with no source edit. Pinning the year or dropping it from the header decouples lint from the wall clock.

.github/workflows/mokka-control-plane-publish.yaml
```diff
@@ -0,0 +1,137 @@
...
+
+      - name: Publish
+        run: make mokka-control-plane-publish
+        env:
+          COSIGN_EXPERIMENTAL: "true"
+          MOKKA_CONTROL_PLANE_IMAGE: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          MOKKA_CONTROL_PLANE_TAGS: ${{ steps.meta.outputs.tags }}
+          MOKKA_CONTROL_PLANE_LABELS: ${{ steps.meta.outputs.labels }}
+          MOKKA_CONTROL_PLANE_CACHE_FROM: type=gha
+          MOKKA_CONTROL_PLANE_CACHE_TO: type=gha,mode=max
```
ArangoGutierrez: type=gha needs ACTIONS_RUNTIME_TOKEN and ACTIONS_CACHE_URL, which the docker/* actions inject into their own process. This build shells out to buildx from a plain run: step through make, so neither is set and the exporter has no credentials. The other image builds in the repo go through docker/build-push-action, which supplies them.

Makefile
```diff
@@ -153,6 +201,222 @@ build: ## Build all CLIs
...
+		echo "ERROR: MOKKA_CONTROL_PLANE_IMAGE, MOKKA_CONTROL_PLANE_TAGS, MOKKA_CONTROL_PLANE_CONTEXT, MOKKA_CONTROL_PLANE_DOCKERFILE, MOKKA_CONTROL_PLANE_PLATFORMS, MOKKA_CONTROL_PLANE_DIGEST_FILE, and MOKKA_CONTROL_PLANE_DOCKER are required." >&2; \
+		exit 2; \
+	fi; \
+	image_name="$${image##*/}"; \
+	if [[ "$$image" == *"@"* || "$$image_name" == *":"* || "$$image" =~ [[:space:]] ]]; then \
+		echo "ERROR: MOKKA_CONTROL_PLANE_IMAGE must be an untagged repository without whitespace: $$image" >&2; \
+		exit 2; \
+	fi; \
+	command -v "$$docker_cmd" >/dev/null 2>&1 || { echo "ERROR: required buildx frontend not found: $$docker_cmd" >&2; exit 2; }; \
+	build_args=(buildx build --file "$$dockerfile" --platform "$$platforms" --push); \
```
ArangoGutierrez: This build never forwards GOLANG_VERSION, so the released image takes the Dockerfile default of 1.26.6 while hack/golang-version.sh resolves 1.26.7. Adding --build-arg GOLANG_VERSION keeps the published artifact on the same patched toolchain every other image build in the repo uses.

tests/kwok/run.sh
```diff
@@ -0,0 +1,446 @@
...
+check_state controller-restart "${scenario_started_seconds}" "${FULL_RACKS}" "${NODE_COUNT}" "${NODE_COUNT}" "${NODE_COUNT}" true
+snapshot_assignments "${ARTIFACT_DIR}/restart.after.json"
+cmp "${ARTIFACT_DIR}/restart.before.json" "${ARTIFACT_DIR}/restart.after.json" || die "assignments changed across controller restart"
+
+readonly REPLACED_NODE="mokka-node-000000"
+readonly OLD_UID="$(kctl get node "${REPLACED_NODE}" -o jsonpath='{.metadata.uid}')"
+scenario_started_seconds="$(date -u +%s)"
+kctl delete node "${REPLACED_NODE}" --wait=true >/dev/null
+scale_nodes "${NODE_COUNT}"
+readonly NEW_UID="$(kctl get node "${REPLACED_NODE}" -o jsonpath='{.metadata.uid}')"
```
ArangoGutierrez: readonly reports its own exit status, so a failing kctl get here leaves NEW_UID empty without tripping set -e, and the OLD_UID != NEW_UID check on the next line then passes on that empty value and reports a successful replacement. Splitting the declaration from the assignment, or asserting NEW_UID is non-empty, makes the failure surface. The same masking applies to OLD_UID four lines up.

internal/mokka/materialize/materialize_test.go
```diff
@@ -0,0 +1,461 @@
...
+		{name: "invalid profile", mutate: func(in *RackInput) { in.Profile.Spec.Node.Topology.GPUSlots = nil }},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			input := base
+			input.Profile = base.Profile.DeepCopy()
+			tt.mutate(&input)
+			_, err := RenderRack(input)
+			require.Error(t, err)
```
ArangoGutierrez: All seven cases assert only that some error came back, so any single early return in RenderRack keeps the whole table green no matter which guard actually fired. The other negative tests in this file pin the message with EqualError or ErrorContains, and matching that here would let the table catch a guard that stops discriminating.

ArangoGutierrez (commented):
The restart-safety work landed well, and compacting the Node informer down to the three projection keys is a neat way to carry 100k Nodes without standing up a second store.

Two release-path problems fail on a schedule rather than on a code change, both inline. The rest are test-side and can ride along.

- This retry still has no ctx.Err() check, unlike the RevokeCleanup path just below it, and it spins while holding the per-inventory lock, so sustained allocation churn can starve that inventory and the loop will not drain on shutdown. (internal/mokkacontroller/controller.go:380)
- The expected config here is controlplane.DefaultConfig() while the actual value comes from flag defaults initialized from that same function, so this case cannot fail on a wrong default. The test also overrides cmd.Action and rebuilds the flag-to-Config mapping itself, which leaves the real mapping in run(), including the nine new leader-election and client-tuning flags, unexercised. (cmd/control-plane/main_test.go:68)
- TestRunnerContract pins run.sh by substring, so renaming KUBECONFIG_PATH or reflowing a flag turns it red while gutting the body of rack_statuses_ready leaves it green. The NotContains guards against curl piped to sh and helm install are worth keeping; the positive text assertions are the part that tracks formatting rather than behavior. (tests/kwok/contract_test.go:63)

deployments/nvml-mock/helm/nvml-mock/templates/controlplane/rbac.yaml
tjrasche: Addressed in `febfcb7b`: the SGPURack RBAC rule now grants `update` and drops the unused `patch` verb, and the Helm unit expectation was updated with it. `make helm-tests` passes (207 tests).

hack/boilerplate.go.txt
tjrasche: Addressed in `f80a0e90`: the generator boilerplate now pins the copyright year to 2026 instead of using gengo’s wall-clock `YEAR` substitution. A fresh codegen run leaves the worktree unchanged.

.github/workflows/mokka-control-plane-publish.yaml
tjrasche: Addressed in `57d0d2c5`: removed the `type=gha` cache settings from this plain `run:`/Makefile build path. The generic Makefile cache inputs remain available to callers that can supply authenticated cache settings. `actionlint` passes.

Makefile
tjrasche: Addressed in `b6fe7d70`: the workflow passes `steps.golang.outputs.version`, the Makefile requires and forwards it as `--build-arg GOLANG_VERSION=...`, and the Dockerfile fallback is aligned to 1.26.7. The publication tests cover the exact build argument and reject an empty version.

tests/kwok/run.sh
tjrasche: Addressed in `1f5f64d3`: both `OLD_UID` and `NEW_UID` are assigned before their separate `readonly` declarations, so a failed `kctl get` propagates through `set -e`. Contract coverage now guards both lookups; Bash syntax and focused tests pass.

internal/mokka/materialize/materialize_test.go
tjrasche: Addressed in `c0337581`: all seven table cases now carry a stable expected substring and assert it with `require.ErrorContains`, so each case verifies its intended validation branch. The materialize package tests pass.

tjrasche:
Addressed the two review-summary findings that were not represented by resolvable inline threads:

- `70031c86` aligns client-gen, lister-gen, and informer-gen on Kubernetes `v0.37.0`; deterministic regeneration and generated-package tests pass.
- `b9c6bfc9` extracts the production `configFrom(*cli.Command)` mapping used by `run()` and tests defaults plus distinct overrides for every current config field. Focused control-plane tests pass.

I left the cleanup retry/locking, KWOK startup port/liveness, tag-protection gate, and broader KWOK behavioral-test findings open; they are not part of these fixes.

cmd/control-plane/main.go
```diff
@@ -16,8 +13,10 @@ import (
 
 	"github.com/NVIDIA/k8s-test-infra/internal/controlplane"
 	"github.com/NVIDIA/k8s-test-infra/internal/logging"
+	"github.com/NVIDIA/k8s-test-infra/internal/mokkacontroller"
 	"github.com/urfave/cli/v3"
 	"golang.org/x/sync/errgroup"
+	"k8s.io/klog/v2"
```
roma-glushko, line 19: we have standardize on zap for logging. No reason to pull in klog.

cmd/control-plane/main.go
```diff
@@ -72,22 +140,23 @@ func run(ctx context.Context, cmd *cli.Command) error {
...
-		ListenAddr:      cmd.String("listen-addr"),
-		ShutdownTimeout: cmd.Duration("shutdown-timeout"),
+	cfg := configFrom(cmd)
+	controller, err := controlplane.NewController(cfg)
+	if err != nil {
+		return err
 	}
 
 	signalCtx, cancel := signal.NotifyContext(ctx, syscall.SIGINT, syscall.SIGTERM)
 	defer cancel()
```
roma-glushko, line 153: I have separated a common health machinery into internal/health package. Can we please use it here for consistency?

internal/controlplane/controller.go
```diff
@@ -0,0 +1,375 @@
...
+	return &Controller{
+		config: config, kubeClient: kubeClient, reconciler: reconciler,
+		readiness: newElectionReadiness(),
+	}, nil
+}
+
+// ValidateControllerConfig rejects settings that cannot make progress safely.
+//
+//nolint:cyclop // Each branch reports a distinct unsafe controller setting.
+func ValidateControllerConfig(config Config) error {
```
roma-glushko, line 66: Does it belong to Config as validate() method similar to controller Option?

internal/mokkacontroller/controller.go
tjrasche: Addressed in `cb9dbf08`: allocation-aware projection cleanup is now limited to five revision attempts while holding the inventory lock. Each attempt checks cancellation, and exhaustion returns a retryable error so the lock is released and the workqueue can rate-limit/requeue the item. The stale-cleanup acknowledgement revocation and non-current-binding fallback remain intact.

Added focused coverage for cancellation on an unsettled revision and bounded exhaustion/requeue. `GOTOOLCHAIN=go1.26.7 make lint-fix` and `GOTOOLCHAIN=go1.26.7 make test` pass.

tests/kwok/run.sh
tjrasche: Addressed in `cb9dbf08`: `controller_ready` now checks `CONTROLLER_PID` and exits immediately with the recent controller log when the process has died, rather than letting `wait_for` consume the full timeout. Controller ports are now reserved atomically with per-port lock directories under the existing artifact lock root, retained across restart, and released during cleanup.

Added behavioral subprocess tests for immediate dead-controller failure and distinct concurrent port reservations. The race-enabled KWOK tests, full `make test`, and `bash -n tests/kwok/run.sh` pass.

.github/workflows/mokka-control-plane-publish.yaml
tjrasche: Addressed in `cb9dbf08`: the unreachable `github.ref_protected` gate was removed. Validation now uses an exact ref allowlist: `refs/heads/main` or a stable `refs/tags/vMAJOR.MINOR.PATCH` tag whose numeric components have no leading zeroes. Other branches, partial versions, prereleases, and build-metadata tags are rejected.

The policy lives in `hack/validate-mokka-control-plane-publish-ref.sh` with table-driven accepted/rejected ref coverage. `actionlint`, `make lint-fix`, and `make test` pass. Without a `v*` ruleset, permission to create tags remains the release trust boundary.

deployments/mokka-crds/helm/mokka-crds/templates/mokka.nvidia.com_sgpuinventories.yaml
```diff
@@ -58,12 +59,16 @@ spec:
...
+                description: |-
+                  RackGroups declares homogeneous sets of racks. One group can expand to
+                  many racks; the controller admits at most 64 groups across all Inventories.
                 items:
                   description: RackGroup is a homogeneous group of racks sharing a
                     profile.
                   properties:
                     count:
                       format: int32
+                      maximum: 100000
```
roma-glushko, line 71: Very generous. I thought of keeping it eq to 64 like the number of listeners in the gateway api implementations.

deployments/nvml-mock/helm/nvml-mock/templates/controlplane/deployment.yaml
```diff
@@ -10,6 +18,10 @@ metadata:
     app.kubernetes.io/component: control-plane
 spec:
   replicas: {{ .Values.controlPlane.replicas }}
+  # Leader election coordinates active/passive replicas within a generation;
+  # Recreate keeps whole-generation replacement deterministic.
+  strategy:
```
roma-glushko, line 23: Control plane is a pure controller. It's going to have REST API for mokka node daemons. The recreate strategy can imply downtime for those API. The Node Daemons will cache the last response in case of issues with REST API, but I'm not sure we should keep it as a common case.

enhancements/meps/0001-mokka-control-plane/README.md
```diff
@@ -87,42 +87,65 @@ The sGPU racks are characterized by:
...
+binding or rack until that cleanup is acknowledged. Missing Nodes and same-name
+replacements are treated as the exact old UID being absent. Controller-owned
+racks use an inventory controller owner reference with
+`blockOwnerDeletion: false`; finalizers, rather than a blocking owner reference,
+coordinate projection cleanup. Racks owned by another object are not adopted.
 
 #### Runtime State
 
-While the high-level architecture is the same, there are two different ways to manage runtime state.
+**Future design.** Runtime-policy evaluation, node-agent heartbeats, simulated
```
roma-glushko, line 138: MEP is a final vision for the implementation, not an implementation log. It doesn't make sense to talk about the current stage of implementation what's ready and what's not. It should summarize where we want to get ultimately.

enhancements/meps/0001-mokka-control-plane/README.md
```diff
@@ -87,42 +87,65 @@ The sGPU racks are characterized by:
...
+telemetry, and fault injection are not implemented by the current controller
```
roma-glushko, line 139: I would also move the previous considerations to the appendix of the spec, so future readers can access them.

hack/update-mokka-codegen.sh
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
```
roma-glushko, line 1: Should we turn it into a golang CLI that is triggered by make gen command?

internal/controlplane/api/v1alpha1/crd_retention_test.go
```diff
@@ -0,0 +1,34 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	apixv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"sigs.k8s.io/yaml"
+)
+
+func TestGeneratedCRDsAreRetainedOnHelmUninstall(t *testing.T) {
```
roma-glushko, line 16: Is not this something helm unit tests are supposed to validate? Not sure we need a golang test here

deployments/nvml-mock/helm/nvml-mock/templates/_helpers.tpl
```diff
@@ -87,7 +87,18 @@ fullname so its Deployment/Service cannot be adopted by the DaemonSet's
...
+{{- printf "%s%s" ((include "nvml-mock.fullname" .) | trunc $baseLength | trimSuffix "-") $suffix }}
+{{- end }}
+
+{{/*
+The control plane mutates cluster-scoped state, so its required cluster-scoped
+RBAC resources use a fixed identity. The ClusterRoleBinding also serves as the
+Helm-owned singleton guard.
+*/}}
+{{- define "nvml-mock.controlPlaneSingletonName" -}}
+mokka-control-plane.mokka.nvidia.com
```
roma-glushko, line 101: `control-plane.mokka.nvidia.com`

deployments/nvml-mock/helm/nvml-mock/values.yaml
```diff
@@ -349,22 +349,33 @@ integrations:
...
     format: json   # json | plain
   # Go duration string: 5s, 1m30s, etc.
   shutdownTimeout: 5s
+  workers: 2
+  # Client-side Kubernetes API throttling. Raise these together with workers
+  # when faster initial convergence is worth the additional control-plane load.
+  kubeAPIQPS: 50
+  kubeAPIBurst: 100
+  leaderElection:
+    name: mokka-control-plane.mokka.nvidia.com
```
roma-glushko, line 378: `control-plane.mokka.nvidia.com`

deployments/nvml-mock/helm/nvml-mock/values.yaml
```diff
@@ -381,6 +392,6 @@ controlPlane:
       memory: 64Mi
     limits:
       cpu: 500m
-      memory: 256Mi
+      memory: 1Gi
```
roma-glushko, line 395: Was this changed based on 100k kwok testing?

internal/mokka/allocate/allocator.go
```diff
@@ -0,0 +1,400 @@
...
+	InventoryUID  types.UID
+	RackGroup     string
+}
+
+func (key GroupKey) String() string {
+	return key.InventoryName + "/" + key.RackGroup
+}
+
+// Group is a resolved placement and capacity declaration.
+type Group struct {
```
roma-glushko, line 32: Does it make sense to call them RackGroup and RackGroupKey?

internal/mokka/allocate/allocator.go
```diff
@@ -0,0 +1,400 @@
...
+	Key          GroupKey
+	Selector     *metav1.LabelSelector
+	Racks        int32
+	NodesPerRack int32
+}
+
+// Node is the allocation-relevant portion of a Kubernetes Node.
+type Node struct {
```
roma-glushko, line 40: Does it make sense to call it K8sNode?

internal/mokka/allocate/allocator.go
```diff
@@ -0,0 +1,400 @@
...
+}
+
+// NodeReference pins a binding to an exact Node instance.
+type NodeReference struct {
+	Name string
+	UID  types.UID
+}
+
+// Coordinate identifies one durable logical Node in a rack.
+type Coordinate struct {
```
roma-glushko, line 55: Does it make sense to call it RackNode?

internal/mokka/allocate/allocator.go
```diff
@@ -0,0 +1,400 @@
...
+	Group     GroupKey
+	RackIndex int32
+	NodeIndex int32
+}
+
+// Binding assigns an exact Node instance to one coordinate.
+type Binding struct {
```
roma-glushko, line 62: Does it make sense to call it RackNodeBinding?

internal/mokka/allocate/allocator.go
```diff
@@ -0,0 +1,400 @@
...
+
+// Stats exposes bounded-work counters for large-snapshot tests and benchmarks.
+type Stats struct {
+	SelectorEvaluations int64
+	BindingLookups      int64
+	NodeVisits          int64
+}
+
+// Input is a complete immutable allocation snapshot.
+type Input struct {
```
roma-glushko, line 112: Input is way too generic. Is Allocations or AllocationRequest fitting here?

internal/mokka/allocate/allocator.go
```diff
@@ -0,0 +1,400 @@
...
+}
+
+func indexNodes(nodes []Node) (map[types.UID]Node, error) {
+	indexed := make(map[types.UID]Node, len(nodes))
+	for _, node := range nodes {
+		if node.Name == "" || node.UID == "" {
+			return nil, errors.New("node identity must include name and UID")
+		}
+		if _, exists := indexed[node.UID]; exists {
+			return nil, fmt.Errorf("duplicate node UID %q", node.UID)
```
roma-glushko, line 301: do we have to fail in this case?

internal/mokka/allocate/allocator.go
```diff
@@ -0,0 +1,400 @@
...
+	selector labels.Selector
+	occupied map[uint64]struct{}
+	cursor   int64
+}
+
+// Allocate produces a deterministic plan without compacting valid bindings or
+// reusing coordinates whose cleanup is still pending.
+//
+//nolint:cyclop // One deterministic pass records each retention, release, conflict, and assignment decision.
+func Allocate(input Input) (Plan, error) {
```
roma-glushko, line 144: Does it really produce a plan or it's more like a State? Plan usually implies that it's meant to be implemented or act upon. If we just rebuild the state of the world, then it's not a plan to me.

internal/mokka/allocate/selectors.go
```diff
@@ -0,0 +1,134 @@
...
+// CompileGroups validates group identity, capacity, and selectors once.
+func CompileGroups(groups []Group) (CompiledGroups, error) {
+	compiled := make([]compiledGroup, 0, len(groups))
+	seen := make(map[GroupKey]struct{}, len(groups))
+	for _, group := range groups {
+		if err := validateGroup(group); err != nil {
+			return CompiledGroups{}, err
+		}
+		if _, exists := seen[group.Key]; exists {
+			return CompiledGroups{}, fmt.Errorf("duplicate group %s", group.Key)
```
roma-glushko, line 46: The same question whether it's worth tolerating this case.

internal/mokka/allocate/selectors.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	compiled, err := metav1.LabelSelectorAsSelector(selector)
+	if err != nil {
+		return nil, err
+	}
+	return compiled, nil
+}
+
+// Classify evaluates every eligible Node against precompiled group selectors.
+func Classify(nodes []Node, groups CompiledGroups) ([]Classification, Stats) {
```
roma-glushko, line 92: I think kubernetes scheduler use term ranking for this process of finding candidates. We should use that as well.

internal/mokka/allocate/allocator.go
```diff
@@ -0,0 +1,400 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright 2026 NVIDIA CORPORATION
+
+// Package allocate plans stable Kubernetes-Node-to-logical-Node bindings from immutable snapshots.
+// It performs no Kubernetes API operations and does not mutate its inputs.
+package allocate
```
roma-glushko, line 6: I assume that `internal/mokka` is supposed to be the domain layer here? Any reason not to call it `internal/sgpu`?

internal/mokkacontroller/nodecatalog/catalog.go
```diff
@@ -0,0 +1,284 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright 2026 NVIDIA CORPORATION
+
+// Package nodecatalog maintains the controller's rebuildable, allocation-
+// oriented view of eligible Kubernetes Nodes.
+package nodecatalog
```
roma-glushko, line 6: Looks like this may belong to internal/mokka package.

ArangoGutierrez (commented):
Thanks for this, rebuilding the rack bindings straight out of the informer caches on restart is what makes the lifecycle usable without a second state store.

Stage 1 of the Mokka control plane: SGPUInventory plus SGPURackProfile materialize into controller-owned SGPURack objects with the bindings projected onto Nodes, behind an opt-in controlPlane.enabled chart flag, finalizer-gated cleanup and a digest-pinned, cosign-signed image. The RBAC verb set, the publish ref allowlist, the bounded cleanup retry and the KWOK harness guards raised earlier all hold at this head. One exit condition in the per-rack loop needs sorting out before this goes in.

- The v* tag trigger admits prerelease tags that validate-mokka-control-plane-publish-ref.sh then rejects, so pushing a tag like v0.3.0-rc1, which already exists on this repo, starts this workflow and ends it red instead of skipping it. Emitting a boolean output from the validate job and gating publish on it turns an unsupported ref into a skipped job rather than a failed run. (.github/workflows/mokka-control-plane-publish.yaml:16)

internal/mokkacontroller/rack/controller.go
```diff
@@ -0,0 +1,1564 @@
...
+				return result, nil
+			}
+			continue
+		}
+		for rackIndex := int32(0); rackIndex < group.group.Count; rackIndex++ {
+			if err := reconcileRack(rackIndex); err != nil {
+				sortResult(&result)
+				return result, err
+			}
+			if !result.ResolvedRefs || capacityGrowthBlocked {
```
ArangoGutierrez, line 577: reconcile sets result.ResolvedRefs from the whole inventory's profile issues before the rack loop starts, so this break fires after rack index 0 of every healthy group as soon as any sibling group has an unresolved profileRef. An inventory with one broken group alongside a healthy count 3 group materializes only rack 0, and because reconcile returns a nil error the key is forgotten and the shortfall never self-heals. Track the per-rack materialization failure in a loop-local flag set inside reconcileRack instead of reusing the inventory-wide ResolvedRefs, and the requestedGroup loop just above needs the same treatment.

internal/controlplane/controller.go
```diff
@@ -0,0 +1,375 @@
...
+	run func(context.Context),
+	readiness *electionReadiness,
+) leaderelection.LeaderElectionConfig {
+	onStartedLeading := run
+	onStoppedLeading := func() {}
+	var onNewLeader func(string)
+	if readiness != nil {
+		onStartedLeading = func(ctx context.Context) {
+			readiness.startLeading()
+			run(ctx)
```
roma-glushko, line 161: So standby replices have no information about assignment whatsoever? How would they participate in serving computed state information back to node agents? I'm fine with one single active replica of controller, but I'm not fine with only one replica that can reply to all node agent requests.

roma-glushko (dismissed):
The changeset raises a few question for me:

- How control plane is supposed to serve node agents via REST API if only leader has compiled state of the world. We have agreed node agent is not going to recalculate its state, the state is going to be given.
- if you look at the code organization it's based around technical concepts rather than our domain terms. That's fine for K8s plumbing. But I would seek for a way to organize everything else around sgpu concepts. After all, when I look for how inventory is handled I don't necessary think in terms of projection/metadata/informers. I look for inventory. So resurfacing that would be good.
- If you zoom out, the whole new codebase is basically behaviourless/anemic data models + a bag of loosly connected stateless functions. I doubt all those functions are really unconnected. Some of them are even named like VerbNounSomething which implies that they could be grouped together as one component. It's beneficial to be able to think about the system as interaction between few components are those higher level abstractions help to simply code understanding. 
- We may have duplicated functions simply because of those anemic data models:

```go
// mokkacontroller/controller.go:678   matches index AND requires NodeRef != nil
func boundRackSlot(rack *SGPURack, index int32) *SGPURackNode

// projection/controller.go:914        matches index
func findSlot(rack *SGPURack, index int32) *SGPURackNode

// rack/projection.go:222              matches index; identical body to findSlot
func boundSlot(rack *SGPURack, index int32) *SGPURackNode
```

This is fine to add methods to the API data models so we don't have to duplicate simple logic around data they hold.

cmd/control-plane/main.go
tjrasche: Addressed. The control-plane server now uses `internal/health.Handler` for both `/healthz` and `/readyz`, with controller readiness adapted to the shared probe response.

deployments/mokka-crds/helm/mokka-crds/templates/mokka.nvidia.com_sgpuinventories.yaml
tjrasche: Addressed. `SGPUInventory.spec.rackGroups` now has `MaxItems=64`; the generated CRD has `maxItems: 64`, and controller admission enforces the same 64-group aggregate bound.

internal/sgpu/inventory/allocate/selectors.go
```diff
@@ -0,0 +1,134 @@
...
+// CompileGroups validates group identity, capacity, and selectors once.
+func CompileGroups(groups []Group) (CompiledGroups, error) {
+	compiled := make([]compiledGroup, 0, len(groups))
+	seen := make(map[GroupKey]struct{}, len(groups))
+	for _, group := range groups {
+		if err := validateGroup(group); err != nil {
+			return CompiledGroups{}, err
+		}
+		if _, exists := seen[group.Key]; exists {
+			return CompiledGroups{}, fmt.Errorf("duplicate group %s", group.Key)
```
tjrasche: Addressed. Selector compilation errors are now returned from `CompileGroups`/`CompilePlacementSelector` rather than skipped. Inventory reconciliation surfaces malformed selectors as validation failures and preserves last-good state.

internal/sgpu/inventory/allocate/allocator.go
```diff
@@ -0,0 +1,400 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright 2026 NVIDIA CORPORATION
+
+// Package allocate plans stable Kubernetes-Node-to-logical-Node bindings from immutable snapshots.
+// It performs no Kubernetes API operations and does not mutate its inputs.
+package allocate
```
tjrasche: Addressed in 5e0ef2e. The inventory domain now lives under `internal/sgpu/inventory` (including allocation, metadata, rack rendering, projection, and status), while `internal/mokkacontroller` retains controller composition.

internal/sgpu/inventory/nodecatalog/catalog.go
```diff
@@ -0,0 +1,284 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: Copyright 2026 NVIDIA CORPORATION
+
+// Package nodecatalog maintains the controller's rebuildable, allocation-
+// oriented view of eligible Kubernetes Nodes.
+package nodecatalog
```
tjrasche: Addressed in 5e0ef2e. The Node catalog moved from `internal/mokkacontroller/nodecatalog` to `internal/sgpu/inventory/nodecatalog`, so it now belongs to the inventory domain while controller composition remains in `internal/mokkacontroller`.

cmd/control-plane/main.go
tjrasche: Addressed. The control-plane entrypoint now uses the shared `internal/logging` Zap logger directly and no longer imports or configures klog.

internal/controlplane/controller.go
tjrasche: Addressed in 95ce8a74. Every control-plane replica now starts and synchronizes the profile, inventory, rack, and eligible-Node informer caches before entering leader election, so standbys maintain the same informer-backed view of the authoritative `SGPURack.spec.nodes[].nodeRef` assignments.

Only the elected leader attaches the existing mutation-routing handlers and starts reconciliation queues/workers. Standbys therefore keep the read state current without enqueueing work or writing Kubernetes resources. On promotion, the leader attaches handlers to the warm informer stores and waits for their initial replay to complete before starting workers, so events observed while standby are not lost. Leader-handler shutdown and worker drain still complete before Lease release, independently of the process-lifetime read cache.

Readiness now requires synchronized caches on every replica; an elected leader additionally waits for handler replay and worker startup. I also added an internal exact name+UID `AssignmentsForNode` lookup over the existing rack index for the future serving layer, without adding a REST contract in this PR.

Tests cover cache-only standby operation with no writes, two replicas converging on assignment updates, warm promotion without a new Kubernetes event, exact UID replacement semantics, and cancellation/drain behavior under the race detector.

internal/sgpu/inventory/reconcile.go
```diff
@@ -0,0 +1,1564 @@
...
+				return result, nil
+			}
+			continue
+		}
+		for rackIndex := int32(0); rackIndex < group.group.Count; rackIndex++ {
+			if err := reconcileRack(rackIndex); err != nil {
+				sortResult(&result)
+				return result, err
+			}
+			if !result.ResolvedRefs || capacityGrowthBlocked {
```
tjrasche: Addressed in [`1fb9ecab`](https://github.com/NVIDIA/k8s-test-infra/commit/1fb9ecab4eedc37e3a07da80984e359768764843). Reconciliation now tracks materialization failure per rack group instead of reusing inventory-wide `Result.ResolvedRefs` as the rack-loop break condition. A missing sibling profile still records `ProfileIssues` and sets `ResolvedRefs=false`, while healthy groups continue processing all requested racks.

I added regression coverage for both full inventory reconciliation and targeted `ReconcileGroup`. Focused race tests and the full test suite pass.

deployments/nvml-mock/helm/nvml-mock/templates/controlplane/deployment.yaml
tjrasche: Addressed in [`d7bfe081`](https://github.com/NVIDIA/k8s-test-infra/commit/d7bfe081c0f95a2aafb3cbebf39f2a25a20dd2bc). The control-plane Deployment now uses an explicit `RollingUpdate` strategy with `maxUnavailable: 0` and `maxSurge: 1`, preserving a serving replica while its replacement becomes ready. Leader election continues to enforce single-writer mutation.

The Helm tests now assert the rollout settings for both default and multi-replica renders, and the controller documentation describes the resulting upgrade behavior. `make helm-tests` passes.

internal/sgpu/inventory/allocate/allocator.go
```diff
@@ -0,0 +1,400 @@
...
+}
+
+func indexNodes(nodes []Node) (map[types.UID]Node, error) {
+	indexed := make(map[types.UID]Node, len(nodes))
+	for _, node := range nodes {
+		if node.Name == "" || node.UID == "" {
+			return nil, errors.New("node identity must include name and UID")
+		}
+		if _, exists := indexed[node.UID]; exists {
+			return nil, fmt.Errorf("duplicate node UID %q", node.UID)
```
tjrasche: Yes, this case should fail. It represents two live Node records with the same Kubernetes object UID, rather than the separately handled case of duplicate persisted rack bindings. UID is the allocator’s durable identity, so continuing would require arbitrarily choosing one Node representation and could assign that UID to multiple rack coordinates.

Normal same-name Node replacement uses a new UID and is handled by the Node catalog. Treating duplicate live UIDs as an invariant violation and returning no allocation plan is therefore the safe behavior; no code change is needed here.

internal/controlplane/api/v1alpha1/crd_retention_test.go
tjrasche: I considered moving this to Helm unit tests, but the retention annotation is authored through Kubebuilder markers and emitted by `controller-gen`; the CRD chart then ships those generated manifests unchanged. This test is therefore a semantic generated-artifact contract, alongside the existing generated CRD schema tests in the same Go package.

`make gen-check` verifies source-to-generated synchronization, while the CRD Helm target verifies that the static chart lints and renders. A Helm unit test would inspect the same YAML without exercising uninstall behavior, so it would duplicate this assertion rather than cover a different boundary. An actual uninstall guarantee would require a cluster integration test. I’m keeping the Go test as-is.

tests/kwok/cmd/kwok-assert/main_test.go
```diff
@@ -0,0 +1,120 @@
...
+	require.True(t, got.Success, got.Errors)
+	require.NotEmpty(t, got.AssignmentDigest)
+	require.Equal(t, 1, got.ProjectedNodes)
+
+	var projected assignment
+	require.NoError(t, json.Unmarshal([]byte(nodes.Items[0].Annotations[assignmentAnnotation]), &projected))
+	projected.NodeUID = "replacement-uid"
+	drifted, err := json.Marshal(projected)
+	require.NoError(t, err)
+	nodes.Items[0].Annotations[assignmentAnnotation] = string(drifted)
```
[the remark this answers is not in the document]
tjrasche: Fixed in [`c901643b`](https://github.com/NVIDIA/k8s-test-infra/commit/c901643b7e40a32c1d6b08ae61b75d3f1af75018). The KWOK assertion tests now locate `mokka-node-000000` by name before mutating its assignment metadata, rather than selecting `nodes.Items[0]`. The same name-based lookup is used by the released-metadata test so neither test depends on Kubernetes list order.

internal/controlplane/api/v1alpha1/examples_test.go
```diff
@@ -0,0 +1,41 @@
...
+	examples := filepath.Join("..", "..", "..", "..", "examples", "mokka-controller")
+	profileData, err := os.ReadFile(filepath.Join(examples, "sgpu-rack-profile.yaml"))
+	require.NoError(t, err)
+	profile := &mokkav1alpha1.SGPURackProfile{}
+	require.NoError(t, yaml.UnmarshalStrict(profileData, profile))
+	require.Equal(t, mokkav1alpha1.SchemeGroupVersion.String(), profile.APIVersion)
+	require.Equal(t, "SGPURackProfile", profile.Kind)
+	require.NoError(t, rackrender.ValidateProfile(profile.Spec))
+	require.Positive(t, profile.Spec.Node.GPUs.Memory.Capacity.Value())
+	require.Positive(t, profile.Spec.Node.GPUs.Clocks.Supported[0].MemoryMHz)
```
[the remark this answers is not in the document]
tjrasche: Fixed in [`c901643b`](https://github.com/NVIDIA/k8s-test-infra/commit/c901643b7e40a32c1d6b08ae61b75d3f1af75018). The example test now requires a non-empty supported-clock collection and validates every entry’s `MemoryMHz` value, avoiding positional selection while strengthening coverage for additional clock entries.

deployments/nvml-mock/helm/nvml-mock/values.yaml
tjrasche: Yes. I increased the control-plane memory limit from `256Mi` to `1Gi` while running the 100k-node KWOK scale scenarios. The controller’s informer caches and reconciliation state grow with the number of Nodes and racks, and the previous limit was too low at that scale. The KWOK harness runs the binary directly on the host, so I used the observed process footprint to size the Helm Deployment limit.

hack/update-mokka-codegen.sh
tjrasche: I’m keeping this as a shell script. It only installs pinned upstream Kubernetes generators and invokes them with fixed arguments; a Go wrapper would reproduce the same subprocess, temporary-directory, and argument plumbing without improving determinism or portability.

internal/sgpu/inventory/allocate/allocator.go
```diff
@@ -0,0 +1,400 @@
...
+	selector labels.Selector
+	occupied map[uint64]struct{}
+	cursor   int64
+}
+
+// Allocate produces a deterministic plan without compacting valid bindings or
+// reusing coordinates whose cleanup is still pending.
+//
+//nolint:cyclop // One deterministic pass records each retention, release, conflict, and assignment decision.
+func Allocate(input Input) (Plan, error) {
```
tjrasche: I’m keeping `Plan`. The allocator input bindings are observed state, while `Plan` records the retained, released, assigned, pending, and conflict decisions that reconciliation acts on; calling that result `State` would erase the transition distinction.

internal/sgpu/inventory/allocate/selectors.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	compiled, err := metav1.LabelSelectorAsSelector(selector)
+	if err != nil {
+		return nil, err
+	}
+	return compiled, nil
+}
+
+// Classify evaluates every eligible Node against precompiled group selectors.
+func Classify(nodes []Node, groups CompiledGroups) ([]Classification, Stats) {
```
tjrasche: I’m keeping `Classify`. It filters eligible Nodes and records every matching rack group; it does not score or choose a preferred group, and multiple matches become overlap conflicts. In Kubernetes terminology, ranking would imply a Score phase that is not present here.

internal/sgpu/inventory/allocate/allocator.go
```diff
@@ -0,0 +1,400 @@
...
+}
+
+// NodeReference pins a binding to an exact Node instance.
+type NodeReference struct {
+	Name string
+	UID  types.UID
+}
+
+// Coordinate identifies one durable logical Node in a rack.
+type Coordinate struct {
```
tjrasche: I’m keeping `Coordinate`. It is the identity tuple of inventory, rack group, rack index, and logical Node index, not a materialized `SGPURack` Node with rendered GPU data and an optional Kubernetes Node reference. `RackNode` would conflate the address with the API object.

internal/sgpu/inventory/allocate/allocator.go
```diff
@@ -0,0 +1,400 @@
...
+	Group     GroupKey
+	RackIndex int32
+	NodeIndex int32
+}
+
+// Binding assigns an exact Node instance to one coordinate.
+type Binding struct {
```
tjrasche: I’m keeping `Binding`. In the package-qualified form, `allocate.Binding` joins one logical coordinate to an exact Kubernetes Node name and UID; it neither contains nor identifies an exact `SGPURack` object. `RackNodeBinding` would imply that richer API object.

internal/controlplane/controller.go
tjrasche: I moved this validation onto `Config` as the unexported `validate()` method. `NewController` now calls it, and the existing predicates, error messages, ordering, and tests are preserved in `f9976f9b`.

enhancements/meps/0001-mokka-control-plane/README.md
tjrasche: I rewrote MEP-0001 as the normative final architecture in `454b4ec2`. It now describes the target Kubernetes-authoritative design, leader-only writes, synchronized read replicas, lazy runtime-view delivery, agent polling and last-known-good state, Kubernetes Node lifecycle reclamation, and the no-Redis boundary. Current Stage 1 status is linked from the operational guide instead of being mixed into the MEP.

enhancements/meps/0001-mokka-control-plane/README.md
tjrasche: I moved the prior considerations into the `Alternatives and Design History` appendix in `454b4ec2`. It preserves the trade-offs around per-Node allocation resources, Redis, heartbeat-based reclamation, and eager fanout, and records why the final architecture retains or rejects each approach.

deployments/nvml-mock/helm/nvml-mock/templates/_helpers.tpl
tjrasche: I changed the fixed cluster-scoped singleton RBAC identity to `control-plane.mokka.nvidia.com` in `b3c94704` and updated the Helm tests and operational documentation.

deployments/nvml-mock/helm/nvml-mock/values.yaml
tjrasche: I changed the default leader-election Lease name to `control-plane.mokka.nvidia.com` in `b3c94704`, including the Go and Helm defaults, tests, and documentation. The Lease name remains configurable independently of the fixed singleton RBAC identity.

internal/sgpu/inventory/allocate/allocator.go
```diff
@@ -0,0 +1,400 @@
...
+	InventoryUID  types.UID
+	RackGroup     string
+}
+
+func (key GroupKey) String() string {
+	return key.InventoryName + "/" + key.RackGroup
+}
+
+// Group is a resolved placement and capacity declaration.
+type Group struct {
```
tjrasche: I renamed these types to `RackGroup` and `RackGroupKey` across the allocator and its consumers in `f9976f9b`, without compatibility aliases or behavior changes.

internal/sgpu/inventory/allocate/allocator.go
```diff
@@ -0,0 +1,400 @@
...
+	Key          GroupKey
+	Selector     *metav1.LabelSelector
+	Racks        int32
+	NodesPerRack int32
+}
+
+// Node is the allocation-relevant portion of a Kubernetes Node.
+type Node struct {
```
tjrasche: I renamed this type to `KubernetesNode` in `f9976f9b`. I used the expanded name rather than `K8sNode` to make its relationship to the Kubernetes API Node explicit while keeping it distinct from `corev1.Node`.

internal/sgpu/inventory/allocate/allocator.go
```diff
@@ -0,0 +1,400 @@
...
+
+// Stats exposes bounded-work counters for large-snapshot tests and benchmarks.
+type Stats struct {
+	SelectorEvaluations int64
+	BindingLookups      int64
+	NodeVisits          int64
+}
+
+// Input is a complete immutable allocation snapshot.
+type Input struct {
```
tjrasche: I renamed `Input` to `Snapshot` in `f9976f9b`. The value is an immutable observed allocation snapshot containing rack groups, Kubernetes Nodes, and current bindings rather than an allocation request or an allocations-only result.

docs/mokka-controller.md
```diff
@@ -0,0 +1,214 @@
+<!-- SPDX-License-Identifier: Apache-2.0 -->
+<!-- SPDX-FileCopyrightText: Copyright 2026 NVIDIA CORPORATION -->
+
+# Mokka controller
```
roma-glushko: Should say control plane. We don't really look at controller in isolation as a separate thing.

roma-glushko (dismissed):
I think the new changes addresses my feedback. @tjrasche thank you for much for rigorously applying them!

Can't want to see this changes finally in the main 🚀

internal/sgpu/inventory/admission_integration_test.go
```diff
@@ -0,0 +1,333 @@
...
+
+	reconcile := func(t *testing.T) {
+		t.Helper()
+		admissionReconcile(ctx, t, mokka, inventory.Name, profile)
+	}
+	reconcile(t)
+	racks, err := mokka.MokkaV1alpha1().SGPURacks().List(ctx, metav1.ListOptions{})
+	require.NoError(t, err)
+	require.Len(t, racks.Items, 1)
+	rack := &racks.Items[0]
```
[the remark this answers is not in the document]
tjrasche: The test requires an otherwise unused disposable cluster and asserts `require.Len(t, racks.Items, 1)` immediately before this access. `require.Len` stops the test on failure, so an unexpected extra rack fails the assertion before any mutation. With exactly one returned rack, list ordering cannot change which resource is selected.

The blanket positional-access rule flags the syntax, but the reported wrong-resource mutation does not apply to this guarded singleton fixture. Resolving this case without changing the test; this is not a general exemption for test code.

roma-glushko (approved):

PR #691, opened 2026-08-21
feat(mocknvml): report clocks event reason counters for nvidia-smi -q

### LINKED ISSUE #678: [Bug]: Clocks Event Reasons Counters report N/A because the perf-policy field IDs are unhandled and GetViolationStatus is hardcoded to zero
## Component / version

`k8s-test-infra` @ `95e94598` (`origin/main`); `nvml-mock` chart `0.3.0`, profile `gb300`.

Found by diffing `nvidia-smi -q` from a real 4-GPU GB300 tray (driver `580.173.02`, CUDA 13.0) against the `gb300` profile running in kind, extending the audit behind #644 to Grace-Blackwell.

## What happened

`nvidia-smi -q` reports a `Clocks Event Reasons Counters` section: the cumulative microseconds a GPU has spent throttled by each cause. On the mock every counter is `N/A`, while the *flags* immediately above them in the same output — `Clocks Event Reasons` — report real, configurable values. The result is an output block that contradicts itself: the mock states with confidence that no throttle reason is currently active, and then cannot say whether any has ever been active.

The counters are how throttling is actually diagnosed after the fact. A workload that ran slow is investigated by reading accumulated `SW Power Capping` or `HW Thermal Slowdown` time, not by sampling the instantaneous flag and hoping to catch it. On the mock that path returns nothing.

### Example output

`profiles/gb300.yaml` — flags populated, counters absent:

```
    Clocks Event Reasons
        Idle                              : Active
        Applications Clocks Setting       : Not Active
        SW Power Cap                      : Not Active
        HW Slowdown                       : Not Active
            HW Thermal Slowdown           : Not Active
            HW Power Brake Slowdown       : Not Active
        Sync Boost                        : Not Active
        SW Thermal Slowdown               : Not Active
        Display Clock Setting             : Not Active
    Clocks Event Reasons Counters
        SW Power Capping                  : N/A
        Sync Boost                        : N/A
        SW Thermal Slowdown               : N/A
        HW Thermal Slowdown               : N/A
        HW Power Braking                  : N/A
```

A real GB300 reports microsecond totals, and they are not uniformly zero — this GPU has been power-capped for 39.6 ms since reset:

```
    Clocks Event Reasons Counters
        SW Power Capping                                   : 39595 us
        Sync Boost                                         : 0 us
        SW Thermal Slowdown                                : 0 us
        HW Thermal Slowdown                                : 0 us
        HW Power Braking                                   : 0 us
```

## Expected behaviour

- The five counters report configurable per-cause microsecond totals rather than `N/A`, and default to `0` on a healthy profile — `0 us` is a real answer, `N/A` is not.
- Counters accumulate while the corresponding throttle reason is active, so a profile that reports `SW Power Cap : Active` does not simultaneously report zero power-capping time forever.
- The same values are readable through the field-value path used by DCGM, not only through `nvidia-smi`.
- `nvidia-smi pmon -c 1` and `topo -m` are unaffected.

## Root cause

Two layers both decline, for different reasons.

`nvmlDeviceGetViolationStatus` is implemented but hardcoded to no violation, in `pkg/gpu/mocknvml/engine/device.go`:

```go
func (d *ConfigurableDevice) GetViolationStatus(perfPolicyType nvml.PerfPolicyType) (nvml.ViolationTime, nvml.Return) {
    if ret := d.tickFailure(); ret != nvml.SUCCESS {
        return nvml.ViolationTime{}, ret
    }
    debugLog("[NVML] nvmlDeviceGetViolationStatus(policy=%d) -> no violation\n", perfPolicyType)
    return nvml.ViolationTime{}, nvml.SUCCESS
}
```

It ignores `perfPolicyType` entirely and returns a zero `ViolationTime` for every policy, so all five causes are indistinguishable.

That path is not the one producing the `N/A`, though. A 580-era `nvidia-smi` reads these counters through `nvmlDeviceGetFieldValues` using the `NVML_FI_DEV_PERF_POLICY_*` field IDs (74–81), and `pkg/gpu/mocknvml/engine/field_values.go` has no case for them — an unhandled field ID comes back per-field `NOT_SUPPORTED` while the overall call succeeds, which is exactly how a field renders as `N/A` with no stub appearing in `MOCK_NVML_DEBUG` output. That is why the debug census for a full `-q` run shows nothing for this section: no unimplemented symbol is reached.

So the flags and the counters are not merely unrelated, they are answered by two different mechanisms, only one of which is wired.

## Steps to reproduce

```bash
  -v "$PWD/deployments/nvml-mock/helm/nvml-mock/profiles/gb300.yaml:/config/config.yaml:ro" \
  nvml-mock:latest nvidia-smi -q | grep -A 6 "Clocks Event Reasons Counters"

# confirm no stub is reached — the N/A comes from an unhandled field ID
  -v "$PWD/deployments/nvml-mock/helm/nvml-mock/profiles/gb300.yaml:/config/config.yaml:ro" \
```

## Environment

```yaml
host: darwin 25.5.0 (arm64), kind cluster kind-nvml-mock-demo
image: nvml-mock built from afd8857c; source claims verified against origin/main 95e94598
profile: gb300 (blackwell)
nvidia-smi in image: 580.65.06
reference hardware: 4x NVIDIA GB300, driver 580.173.02, CUDA 13.0
```

## Acceptance criteria

- [ ] The five `NVML_FI_DEV_PERF_POLICY_*` field IDs are handled in `getDeviceFieldValue`, and `nvidia-smi -q` reports microsecond totals rather than `N/A`.
- [ ] `nvmlDeviceGetViolationStatus` honours `perfPolicyType` and returns per-cause values, so the five causes are distinguishable.
- [ ] Per-cause violation time is configurable per device, defaulting to `0`.
- [ ] Counters accumulate while the matching throttle reason is active, so the flags and the counters cannot contradict each other.
- [ ] The same values are visible through the field-value path used by DCGM, not only through `nvidia-smi`.
- [ ] `nvidia-smi pmon -c 1` does not regress against `main` and `topo -m` still exits `0`.
- [ ] Unit tests cover each policy type separately, including that setting one cause leaves the others at zero.
- [ ] **Dedicated e2e test.** A new Ginkgo spec under `tests/e2e/go/` with `Label("nvidia-smi")` that asserts, from `nvidia-smi -q -x`, that a profile configuring a non-zero `SW Power Capping` time reports exactly that value while the other four causes report `0 us` — using a distinctive value so it cannot pass by coincidence — and that a healthy profile reports `0 us` rather than `N/A` for all five. The test must fail against the current `main`.

## Out of scope

The `Clocks Event Reasons` flags, which already work and are configurable via `clocks_throttle_reasons`. Also out of scope: making throttling actually reduce reported clocks, which belongs with the dynamic-metrics coupling in #643.

giuliocalzo:
Fixes #678

## Summary

All five counters in the `nvidia-smi -q` `Clocks Event Reasons Counters` block read `N/A`, so the output contradicted itself: the flags directly above them stated confidently that no throttle reason was active, and the counters could not say whether one ever had been. The counters are how throttling is diagnosed after the fact — a workload that ran slow is investigated by reading accumulated `SW Power Capping` or `HW Thermal Slowdown` time, not by sampling the instantaneous flag and hoping to catch it.

Two layers declined, for different reasons:

- `GetViolationStatus` was implemented but ignored its `perfPolicyType` and returned a zero `nvmlViolationTime_t` for every policy, leaving the five causes indistinguishable.
- That is not the path producing the `N/A`. A 580-era `nvidia-smi` reads the counters through `nvmlDeviceGetFieldValues`, and an unhandled field id there comes back per-field `NOT_SUPPORTED` while the overall call succeeds — which renders as `N/A` with no unimplemented-symbol stub reached, and is why the debug census for a full `-q` run showed nothing for this section.

Both paths now resolve from one per-device accrual, so they cannot disagree.

A `clocks_throttle_reasons.counters` block seeds the accrued time per cause per device in microseconds, defaulting to `0 us` — a real answer, where `N/A` is not. A device accrues further time on top of its baseline while the matching flag is set, within one process — see the behavioural note below for what that does and does not buy.

Against the `gb200` profile with `sw_power_cap_us: 39595` on GPU 0:

```
    Clocks Event Reasons Counters
        SW Power Capping                  : 39595 us
        Sync Boost                        : 0 us
        SW Thermal Slowdown               : 0 us
        HW Thermal Slowdown               : 0 us
        HW Power Braking                  : 0 us
```

## The field ids are not the ones the issue reports

`nvidia-smi` 580.65.06 asks for **74, 76, 251, 252 and 253** at the whole-GPU scope, established by tracing a real `-q` field-value batch. The vendored CUDA 13 header numbers the three slowdown causes 269-271 and reuses 251-253 for power smoothing, so taking the header at its word left those three still reading `N/A` after the other two were fixed. The simulated driver's numbering is what consumers see, so it is the one used here, pinned by a test that fails if a header bump moves it silently.

## Second defect found on the way

Applying the counters per device surfaced an unrelated bug: `mergeDeviceOverride` had no case for `clocks_throttle_reasons` at all, so per-device throttle config in **any** profile was being dropped.

## Behavioural note

Accrual is folded in on read and is per process, matching the dynamic-metrics simulator: a long-lived consumer such as dcgm-exporter watches a counter climb while a reason is active, and each `nvidia-smi` invocation accrues only over its own lifetime. Cross-process accrual would need shared node state, so this is documented in `docs/configuration.md` rather than worked around.

The consequence is that a throttle state entered at runtime carries no history of its own — `nvidia-smi` shows the reason `Active` beside `1 us`, the age of that `nvidia-smi` process. The counters are what make an `Active` flag read as a duration, so seed both together:

```bash
nvml-mock-ctl throttle --gpu 3 thermal
nvml-mock-ctl set --gpu 3 clocks_throttle_reasons.counters.hw_thermal_slowdown_us=39595
```

Both land in the same override document, so the second command keeps the first.

## Test plan

- [x] `make test` — engine unit tests cover each policy and each cause separately, that setting one leaves the others at zero, that accrual advances while a reason is active and survives it clearing, and that the field-id mapping matches the simulated driver
- [x] `make test-mocknvml-bridge` — reads all five counters and three violation statuses across the C ABI on two differently-configured devices, so the marshalling is exercised rather than assumed
- [x] `make helm-tests` — 169 tests, 15 snapshots, no profile churn
- [x] `nvidia-smi -q` / `-q -x` verified in the mock image against a seeded profile and, separately, through the `nvml-mock-ctl set` → read back → `reset` loop
- [x] `nvidia-smi pmon -c 1` exits 255 and `topo -m` exits 0, both unchanged from the baseline
- [ ] e2e: a `Label("nvidia-smi")` spec asserts five zeros on every profile, and a runtime spec seeds a distinctive total on one GPU and requires the other four causes — and the node's other GPUs — to stay at zero. The pre-fix capture already in `testdata` is the defect the checks must reject, which is verified by unit test; the cluster legs run in CI here.

### DIFF
--- a/pkg/gpu/mocknvml/engine/failure_injector_test.go
+++ b/pkg/gpu/mocknvml/engine/failure_injector_test.go
@@ -16,6 +16,7 @@ package engine
 import (
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/NVIDIA/go-nvml/pkg/nvml"
 	"github.com/stretchr/testify/require"
@@ -392,10 +393,16 @@ func TestFailureInjection_GetViolationStatus_StaysSpecCompliantAfterTrip(t *test
 	// SECOND tick, which is GetViolationStatus itself). Neither field may
 	// carry the Xid: violation time reflects the (unthrottled) device, and
 	// reference time is a timestamp.
+	before := uint64(time.Now().UnixMicro())
 	vt, ret := dev.GetViolationStatus(nvml.PERF_POLICY_POWER)
 	require.Equal(t, nvml.SUCCESS, ret, "expected SUCCESS")
 	require.Zero(t, vt.ViolationTime, "ViolationTime must be 0 ns post-trip (Xid no longer overloaded)")
-	require.NotEqual(t, uint64(64), vt.ReferenceTime, "ReferenceTime must not carry the Xid either")
+	// Bracketing the call rather than merely excluding 64: any real clock
+	// reading is ~1.7e15, so "not the Xid" is satisfied by every implementation
+	// including the one that returned zero, and would not catch a regression.
+	require.GreaterOrEqual(t, vt.ReferenceTime, before,
+		"ReferenceTime must be a CPU timestamp taken during the call, not an Xid or a zero")
+	require.LessOrEqual(t, vt.ReferenceTime, uint64(time.Now().UnixMicro()))
 }
 
 func TestFailureInjection_GetViolationStatus_LostModeReturnsError(t *testing.T) {
--- a/pkg/gpu/mocknvml/engine/failure_injector_test.go
+++ b/pkg/gpu/mocknvml/engine/failure_injector_test.go
@@ -363,8 +363,10 @@ func TestFailureInjection_GetViolationStatus_HealthyReportsNoViolation(t *testin
 	dev := newTestDeviceWithConfig(t, healthyConfig())
 	vt, ret := dev.GetViolationStatus(nvml.PERF_POLICY_POWER)
 	require.Equal(t, nvml.SUCCESS, ret, "expected SUCCESS")
-	require.Zero(t, vt.ViolationTime, "healthy device must report empty ViolationTime")
-	require.Zero(t, vt.ReferenceTime, "healthy device must report empty ReferenceTime")
+	require.Zero(t, vt.ViolationTime, "healthy device must report no accrued violation time")
+	// ReferenceTime is the CPU timestamp the reading was taken at, so it is
+	// populated even on a device that has never been throttled.
+	require.NotZero(t, vt.ReferenceTime, "healthy device must still timestamp the reading")
 }
 
 func TestFailureInjection_GetViolationStatus_StaysSpecCompliantAfterTrip(t *testing.T) {
@@ -387,13 +389,13 @@ func TestFailureInjection_GetViolationStatus_StaysSpecCompliantAfterTrip(t *test
 	require.Equal(t, nvml.SUCCESS, ret, "setup call 1")
 
 	// At this point the device has just tripped (AfterCalls=2 met by the
-	// SECOND tick, which is GetViolationStatus itself). It must NOT
-	// stuff the Xid into the violation_time field; both fields stay
-	// at their healthy zero values.
+	// SECOND tick, which is GetViolationStatus itself). Neither field may
+	// carry the Xid: violation time reflects the (unthrottled) device, and
+	// reference time is a timestamp.
 	vt, ret := dev.GetViolationStatus(nvml.PERF_POLICY_POWER)
 	require.Equal(t, nvml.SUCCESS, ret, "expected SUCCESS")
 	require.Zero(t, vt.ViolationTime, "ViolationTime must be 0 ns post-trip (Xid no longer overloaded)")
-	require.Zero(t, vt.ReferenceTime, "ReferenceTime must be 0 ns post-trip")
+	require.NotEqual(t, uint64(64), vt.ReferenceTime, "ReferenceTime must not carry the Xid either")
 }
 
 func TestFailureInjection_GetViolationStatus_LostModeReturnsError(t *testing.T) {
--- a/pkg/gpu/mocknvml/engine/throttle_counters.go
+++ b/pkg/gpu/mocknvml/engine/throttle_counters.go
@@ -1,0 +1,241 @@
+// Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Cumulative clocks-event (throttle) time, per cause. The instantaneous
+// clocks_throttle_reasons flags say whether the GPU is being held back right
+// now; these counters say how long it has been held back since reset, which is
+// how throttling is diagnosed after a workload has already run slow.
+//
+// nvidia-smi renders them as the "Clocks Event Reasons Counters" block, reading
+// them through nvmlDeviceGetFieldValues rather than any dedicated getter — an
+// unhandled field id there comes back per-field NOT_SUPPORTED and renders as
+// N/A, so the counters must be answered even when the answer is zero.
+
+package engine
+
+import (
+	"sync"
+	"time"
+
+	"github.com/NVIDIA/go-nvml/pkg/nvml"
+)
+
+// NVML field ids for the throttle counters. All carry nanoseconds.
+//
+// Two causes have long-standing ids in the perf-policy family, which
+// NVML_FI_DEV_CLOCKS_EVENT_REASON_SW_POWER_CAP and _SYNC_BOOST are defined as;
+// NVML_FI_DEV_PERF_POLICY_THERMAL is the older name for the SW thermal counter.
+// The three slowdown causes only exist under the newer names.
+//
+// Those three are NOT taken from the header this repo vendors, which is a CUDA
+// 13 header numbering them 269-271 and reusing 251-253 for power smoothing.
+// The driver these profiles report — and the nvidia-smi 580.65.06 the image
+// bundles — has them at 251-253 and power smoothing at 256-273 instead. The
+// simulated driver's numbering is the one consumers see, so it is the one used
+// here; verified by tracing a `nvidia-smi -q` field-value batch, which asks for
+// exactly 74, 76, 251, 252 and 253 at the whole-GPU scope.
+const (
+	fiPerfPolicyPower     = 74
+	fiPerfPolicyThermal   = 75
+	fiPerfPolicySyncBoost = 76
+
+	fiClocksEventSwThermSlowdown = 251
+	fiClocksEventHwThermSlowdown = 252
+	fiClocksEventHwPowerBrake    = 253
+)
+
+// throttleCause enumerates the causes that accrue time independently. These are
+// exactly the five rows nvidia-smi prints, which is also the granularity NVML
+// exposes: the umbrella "HW Slowdown" flag has no counter of its own.
+type throttleCause int
+
+const (
+	throttleCauseSWPowerCap throttleCause = iota
+	throttleCauseSyncBoost
+	throttleCauseSWThermal
+	throttleCauseHWThermal
+	throttleCauseHWPowerBrake
+	throttleCauseCount
+)
+
+// throttleAccrual tracks how long each cause has been active on one device.
+// Time is folded in on read rather than by a background ticker: a mock has no
+// reason to burn a goroutine per device, and a counter nobody reads has no
+// observable value.
+type throttleAccrual struct {
+	device *ConfigurableDevice
+
+	// now is overridable in tests so accrual can be driven deterministically
+	// without sleeping.
+	now func() time.Time
+
+	mu sync.Mutex
+	// accrued is the time each cause has been observed active for, and
+	// activeSince is when it was last observed (zero == currently inactive).
+	// A read advances accrued to the moment of that read, so a cause the mock
+	// never saw active contributes nothing and the totals never go backwards.
+	accrued     [throttleCauseCount]time.Duration
+	activeSince [throttleCauseCount]time.Time
+}
+
+// throttleAccrual returns the device's accrual state, creating it on first use.
+// Lazy creation keeps every device-construction path (including the legacy
+// no-YAML one) from having to know about it.
+func (d *ConfigurableDevice) throttleAccrual() *throttleAccrual {
+	if a := d.throttle.Load(); a != nil {
+		return a
+	}
+	fresh := &throttleAccrual{device: d, now: time.Now}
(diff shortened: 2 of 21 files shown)


pkg/gpu/mocknvml/engine/throttle_counters.go
```diff
@@ -0,0 +1,241 @@
...
+	switch perfPolicyType {
+	case nvml.PERF_POLICY_POWER:
+		cause = throttleCauseSWPowerCap
+	case nvml.PERF_POLICY_THERMAL:
+		cause = throttleCauseSWThermal
+	case nvml.PERF_POLICY_SYNC_BOOST:
+		cause = throttleCauseSyncBoost
+	default:
+		debugLog("[NVML] nvmlDeviceGetViolationStatus(policy=%d) -> NOT_SUPPORTED\n", perfPolicyType)
+		return nvml.ViolationTime{}, nvml.ERROR_NOT_SUPPORTED
```
ArangoGutierrez, line 228: This narrows four policies that previously returned SUCCESS with a zero ViolationTime to ERROR_NOT_SUPPORTED. The vendored nvml.h maps BOARD_LIMIT, LOW_UTILIZATION, RELIABILITY and TOTAL_BASE_CLOCKS onto real field ids and marks only TOTAL_APP_CLOCKS as "DEPRECATED, Do not use", so the comment above about them having no counterpart doesn't match the header. Declining just TOTAL_APP_CLOCKS and keeping SUCCESS plus the timestamp for the other four would avoid a behaviour change no consumer asked for, and it isn't in the CHANGELOG either way.

tests/mocknvml/bridge_tests.go
```diff
@@ -43,9 +44,146 @@ func bridgeTests(deviceCount int) []testResult {
...
 	results = append(results, testFabricHealth(deviceCount)...)
+	results = append(results, testThrottleCounters(deviceCount)...)
 	return results
 }
 
+// --- Throttle counter tests ---
+
+// Field ids for the "Clocks Event Reasons Counters", in the numbering the
+// simulated driver uses. go-nvml names the first two (FI_DEV_PERF_POLICY_POWER
+// and _SYNC_BOOST) but has no constant for the three slowdown causes at these
```
ArangoGutierrez: This module pins go-nvml v0.13.0-1, and its vendored const.go defines FI_DEV_CLOCKS_EVENT_REASON_SW_THERM_SLOWDOWN, _HW_THERM_SLOWDOWN and _HW_POWER_BRAKE_SLOWDOWN as exactly 251, 252 and 253, so the constants do exist here and the ids can be named instead of hardcoded. The 269-271 claim holds for the root module's v0.13.3-1, not for this one. Given how often a go-nvml bump bites this repo, worth the comment being right about which tree it describes.

pkg/gpu/mocknvml/engine/failure_injector_test.go
```diff
@@ -387,13 +389,13 @@ func TestFailureInjection_GetViolationStatus_StaysSpecCompliantAfterTrip(t *test
...
-	// stuff the Xid into the violation_time field; both fields stay
-	// at their healthy zero values.
+	// SECOND tick, which is GetViolationStatus itself). Neither field may
+	// carry the Xid: violation time reflects the (unthrottled) device, and
+	// reference time is a timestamp.
 	vt, ret := dev.GetViolationStatus(nvml.PERF_POLICY_POWER)
 	require.Equal(t, nvml.SUCCESS, ret, "expected SUCCESS")
 	require.Zero(t, vt.ViolationTime, "ViolationTime must be 0 ns post-trip (Xid no longer overloaded)")
-	require.Zero(t, vt.ReferenceTime, "ReferenceTime must be 0 ns post-trip")
+	require.NotEqual(t, uint64(64), vt.ReferenceTime, "ReferenceTime must not carry the Xid either")
```
ArangoGutierrez: This one can't fail. ReferenceTime is now time.Now().UnixMicro(), around 1.7e15, so it never equals 64 for any implementation, including the old one that returned zero, which the require.Zero it replaced would have caught. Comparing against a window around time.Now().UnixMicro() keeps the guard discriminating while still ruling out the Xid.

ArangoGutierrez (commented):
Nothing here blocks merge.

The field-id archaeology is the part that makes this land: pinning 74/76/251-253 to what the simulated driver actually answers, rather than to the vendored CUDA 13 header, is the right call, and TestThrottleCounters_FieldIDsMatchTheSimulatedDriver keeps a header bump from moving it quietly. Fold-on-read accrual with no background goroutine reads well, and the per-device isolation is checked on both the Go and the C ABI side. I also checked all seven shipped profiles have the five causes set false, so the e2e five-zeros assertion is sound rather than lucky. A few notes below, none of them about the core mechanism.

1. "the flags and the counters cannot contradict each other" is a bit stronger than the mechanism supports, and the next sentence is what undoes it: since accrual is per process, a GPU put into a throttle state by nvml-mock-ctl throttle carries no baseline, so nvidia-smi renders the reason Active next to 0 us, which is #678's contradiction inverted. Either have the flag-setting paths seed a baseline, or soften the claim here and in the CHANGELOG. (docs/configuration.md:264)
2. Worth a line in the new section noting that a per-device clocks_throttle_reasons block replaces the device_defaults one wholesale rather than merging into it, so a devices: entry carrying only counters drops the inherited flags. It matches how Memory, Power, Thermal and Clocks already behave, so it's no surprise in the merge code, but it is from the YAML side: util-test-config.yaml here has to restate gpu_idle: true for exactly that reason. (docs/configuration.md:252)

pkg/gpu/mocknvml/engine/throttle_counters.go
giuliocalzo: Right, and the header is not the only reason — declining them contradicted the argument this change rests on. Zero is the answer for a limiter that never fired; NOT_SUPPORTED says the driver cannot tell, which is exactly the misreading that made these counters render as N/A.

Fixed in 8f961cac: BOARD_LIMIT, LOW_UTILIZATION, RELIABILITY and TOTAL_BASE_CLOCKS report zero with the timestamp again, so no consumer sees a behaviour change. Only TOTAL_APP_CLOCKS declines, since it is the one the header gives no field id. A value outside the enum — which has a 6-9 gap as well as an upper bound — now reports INVALID_ARGUMENT rather than a plausible zero. The CHANGELOG spells all three out.

tests/mocknvml/bridge_tests.go
giuliocalzo: Correct, and my comment described the wrong tree. This module pins go-nvml v0.13.0-1, where the three names are exactly 251, 252 and 253; the 269-271 renumbering is the root module's v0.13.3-1, which is why the engine states its ids literally instead of naming them.

Named in 8f961cac. I added `checkThrottleFieldIDs` alongside, because the coupling cuts both ways: if this module is bumped too, the names would silently retarget all five counter assertions at ids the engine deliberately does not answer, and the failure would read as a counter regression. The guard reports the bump instead.

pkg/gpu/mocknvml/engine/failure_injector_test.go
giuliocalzo: Agreed — 64 against a ~1.7e15 microsecond clock is unfalsifiable, including against the zero it replaced, which is the one thing it was meant to catch.

8f961cac brackets the call: ReferenceTime must land between a timestamp taken just before it and one taken just after, which rules out the Xid, a zero and a stale reading in one assertion.

ArangoGutierrez (dismissed):
/lgtm

giuliocalzo:
Rebased onto `main` — the only conflict was both of us adding an entry at the top of `[Unreleased]`, so both are kept.

I missed the two notes in your review body on the first pass; both are addressed in 11bcfa06.

**On the guarantee being stronger than the mechanism.** You're right, and it's worse than the inverted contradiction you describe — I reproduced it against `gb200`, and the counter reads `1 us`, not `0 us`:

```
    HW Thermal Slowdown               : Active
...
        HW Thermal Slowdown               : 1 us
```

That microsecond is the age of the `nvidia-smi` process, so a small plausible-looking number is arguably a worse answer than a zero would be.

I looked at making the claim true rather than softening it. The only timestamp that outlives the process is the override file's mtime, and any later unrelated write to it — a temperature pin, say — would bump the accrual origin and move a counter backwards, which no real NVML counter does. Not worth buying inside this PR, so the claim is now scoped to one process in `docs/configuration.md`, the `CHANGELOG`, the `ThrottleCountersConfig` doc comment and the PR description, with the recipe that makes an `Active` flag read as a duration:

```bash
nvml-mock-ctl throttle --gpu 3 thermal
nvml-mock-ctl set --gpu 3 clocks_throttle_reasons.counters.hw_thermal_slowdown_us=39595
```

Verified those compose — `set` merges into the existing override document, so the flag survives and `nvidia-smi` reports `Active` beside `39595 us`.

**On the wholesale replacement.** Documented in the new section, naming `memory`, `power`, `thermal` and `clocks` as the precedent and `util-test-config.yaml`'s repeated `gpu_idle: true` as the consequence.

PR #694, opened 2026-08-21
test(e2e): add real-hardware nvidia-smi captures

giuliocalzo:
## Summary

Adds `nvidia-smi -q -x` captures from real GPU nodes under `tests/e2e/go/assertions/nvidiasmi/testdata/hardware/`, kept whole rather than trimmed to two GPUs like the mock-produced fixtures one directory up.

Every nvidia-smi fixture we had came from the mock, so together they only proved that `schema.go` agrees with what we emit. These are the counterpart the mock is supposed to look like.

| File | Node | Driver / CUDA |
| --- | --- | --- |
| `qx-a100.xml` | 1x A100-SXM4-40GB | 570.148.08 / 12.8 |
| `qx-b200.xml` | 8x B200 | 580.105.08 / 13.0 |
| `qx-gb200.xml` | 4x GB200 (arm64) | 580.173.02 / 13.0 |
| `qx-gb300.xml` | 4x GB300 | 580.173.02 / 13.0 |
| `qx-h100.xml` | 8x H100 80GB HBM3 | 580.105.08 / 13.0 |
| `qx-l40s.xml` | 1x L40S | 595.91.07 / 13.2 |
| `qx-t4.xml` | 1x Tesla T4 | 595.91.07 / 13.2 |

The driver spread from 570 to 595 is deliberate: a driver bump can rename an element, and a renamed element decodes as an absent reading rather than as a parse error, so `hardware_test.go` asserts on the readings instead of on `ParseSnapshot` succeeding.

Coverage worth calling out:

- GB200 and GB300 are the only nodes in a fabric cluster, so the only captures where `clusterUuid` and `cliqueId` carry values rather than zeros or `N/A`.
- GB200 is the only arm64 host.
- A100 matches the board `mock-nvml-config-a100.yaml` models (`NVIDIA A100-SXM4-40GB`).
- T4 and L40S are the single-GPU, no-NVLink, no-fabric end of the range.

## Identifiers

Per-unit identifiers are replaced with placeholders: GPU serials and UUIDs, the chassis serial, the fabric `clusterUuid`, and the `pdi` and `gpu_fabric_guid` that name an individual die. Model-level fields are verbatim, since board and part numbers, VBIOS, PCI addresses, clocks and thresholds are what makes a capture worth keeping.

Two subtleties are documented in the directory README, because both are easy to get wrong when adding a capture:

- Bodies reporting an absent value (`N/A`, empty, an all-zero `clusterUuid`) are left alone, so a scrubbed file does not claim hardware the node lacks. `clusterUuid` placeholders are numbered from 1 for the same reason: zero is how nvidia-smi says "not in a fabric cluster".
- One placeholder per distinct value, so GPUs that shared an identifier upstream still share one here — the GB200 and GB300 modules each cover two GPUs and so report a serial twice. `pdi` and `gpu_fabric_guid` share one numbering, since Grace-Blackwell reports the same value in both.

`TestHardwareCapturesAreScrubbed` asserts the placeholder shapes over every file, so a capture added without scrubbing fails CI rather than merging. The patterns are narrow enough that a real reading cannot satisfy them.

## Test plan

- [x] `go test ./tests/e2e/go/assertions/...` passes; all seven captures decode and every asserted reading resolves
- [x] `make lint-fix` reports 0 issues and vet passes
- [x] Captures verified against the source: byte-identical by SHA-256 where the transport allowed it, and otherwise well-formed with `attached_gpus` matching the number of `<gpu>` blocks
- [x] Checked for MIG partitioning, vGPU mode and leftover process lists; none present, all report `Pass-Through`

Note: `make lint-fix` also runs `govulncheck`, which fails on pre-existing Go standard library advisories (needs go1.26.6) unrelated to this change; no Go dependencies are touched here.

### DIFF
--- a/tests/e2e/go/assertions/nvidiasmi/hardware_test.go
+++ b/tests/e2e/go/assertions/nvidiasmi/hardware_test.go
@@ -1,0 +1,264 @@
+// Copyright 2026 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+package nvidiasmi
+
+import (
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func hardwareCaptures(t *testing.T) []string {
+	t.Helper()
+	files, err := filepath.Glob("testdata/hardware/*.xml")
+	require.NoError(t, err)
+	require.NotEmpty(t, files)
+	return files
+}
+
+// The other fixtures in this package come from the mock, so on their own they
+// only prove the schema agrees with what we emit. These come from real nodes on
+// a newer driver than the mock targets, which is where an element rename would
+// show up first: a reading the schema can no longer find decodes as absent
+// rather than as an error, so the assertions here are on the readings, not on
+// ParseSnapshot succeeding.
+//
+// The coverage is the set of elements the checks in checks.go and temperature.go
+// compare, since those are the readings a rename would silently turn into "the
+// mock is wrong" for every profile at once. present() is the right test for
+// them: an unsupported query is "N/A", which passes, while a renamed element
+// leaves an empty body, which fails.
+func TestHardwareCapturesDecode(t *testing.T) {
+	for _, file := range hardwareCaptures(t) {
+		t.Run(filepath.Base(file), func(t *testing.T) {
+			raw, err := os.ReadFile(file)
+			require.NoError(t, err)
+
+			snap, err := ParseSnapshot(string(raw))
+			require.NoError(t, err)
+
+			attached, ok := snap.AttachedGPUs()
+			require.True(t, ok, "attached_gpus")
+			require.Equal(t, attached, snap.Count(), "every <gpu> block decoded")
+			require.False(t, snap.HasFailedGPU(), "captures are of healthy nodes")
+
+			for i := range snap.Count() {
+				gpu, err := snap.GPU(i)
+				require.NoError(t, err)
+				element := snap.doc.GPUs[i]
+				label := gpu.Label()
+
+				_, ok := gpu.TemperatureC()
+				require.True(t, ok, "%s: gpu_temp", label)
+				_, ok = gpu.SMClockMHz()
+				require.True(t, ok, "%s: sm_clock", label)
+				_, ok = gpu.MemoryTotalMiB()
+				require.True(t, ok, "%s: fb_memory_usage/total", label)
+				_, ok = gpu.PowerLimitW()
+				require.True(t, ok, "%s: current_power_limit", label)
+				require.NotEmpty(t, gpu.UUID(), "%s: uuid", label)
+				require.NotEmpty(t, gpu.PerformanceState(), "%s: performance_state", label)
+
+				for _, r := range comparedReadings(element) {
+					require.True(t, r.raw.present(),
+						"%s: %s is absent; the driver may have renamed it", label, r.element)
+				}
+				requireOneThresholdPresentation(t, label, element.Temperature)
+				requireOneSramRendering(t, label, element.ECCErrors)
+				requireWholeRemapBlock(t, label, element.RemappedRows)
+			}
+		})
+	}
+}
+
+// comparedReadings are the elements every board in the captures emits, whatever
+// its architecture. The architecture-dependent families are asserted separately,
+// because for those absence is the driver's way of saying "not this rendering"
+// rather than a rename.
+func comparedReadings(g gpuElement) []namedReading {
+	return []namedReading{
+		{"product_name", g.ProductName},
+		{"product_architecture", g.ProductArchitecture},
+		{"uuid", g.UUID},
+		{"board_id", g.BoardID},
+		{"c2c_mode", g.C2CMode},
+		{"fan_speed", g.FanSpeed},
+		{"performance_state", g.PerformanceState},
+		{"accounting_mode_buffer_size", g.AccountingModeBufferSize},
+
+		{"platformInfo/slot_number", g.PlatformInfo.SlotNumber},
+		{"platformInfo/tray_index", g.PlatformInfo.TrayIndex},
+		{"platformInfo/host_id", g.PlatformInfo.HostID},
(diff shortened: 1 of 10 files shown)


ArangoGutierrez (commented):
Nothing here blocks merge.

Nice set of captures, and the scrubbing holds up. I checked every serial, chassis_serial_number, uuid and clusterUuid across the seven files against the placeholders the README documents: all of them match, no process blocks, no hostnames, and the only IPv4-looking strings are VBIOS versions. The clusterUuid rule works exactly as written too, the four GB200 and four GB300 GPUs carry 00000001 while the non-fabric boards keep their real all-zero body, so nothing comes out claiming a fabric it was not in. The table checks out against the files as well: attached_gpus, per-GPU block count, product_name, driver and CUDA match every row, and the newer-driver framing holds per profile once you compare against the matching mock fixture rather than the 580.65.06 the parent README quotes. Comments below are on the pointer from the parent directory and on what the test is actually guarding.

1. The rename guard covers six readings, but schema.go decodes a good deal more, and the ones diff.go and temperature.go actually consume are mostly outside that six: utilization, the ECC aggregate counters, the temperature thresholds and platformInfo. A rename in any of those still decodes as absent and slips through here. Extending the per-GPU block to the readings the comparison code reads would make the guard match the invariant the comment claims for it. (tests/e2e/go/assertions/nvidiasmi/hardware_test.go:42)

tests/e2e/go/assertions/nvidiasmi/testdata/README.md
```diff
@@ -13,6 +13,9 @@ per-GPU indexing and override scoping.
 | `qx-gb200-lost.xml` | GPU 0 healthy, GPU 1 lost (`GPU is lost` bodies). |
 | `qx-gb200-ecc-injected.xml` | GPU 0 has a non-zero `ecc_errors/aggregate/dram_uncorrectable`. |
 
+`hardware/` holds untrimmed captures from real B200, H100 and GB300 nodes, for
```
ArangoGutierrez: hardware/ ships seven captures, not three: A100, GB200, L40S and T4 are missing from this sentence. Since it is the pointer someone follows from the parent directory, either list all seven or drop the enumeration and let the hardware/README.md table carry it.

tests/e2e/go/assertions/nvidiasmi/hardware_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+				_, ok := gpu.TemperatureC()
+				require.True(t, ok, "%s: gpu_temp", gpu.Label())
+				_, ok = gpu.SMClockMHz()
+				require.True(t, ok, "%s: sm_clock", gpu.Label())
+				_, ok = gpu.MemoryTotalMiB()
+				require.True(t, ok, "%s: fb_memory_usage/total", gpu.Label())
+				_, ok = gpu.PowerLimitW()
+				require.True(t, ok, "%s: current_power_limit", gpu.Label())
+				require.NotEmpty(t, gpu.UUID(), "%s: uuid", gpu.Label())
```
ArangoGutierrez: The scrub is the one invariant here with a real cost if it slips, and right now it rests entirely on whoever adds the next capture remembering to grep. The placeholders are regular enough to assert: serial and chassis_serial_number match 16500000000NN and 18200000000NN, uuid matches GPU-000000NN-0000-0000-0000-0000000000NN, and clusterUuid is either all-zero, N/A or 000000NN-.... A check over those four elements in this loop turns a forgotten scrub into a CI failure instead of a public commit.

roma-glushko (dismissed):

giuliocalzo:
Thanks — all three are addressed, and the scrub point turned up something bigger when I went looking.

**Unscrubbed per-die identifiers.** Auditing the scrub the way you suggested found two elements neither of us had checked: `pdi` and `gpu_fabric_guid`. Together they carried 35 distinct values naming individual dies — `pdi` on 26 GPUs across six files, `gpu_fabric_guid` on the 16 fabric-attached ones — and on the B200 the GUIDs shared a chassis-derived suffix. `pdi` sits directly beside `<serial>` and `<uuid>`, so it is the same class of identifier as the ones already scrubbed; it was simply missed. Both are placeholders now under one shared numbering, because on Grace-Blackwell the two elements report the same value and scrubbing them apart would invent a difference the node did not report. I squashed and force-pushed rather than adding a follow-up commit, so the values are out of the branch tip and the PR diff.

**The scrub is asserted now.** `TestHardwareCapturesAreScrubbed` checks all six elements against the placeholder shapes, with patterns narrow enough that a real reading cannot satisfy them (`^0x0{14}[0-9a-f]{2}$` for the two new ones). `N/A` and empty bodies are skipped, since a board that answers nothing has no identity to scrub. I confirmed it fails by putting a real PDI back into one capture.

**Reading coverage.** The guard now asserts `present()` over the elements `checks.go` and `temperature.go` compare — utilization, both ECC scopes, the platformInfo block, the PCIe maxima, virtualization, the encoder and FBC stats, `c2c_mode`, `accounting_mode_buffer_size` — instead of six readings. `present()` is the right test here, as `C2CModeProblems` already uses it: `N/A` passes, while a renamed element leaves an empty body and fails. One correction on the pointer: there is no `diff.go` in this package. The readings you meant live in `checks.go` and `platform.go`, and `temperature.go` works on the reading types rather than the `GPU` accessors.

The two architecture-dependent families needed a different assertion, since for them absence is the driver saying "not this rendering" rather than a rename. The test asserts exactly one presentation per board, and the captures confirm the split those checks assume: absolute thresholds on the pre-Ada T4 and A100, `*_tlimit_threshold` on Ada and later, the combined SRAM row only on Turing with the parity/SEC-DED pair from Ampere on, and `remapped_rows` all-or-nothing, absent entirely on the T4. That turns the premise behind #635 and the SRAM layouts into evidence rather than an assumption. Each of these fails on a deliberately renamed or dropped element; I checked all five.

**Parent README.** Dropped the enumeration and pointed at the `hardware/README.md` table, so it cannot go stale again. I also tightened the "newer driver" framing you flagged: it read as a blanket claim while the A100 at 570.148.08 is older than the 580.65.06 the parent quotes, so it now says the drivers spread either side of the one the mock targets.

Issue #384, opened 2026-06-11
docs: drift between documented examples and the current tree (2026-06-11 verification sweep)


ArangoGutierrez:
After merging #367, #382, and #383, every documented example and demo was executed against main (`b2d7084e`): quickstart options 1-3, all four demos under `docs/demo/`, the fake-gpu-operator integration guide, and all 21 example blocks in `docs/examples.md`. Everything the merged code touches works end to end (cross-node ibping/iblinkinfo, failure injection, compute-domain cliques, FGO split pools, full 7-profile e2e matrix). The sweep also surfaced doc/build drift, recorded here so it does not get lost.

## Breaks the documented flow

- `pkg/gpu/mocknvml/Makefile:17` sets `GOLANG_VERSION ?= 1.25.0` while `go.mod` requires `go >= 1.26.0`, so `make docker-build` fails exactly as documented in quickstart.md:55, development.md:87, troubleshooting.md:47, and the examples.md CI block. The deployment Dockerfile ARG was bumped to 1.26.0; this default was missed. Same class: `nvml-mock-e2e.yaml` workflow_dispatch default `golang_version: '1.25'` (run 27340427094 failed uniformly at "Build nvml-mock image").
- `docs/demo/standalone/demo.sh:204` uses `mapfile -t` (bash >= 4) and dies with exit 127 on macOS stock bash 3.2 at the cross-node ibping step. The compute-domain README states the repo convention is no bash-4 features. Steps 1-9 pass; steps 10-11 verified manually (ibping 3/3 pongs, iblinkinfo cross-pod peer reached) — the functionality is fine, only the script is.
- `docs/quickstart.md:58-60` claims `make docker-build` artifacts include `nvidia-smi`; no Makefile target produces a local nvidia-smi (only `deployments/nvml-mock/Dockerfile` extracts it into the K8s image).

## Stale or contradictory

- Helm chart README "Known Limitations" (~1034-1069) still claims PCI sysfs entries and `dra.k8s.io/pcieRoot` are missing ("see #265") — contradicted by the same README's PCIe topology section and by live verification (8 PCI symlinks, readable numa_node, 2 root complexes).
- Chart `templates/NOTES.txt:53-57` marks GPU Operator "UNTESTED — contributions welcome" — contradicts `tests/e2e/gpu-operator-values.yaml`, troubleshooting.md:349, and the root README listing GPU Operator as tested.
[...]

PR #695, opened 2026-08-21
docs: serve Mokka documentation from GitHub Pages

ArangoGutierrez:
GitHub Pages currently serves a React dashboard that tracks CI across nine
other NVIDIA repositories. It works and deploys every six hours, but it does
not cover this repository, so the front door of the Mokka repo documents nine
projects that are not Mokka. This replaces it with the Mokka documentation.

Live preview of the nav and landing page is reproducible locally with
`make docs-deps && make docs-serve`.

## What this does

- Adds an MkDocs Material site covering quickstart, architecture,
  configuration, runtime control, examples, troubleshooting, all six demos and
  the fake-gpu-operator integration, with a landing page and NVIDIA-green
  theming.
- Moves the 1424-line Helm chart guide into `docs/` so the site can publish it.
  The chart keeps a shorter README rather than a bare stub: there is no
  `.helmignore`, so `README.md` ships inside the packaged chart and
  `helm show readme` would otherwise have regressed to a pointer.
- Fixes pre-existing links that escaped the docs tree. `mkdocs build --strict`
  found ten of them, plus two anchor links into `nvml-mock-ctl.md` that were
  already broken.
- Deletes `trigger-pages-deploy.yaml`. It ran every six hours and would have
  restored the dashboard over the docs within six hours of the cutover.

## Anchors

The docs were written to render on GitHub, so anchors inside them use GitHub's
slug scheme. The site uses pymdownx's GitHub-compatible slugifier so existing
intra-document anchor links keep working.

## Internal plans

`exclude_docs` drops `docs/plans/` and `docs/superpowers/`. Both are gitignored
scratch directories, so they never reach CI and grepping the built site for
them would pass unconditionally. `make docs-check-exclusion` instead plants a
file and proves the exclusion drops it. Verified both ways: with `exclude_docs`
the canary is absent, without it the file is published and the target exits 2.
`--strict` alone does not catch this, because MkDocs reports an un-navigated
page as INFO rather than a warning.

## The dashboard

The `gh-pages` branch is untouched, so the dashboard and its history stay
recoverable from git. Accumulated `history.json` trend data lived only on the
deployed site; it has been archived locally before the cutover.

## Verification

```
make docs          # exclusion check + strict build, rc=0, 18 pages
make helm-tests    # 169 passed, 15 snapshots
actionlint         # clean
```

CI calls the same make targets, so a failure reproduces locally with one
command, per AGENTS.md.

## Known gaps, not addressed here

- Nine of ten `cmd/` binaries have no documentation.
- MEPs under `enhancements/` are not published.
- Existing doc drift (#384) is not audited. One self-contradicting profile
  count was corrected in passing.


giuliocalzo (approved):

docs/stylesheets/extra.css
```diff
@@ -0,0 +1,99 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* NVIDIA green. #76B900 is the brand value; the darker tone is used for
+   hover and for text on light backgrounds, where #76B900 fails contrast. */
+:root {
```
roma-glushko, line 19: Would be nice to see how it looks like. Can you attach a screenshot to the PR description?

I assume this is NVIDIA color styling.

docs/helm-chart.md
```diff
@@ -0,0 +1,1424 @@
+# nvml-mock Helm Chart
+
+Mock GPU infrastructure for Kubernetes testing. Turns any cluster into a
+multi-GPU environment using a CGo-based mock NVML library — no physical
+NVIDIA hardware required.
+
+## What It Does
+
+Deploys a DaemonSet that creates on every node:
```
roma-glushko, line 9: I'm not sure we need to expose these details to our end users. For this place, I would briefly list subsystems that we mocks. Other places of the documentation can go deeper into details if appropriate.

docs/helm-chart.md
```diff
@@ -0,0 +1,1424 @@
...
+  PCI sysfs — `lspci` and anything else reaching it through libc — resolve the
+  PCIe root complex via a standard `readlink()`. The NVIDIA DRA driver is a Go
+  binary and does not see this tree, so `dra.k8s.io/pcieRoot` is still absent
+  from its ResourceSlices; see [Known Limitations](#known-limitations) and
+  issue [#265](https://github.com/NVIDIA/k8s-test-infra/issues/265)
+
+Consumers (DRA driver, device plugin) point at `/var/lib/nvml-mock/driver`
+as the NVIDIA driver root and discover GPUs through standard NVML APIs.
+
+When `nri.enabled=true` (opt-in; default `false`), the chart also deploys
```
roma-glushko, line 32: The `nri.enabled=false` state doesn't make sense to me as an end user. What am I supposed to do in that case? Go and add LD_PRELOAD to all my workloads. That will never happen. 

We should remove this state at all if we are that NRI is the way we insert the needed env vars across the cluster.

docs/helm-chart.md
```diff
@@ -0,0 +1,1424 @@
...
+user runs. Nothing reports this: the DaemonSet is Ready, `/readyz` returns 200
+because the plugin *is* registered, and skipped containers produce no log line
+at any level. The pods simply start with no mock GPU.
+
+## Prerequisites
+
+| Tool | Version | Required For |
+|------|---------|-------------|
+| [Docker](https://docs.docker.com/get-docker/) | 20.10+ | Building the image |
+| [Kind](https://kind.sigs.k8s.io/) | 0.20+ | Local cluster (or use your own) |
```
roma-glushko, line 66: We should not focus docs on kind-first experience. Kind and NCPs should be two first-class installation paths, so it makes sense to separate those not to make an impression kind is a requirement for AWS EKS deployment, for example.

docs/helm-chart.md
```diff
@@ -0,0 +1,1424 @@
...
+
+Expected: worker with `a100` profile shows `4` GPUs, worker with `t4` profile shows `2` GPUs.
+
+### 7. Clean up
+
+```bash
+kind delete cluster --name gpu-fleet
+```
+
+## Integration: fake-gpu-operator
```
roma-glushko, line 490: This is the longest helm chart README.md I have ever seen and it's because it focuses on the wrong details. These installation guides are great to have in docs as how-to/guides. 

The helm chart readme though should focus on describing helm chart values knobs, not copy-paste our entire documentation. Whoever wants to learn more should go to our docs.

docs/index.md
```diff
@@ -0,0 +1,106 @@
...
+
+    [Get started](quickstart.md)
+
+-   **Architecture**
+
+    How the mock NVML library, the CGo bridge and the GPU profiles fit together.
+
+    [Read the design](architecture.md)
+
+-   **Demos**
```
roma-glushko, line 28: I think it's more common to see this called guides/how-to articles, not demo. Worth considering to rename it.

docs/index.md
```diff
@@ -0,0 +1,106 @@
...
+
+    [See the reference](configuration.md)
+
+</div>
+
+## Try it
+
+```bash
+# 1. Create a cluster
+kind create cluster --name mokka
```
roma-glushko, line 47: `kind create cluster --name mokka` is not sufficient for mokka to be deployed well I think. This is why we have `make cluster-create` command. We should reflect that here.

A bigger question is what we expect the end user to be starting with here? Are they git clone mokka repository or we expect them to just have kind & helm installed?

In the first scenario, we could describe installation process as:

```bash
make cluster-create
tilt up -- --gpu-operator
```

If we target the second scenario, then we are in trouble because `kind create cluster --name mokka` won't give you a configured cluster for mokka.

docs/index.md
```diff
@@ -0,0 +1,106 @@
...
+| NVIDIA Device Plugin | `nvidia.com/gpu` extended resources | Tested |
+| NVIDIA DRA Driver | Dynamic Resource Allocation | Tested |
+| NVIDIA GPU Operator | Full stack device plugin, GFD and validator | Tested |
+| GPU Feature Discovery | Node labeling from NVML | Tested |
+
+## Integrations
+
+| Integration | Description |
+|-------------|-------------|
+| [fake-gpu-operator](integrations/fake-gpu-operator.md) | Run:ai's K8s-level GPU simulation plus nvml-mock driver fidelity |
```
roma-glushko, line 106: Does it make sense to separate this one from Tested Consumers? It's kinda a test consumer too.

Makefile
```diff
@@ -333,3 +333,39 @@ e2e-nri: ## e2e — NRI ambient-injection scenario
...
+
+# The docs site is MkDocs Material. CI calls these same targets so a failure
+# reproduces locally with one command.
+PYTHON ?= python3
+MKDOCS ?= mkdocs
+
+.PHONY: docs-deps docs-build docs-serve docs-check-exclusion docs
+
+docs-deps: ## Install the pinned MkDocs toolchain
+	$(PYTHON) -m pip install -r requirements-docs.txt
```
roma-glushko, line 347: We should use `uv` for deps management.

PR #699, opened 2026-08-21
feat(helm): default gpu.profile to gb300

giuliocalzo:
## What This PR Does

Flips the nvml-mock chart default `gpu.profile` from `a100` to `gb300`, so a bare `helm install` simulates a Grace-Blackwell Ultra fleet. `--set gpu.profile=a100` restores the previous behaviour.

Three chart tests silently relied on `a100` being the default and now pin the profile explicitly: the non-Blackwell `DRIVER_VERSION` check in the DaemonSet suite, and the two ConfigMap dynamic-metrics cases whose expectations only hold for a profile that has no power base of its own. The snapshot suites swap roles — `gb300` covers the default case, `a100` gets its own explicit case — so per-profile coverage is unchanged.

The quick start also loses the kind image preload, in the root README, in `docs/index.md` and in the packaged chart README, which each presented it as mandatory. The multi-arch `kind load docker-image` failure those blocks cite is a limitation of preloading, not a prerequisite for installing: the published image is an OCI index with `linux/amd64` and `linux/arm64` manifests, so a node's containerd resolves its own platform when it pulls directly. `docs/troubleshooting.md` keeps the `docker save --platform` workaround for anyone who does preload.

Finally, the quick start installs release `nvml-mock` into the default namespace instead of `mokka` into a `mokka` namespace. Release name feeds the chart's fullname template, so `mokka` renders `mokka-nvml-mock` and the `daemonset/nvml-mock` commands in the consumer guides this page links to would not have resolved.

## Why

Ampere is the architecture least likely to expose a gap in software being tested against a mock fleet. GB300 exercises what consumers are actually being ported to: NVLink-C2C to Grace, FP4/FP6, NVLink v5, PCIe Gen6, and the 570 driver line. The default should be the hardware people are porting to, not the one their code already handles.

## Checklist

- [x] Tests pass (`go test -v -race ./...`) — plus `make helm-tests`: 170 passed, 15 snapshots, and `make docs-build` (strict)
- [x] Linter passes (`make lint-fix`) — `golangci-lint` clean; `govulncheck` flags pre-existing stdlib CVEs fixed in go1.26.6, unrelated to this change (no Go files touched)
- [x] New code has SPDX license headers — n/a, no new files
- [x] Documentation updated (if applicable) — root README, `docs/index.md`, `docs/helm-chart.md` values table and profile guidance, and the packaged chart README

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,21 +24,15 @@ Turn any Kubernetes cluster into a multi-GPU environment for testing.
 No physical NVIDIA hardware required.
 
 ```bash
-# 1. Create cluster
 kind create cluster --name mokka
 
-# 2. Load the published image (or build locally with: docker build -t nvml-mock:local -f deployments/nvml-mock/Dockerfile .)
-# The published image is multi-arch. `kind load docker-image` cannot load a
-# multi-arch image from Docker Desktop's containerd store and fails with
-# "ctr: content digest ...: not found", so save one platform first.
-ARCH=$(uname -m | sed 's/x86_64/amd64/; s/aarch64/arm64/')
-docker pull ghcr.io/nvidia/nvml-mock:latest
-docker save --platform "linux/${ARCH}" ghcr.io/nvidia/nvml-mock:latest -o nvml-mock.tar
-kind load image-archive nvml-mock.tar --name mokka
-
-# 3. Install
-helm install nvml-mock oci://ghcr.io/nvidia/k8s-test-infra/chart/nvml-mock
+helm install mokka oci://ghcr.io/nvidia/k8s-test-infra/chart/nvml-mock \
+    --namespace mokka --create-namespace \
+    --set gpu.profile=gb300
 ```
+
+Every node now reports 8 mock GB300 GPUs. `gb300` is the chart default; swap in
+`a100`, `h100`, `b200`, `gb200`, `l40s`, or `t4` for other hardware.
 
 After install, deploy a consumer to test:
 
@@ -65,11 +59,6 @@ node-wide NRI, and NFD label-provenance coverage. Run manually via
 | **Multi-Node Fleet** | Heterogeneous A100/T4 workers, mock files, InfiniBand behavior, device plugin resources, and GPU workload scheduling | Fixed multi-node topology |
 | **Node-Wide NRI Injection** | Ambient mock GPU injection into ordinary pods without GPU requests or hostPath mounts | Workflow-selected profiles |
 | **NFD Label Provenance** | That NFD creates `feature.node.kubernetes.io/pci-10de.present` from the feature file nvml-mock writes, and that nvml-mock does not write the label itself | Pinned to `a100` — the label is vendor-only and byte-identical across profiles |
-
-Manual dispatch accepts a JSON array of GPU profiles; local runs default to
-`gb200`.
-
-See [`.github/workflows/nvml-mock-e2e-go.yaml`](.github/workflows/nvml-mock-e2e-go.yaml) for details.
 
 ## Mock NVML Library
 
--- a/README.md
+++ b/README.md
@@ -26,8 +26,7 @@ No physical NVIDIA hardware required.
 ```bash
 kind create cluster --name mokka
 
-helm install mokka oci://ghcr.io/nvidia/k8s-test-infra/chart/nvml-mock \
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -24,31 +24,25 @@ Turn any Kubernetes cluster into a multi-GPU environment for testing.
...
-# multi-arch image from Docker Desktop's containerd store and fails with
-# "ctr: content digest ...: not found", so save one platform first.
-ARCH=$(uname -m | sed 's/x86_64/amd64/; s/aarch64/arm64/')
-docker pull ghcr.io/nvidia/nvml-mock:latest
-docker save --platform "linux/${ARCH}" ghcr.io/nvidia/nvml-mock:latest -o nvml-mock.tar
-kind load image-archive nvml-mock.tar --name mokka
-
-# 3. Install
-helm install nvml-mock oci://ghcr.io/nvidia/k8s-test-infra/chart/nvml-mock
+helm install mokka oci://ghcr.io/nvidia/k8s-test-infra/chart/nvml-mock \
```
ArangoGutierrez, line 29: Dropping the preload leaves this at odds with the docs site: `docs/index.md` still walks through `docker pull` plus `docker save --platform` before installing, on the grounds that `kind load docker-image` fails against Docker Desktop's containerd store for a multi-arch image. If the node really can pull unaided now, `docs/index.md` should lose that block in the same change, otherwise the front page and the README disagree about whether the step is needed. The release name and namespace also diverge here (`mokka` in `mokka` vs `nvml-mock` in default), which is worth aligning while you are in there.

ArangoGutierrez (commented):
Nothing here blocks merge.

The profile swap itself holds up. The three chart tests that were implicitly relying on `a100` now pin it explicitly, and the snapshot suites trade places rather than dropping a case, so per-profile coverage is unchanged. The reasoning for `gb300` in the description is convincing: defaulting to the hardware people are porting to surfaces more than defaulting to the one their code already handles.

The thing to sort out is outside the diff, and it is my doing rather than yours. This branch predates #695, which moved the chart README to `docs/helm-chart.md` and left an 84-line file in its place. Your three edits land at lines 924 and 1068-1078 of the old 1424-line version, so a plain rebase drops you into resolving a file that was effectively replaced. Both the parameter table and the "When to Use Each Profile" list live in `docs/helm-chart.md` now, and that is where those hunks want to go. Sorry for the churn.

README.md
giuliocalzo: Both fixed in ffe61c75.

On the preload: agreed the two pages should not disagree, and I dropped the block from `docs/index.md` rather than restoring it in the README. The multi-arch failure you're pointing at is real, but it's a limitation of `kind load docker-image` against Docker Desktop's containerd store — not a prerequisite for installing. `ghcr.io/nvidia/nvml-mock:latest` is an OCI index with `linux/amd64` and `linux/arm64` manifests, so the node's containerd resolves its own platform when it pulls directly, and nothing goes through Docker Desktop. `docs/troubleshooting.md` keeps the `docker save --platform` workaround for anyone who does choose to preload, which seems like the right home for it now that no happy path needs it. The packaged chart README repeated the same claim after #695, so that copy is gone too — all three entry points now install directly.

On the naming: you were right that this was more than cosmetic. Release name feeds the fullname template, so `mokka` renders `mokka-nvml-mock` and the `kubectl rollout status daemonset/nvml-mock` in the guides I link to would never have resolved, namespace aside. The quick start now installs `nvml-mock` into the default namespace, matching the guides and the docs site.

I did look at going the other way — `mokka` everywhere — but it reaches about 78 `kubectl` references across the docs plus the runnable demo scripts, the Tiltfile and `scenario_nri_test.go`, which felt like its own PR rather than something to fold into a default-profile change. Happy to open one if you think the branding is worth it.

`make docs-build` (strict) and `make helm-tests` (170 tests) both pass.

giuliocalzo:
@ArangoGutierrez both of your points are addressed in ffe61c75, and the branch is rebased on `main` as of #702.

The preload block is gone from `docs/index.md` and the packaged chart README as well as the quick start, since the multi-arch failure it cited is a `kind load` limitation rather than an install prerequisite; `docs/troubleshooting.md` keeps the workaround. The quick start now installs `nvml-mock` into the default namespace, so the `daemonset/nvml-mock` commands in the linked guides resolve. Ready for another look.

PR #700, opened 2026-08-21
docs(tools): add reference pages for the cmd/ binaries

### LINKED ISSUE #697: docs: document the cmd/ binaries with no user-facing guide
`cmd/` ships ten binaries. Only `nvml-mock-ctl` has a documentation page.

Undocumented: `mokka-control-plane`, `nvml-mock-nri`, `mock-ib`,
`fake-fabricmanager`, `fake-imex`, `check-fabric`, `generate-bridge`,
`imex-nogpu-shim`, `render-imex-procdevices`, `render-pci-sysfs`.

Now that the documentation is published at
https://nvidia.github.io/k8s-test-infra/ the gap is user-visible rather than an
internal inconsistency. `mokka-control-plane` is likely the highest value given
it is the newest surface.

Adding a page means creating `docs/<name>.md` and a nav entry in `mkdocs.yml`.
`make docs` fails on a broken link or a nav entry pointing at a missing file.

ArangoGutierrez:
Closes #697.

`cmd/` ships thirteen binaries across eleven directories. Only `nvml-mock-ctl`
had a documentation page. This adds one page per tool, plus an index, and wires
them into the nav under Tools.

Every flag, default and environment variable was read out of the Go source
rather than inferred, and a review pass re-derived them a second time. Three
claims from the first pass were corrected that way: `start-mock-ib.sh` appends
`-fabric` only when `MOCK_IB_PING_FABRIC=1`, the Makefile target is `gen` rather
than `generate`, and the control-plane health handlers use `Fprintln`, so the
body carries a trailing newline.

The second commit exists because #687 landed mid-flight and renamed
`cmd/mokka-control-plane` to `cmd/control-plane`. Re-deriving that page against
the new tree surfaced three further errors: `--log-level` reads
`MOKKA_LOG_LEVEL` rather than `MOKKA_CP_LOG_LEVEL`, `--log-format` was missing
altogether, and the accepted level list omitted the `warning` alias.

Where the chart and the source disagree the pages say so. `--shutdown-timeout`,
`--cdi-device-name`, the two topology flags and `--ld-preload-shims` are never
templated, so deployed pods run them at compiled-in defaults.

Known gap, not addressed here: these pages are prose describing flags that live
in Go source, with no executable check tying the two together. A rename will
desync them silently. A golden-flags test would be the durable fix.

`make docs` passes.


docs/tools/render-pci-sysfs.md
roma-glushko, line 1: This one is removed already ✅

docs/tools/render-imex-procdevices.md
roma-glushko, line 1: This one is removed already ✅

docs/tools/mock-ib.md
roma-glushko, line 1: This one is being removed in my current PR ✅

docs/tools/fake-imex.md
roma-glushko, line 1: This one is removed already ✅

docs/tools/fake-fabricmanager.md
roma-glushko, line 1: This one is gonna be removed soon ✅

docs/tools/imex-nogpu-shim.md
roma-glushko, line 1: This one was moved to shims ✅

mkdocs.yml
```diff
@@ -111,6 +111,18 @@ nav:
       - Examples: examples.md
       - Runtime Control: nvml-mock-ctl.md
       - Troubleshooting: troubleshooting.md
+  - Tools:
```
roma-glushko: May be a good idea to include shims, but I would not mix them with the main CLIs.

roma-glushko: Would Components be a better name? Tools sound like a miscellaneous stuff, but we talk about the main CLI Mokka consists of.

docs/tools/generate-bridge.md
roma-glushko, line 1: I hope to pull out nvml mock into shims and put this CLIs there as it's a misc tool to get stubs, not a main component of the system.

giuliocalzo (changes requested):
most of the cli are not present anymore

roma-glushko (approved):

giuliocalzo (approved):

PR #701, opened 2026-08-21
docs: publish enhancement proposals on the documentation site

### LINKED ISSUE #698: docs(pages): publish the enhancement proposals on the documentation site
`enhancements/meps/` holds two MEPs with architecture diagrams:
`0001-mokka-control-plane` and `0002-device-plugin-nri-composition`. They are
referenced from `docs/helm-chart.md`, but only as links out to GitHub, because
MkDocs can publish only files under `docs/`.

They were deliberately left out of #695 to keep that change scoped to
publishing what already lived in `docs/`.

Two ways to fix it:

1. Move them under `docs/` and add nav entries.
2. Add a build step that copies `enhancements/meps/**` into the docs tree
   before `mkdocs build`.

The copy approach keeps the MEP process where contributors expect to find it.

ArangoGutierrez:
Closes #698.

`enhancements/meps/` holds two proposals with architecture diagrams that were
only reachable as links out to GitHub, because MkDocs publishes nothing outside
`docs_dir`.

Rather than move them, `hack/sync-enhancement-docs.sh` stages them into the
docs tree at build time. That keeps the MEP process where contributors expect
it: a new MEP lands in `enhancements/` as it always has and appears on the site
with no second location to remember. The staged copy is gitignored and rebuilt
from scratch each run, so a stale copy cannot survive.

Every target that runs mkdocs depends on the sync, so `make docs`, `make
docs-serve` and the exclusion guard all see the same tree, and CI needs no
special-casing.

One deviation from the issue as filed: `deploy-pages.yaml` does change. Its path
filters had to learn about `enhancements/**` and the sync script, because
without that a MEP edit would not rebuild the site and the published copy would
drift from the source.

`docs/helm-chart.md` previously linked MEP-0002 by absolute GitHub URL; it now
points at the in-site page.

`make docs` passes, and the MEP images are emitted into the built site rather
than only the pages that reference them.

### DIFF
--- a/.github/workflows/deploy-pages.yaml
+++ b/.github/workflows/deploy-pages.yaml
@@ -51,12 +51,14 @@ jobs:
       - name: Check out code
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
-      - name: Set up Python
-        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+      # uv both installs the pinned toolchain and provisions the interpreter,
+      # so no separate Python setup step is needed. `make docs-deps` pins the
+      # version via DOCS_PYTHON_VERSION.
+      - name: Set up uv
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
         with:
-          python-version: "3.12"
-          cache: pip
-          cache-dependency-path: requirements-docs.txt
+          enable-cache: true
+          cache-dependency-glob: requirements-docs.txt
 
       - name: Install documentation dependencies
         run: make docs-deps
--- a/.github/workflows/deploy-pages.yaml
+++ b/.github/workflows/deploy-pages.yaml
@@ -14,18 +14,25 @@
 
 name: Deploy documentation to Pages
 
+# The MEPs are published from enhancements/, which `make docs` stages into the
+# docs tree. Without it in the filters a MEP-only change would build nothing and
+# never reach the site.
 on:
   push:
     branches:
       - main
     paths:
       - "docs/**"
+      - "enhancements/**"
+      - "hack/sync-enhancement-docs.sh"
       - "mkdocs.yml"
       - "requirements-docs.txt"
       - ".github/workflows/deploy-pages.yaml"
   pull_request:
     paths:
       - "docs/**"
+      - "enhancements/**"
+      - "hack/sync-enhancement-docs.sh"
       - "mkdocs.yml"
       - "requirements-docs.txt"
       - ".github/workflows/deploy-pages.yaml"
--- a/.github/workflows/deploy-pages.yaml
+++ b/.github/workflows/deploy-pages.yaml
@@ -14,9 +14,8 @@
 
 name: Deploy documentation to Pages
 
-# The MEPs are published from enhancements/, which `make docs` stages into the
-# docs tree. Without it in the filters a MEP-only change would build nothing and
-# never reach the site.
+# make docs stages enhancements/ into the docs tree, so MEP-only edits need a
+# rebuild too.
 on:
   push:
     branches:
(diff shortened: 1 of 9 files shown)


.github/workflows/deploy-pages.yaml
```diff
@@ -14,18 +14,25 @@
 
 name: Deploy documentation to Pages
 
+# The MEPs are published from enhancements/, which `make docs` stages into the
```
roma-glushko: The comment is redundant for long-term readers, it's the agent trying to document what it's done for the specific task/chat.

.github/workflows/deploy-pages.yaml
```diff
@@ -14,18 +14,25 @@
...
+      - "enhancements/**"
+      - "hack/sync-enhancement-docs.sh"
       - "mkdocs.yml"
       - "requirements-docs.txt"
       - ".github/workflows/deploy-pages.yaml"
   pull_request:
     paths:
       - "docs/**"
+      - "enhancements/**"
+      - "hack/sync-enhancement-docs.sh"
```
roma-glushko, line 34: I know ppl create these hack dirs but it's not it better to name it "tools"? 😃

mkdocs.yml
```diff
@@ -121,5 +121,14 @@ nav:
       - NVSentinel: demo/nv-sentinel/README.md
   - Integrations:
       - fake-gpu-operator: integrations/fake-gpu-operator.md
+  # These pages are a build-time copy of enhancements/ staged by
+  # hack/sync-enhancement-docs.sh; docs/enhancements/ is gitignored. Adding a MEP
+  # means adding it to enhancements/meps/ and listing it here.
```
roma-glushko, line 126: Would this be easier to move enhancements under docs? I feel like it may save us this back-and-forth with coding docs. We can make it a requirement to register new MEPs in this table of content.

.github/workflows/deploy-pages.yaml
ArangoGutierrez: Fair, that was narrating the change rather than the code. Trimmed it to the part a later reader actually needs, which is that `enhancements/` sits outside `docs_dir` so the filter only works because `make docs` stages it. Pushed as e78ad8bc.

.github/workflows/deploy-pages.yaml
ArangoGutierrez: I'd keep `hack/` here. `hack/golang-version.sh` is already in there and both `variables.yaml` and `nvml-mock-e2e-go.yaml` call it, and there's no `tools/` yet, so renaming just this one would leave us with two script directories. Happy to move both in a separate change if you'd rather standardise on `tools/`.

mkdocs.yml
ArangoGutierrez: Your instinct about the cost is right, though I think the script buys less than it looks like it does.

Registration is manual either way. The nav above lists every MEP by hand, and moving the directory would not remove that step, so the sync is not saving us the back-and-forth you are thinking of. What it actually buys is only that MEPs stay at `enhancements/`, next to the README that documents the process and where someone browsing the repo expects to find them. The price is the script plus the Makefile wiring and two path filters.

Keeping it on that basis for now, but it is a close call and I will move it if you would rather have the ~120 fewer lines. The thing that would flip me is wanting MEPs to use the rest of the docs tooling, snippets and cross-page links, since that is markedly simpler inside `docs_dir`.

giuliocalzo (commented):
do we really need to publish the MEP inside the web doc?

mkdocs.yml
ArangoGutierrez: I looked at this again after reading your #805 page, and I think the two are complementary rather than duplicated, so I have kept the directory where it is and fixed the collision instead.

`docs/contributing/enhancements.md` in #805 describes the process: when to write a MEP, how to number it, how to keep it honest. It points at `enhancements/template.md`. This PR publishes the proposals themselves, which are already on main under `enhancements/meps/`. Moving that directory would break the `enhancements/template.md` path your own page links to.

The real problem was that both would have produced a nav entry called "Enhancement Proposals". So in 0f207335 the MEP index is now nested under Contributing as "Proposals (MEPs)", named for the documents rather than the process. Your page can keep the name it has.

On the sync script: nothing is duplicated in git. `docs/enhancements/` is gitignored and staged at build time, so the originals in `enhancements/` stay the single source. Happy to revisit if you would still rather have the ~120 fewer lines once #805 lands.

roma-glushko (approved):

ArangoGutierrez:
I am not sure about this one

PR #702, opened 2026-08-21
docs: unpublish the CUDA mock page

ArangoGutierrez:
CUDA mocking is not usable, and the site was presenting it as though it were.

Fourteen of the mock's fifteen exports are CUDA Runtime entry points and only
`cuInit` belongs to the driver API. A real workload such as
`cuda-sample:vectoradd` links cudart statically and reaches CUDA through
`dlopen` plus `cuGetProcAddress`, so it resolves 450 driver entry points, finds
one, and fails identically whether the mock is mounted or absent.

The page said all of this plainly. The problem was the framing around it: a
Mock CUDA row in the landing page components table alongside Production
entries, and a nav entry sitting beside the architecture overview. Both read as
a shipped capability. This removes them, along with the inbound links from the
quickstart, the architecture page and the top-level README.

The page itself stays in the repository. It carries the measured evidence a
contributor needs, and republishing is one deleted line in `exclude_docs` once
the driver surface exists.

Because the page is excluded from the build, `mkdocs build --strict` fails on
any inbound link left behind, so the build itself proves the removal is
complete.

Not fixed here, since the page is no longer published: it claims the
`libcudart.so.12` symlink lets vectoradd resolve to the mock, which contradicts
the measurement in its own status block. Worth a follow-up that rewrites the
page for a contributor audience.

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,7 +82,6 @@ Use standalone for local development and CI pipelines.
 | [Quick Start](docs/quickstart.md) | Build and run in 5 minutes |
 | [Configuration](docs/configuration.md) | YAML configuration reference |
 | [Architecture](docs/architecture.md) | System design and components |
-| [CUDA Mock](docs/cuda-mock.md) | Mock CUDA library overview |
 | [Development](docs/development.md) | Contributing and extending the library |
 | [Examples](docs/examples.md) | Usage patterns and scenarios |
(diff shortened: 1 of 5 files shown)


roma-glushko (approved):
I would go even further and remove it from the actual dockerfile/image we ship until the plan around CUDA is sorted out.

PR #612, opened 2026-08-06
refactor(e2e): Tilt for setting up E2E envs

roma-glushko:
## What This PR Does

Expanding usage of Tilt by leveraging it for setting up all E2E environments. 

Cleaning up E2E test harness responsible for cluster setup and deployment.

## Why

Unification of local, E2E and demo environments.

## Checklist

- [x] Commits are signed off (`git commit -s`)
- [x] Tests pass (`go test -v -race ./...`)
- [x] Linter passes (`golangci-lint run`)
- [ ] New code has SPDX license headers
- [ ] Documentation updated (if applicable)
- [ ] CHANGELOG.md updated (if user-facing change)

PR #704, opened 2026-08-21
docs: split the quickstart into two install paths and manage docs deps with uv

ArangoGutierrez:
Addresses the review comments on #695.

## Quickstart accuracy

The review said `kind create cluster --name mokka` is not sufficient, and asked
the sharper question underneath it: do we expect the reader to have cloned the
repository, or only to have kind and helm?

Reading the source says the answer is "both, and they get different things", so
the page now says that instead of picking one. Path 1 is kind plus Helm with
nothing cloned; Path 2 is `make cluster-create` plus tilt. Each lists what it
gives you and what it does not, and the comparison table says plainly which one
CI gates.

Worth being precise about one thing, because it changes what needed fixing:
plain kind is not broken. The chart tolerates every taint so it schedules on a
single control-plane node, and `setup.sh` stages the whole mock driver root,
`nvidia-smi`, the device nodes, PCI sysfs and the IB mock. The device plugin
works there too, because `tests/e2e/device-plugin-mock.yaml` reads the mock
through a hostPath and passes device specs to kubelet itself. That path was
CI-gated green on stock `kindest/node` before #612.

What plain kind genuinely cannot do is CDI, the GPU Operator, DRA, NRI and
multi-node, because those need the nvidia-container-toolkit, the containerd
runtime handler, the DRA feature gate and the NRI socket that
`deployments/kind-nvidia-cdi` and `local/kind/*.kind.yaml` provide. So the fix
is a split, not a replacement.

## Tested consumers

fake-gpu-operator was in its own table, separate from Tested Consumers. It is
now in the same table with a column marking it a third-party integration rather
than an NVIDIA component.

## uv for the docs toolchain

`make docs-deps` now installs through uv, and the workflow uses the uv action
in place of setup-python. `requirements-docs.txt` stays as the pinned manifest
so Dependabot keeps tracking it.

The install goes into a project-local `.venv-docs` rather than `--system`.
`uv pip install` refuses a non-virtual environment, and `--system` resolves
whichever interpreter it finds first, which is an Xcode python3.9 on macOS and
trips Debian's externally-managed marker on the runner. The build targets put
that venv ahead of PATH, and the entry is inert when the venv is absent, so
anyone who already has mkdocs on PATH is unaffected.

Missing uv is a hard failure with install instructions rather than a pip
fallback. A fallback would let CI and a developer resolve different
interpreters while both report success, which is the divergence a pinned
manifest exists to prevent.

## Not addressed here

The nav restructure, the helm-chart split and the architecture rewrite are
larger and are being handled separately. `mkdocs.yml` is deliberately untouched
so this does not collide with them.

### DIFF
--- a/.github/workflows/deploy-pages.yaml
+++ b/.github/workflows/deploy-pages.yaml
@@ -51,12 +51,14 @@ jobs:
       - name: Check out code
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
-      - name: Set up Python
-        uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+      # uv both installs the pinned toolchain and provisions the interpreter,
+      # so no separate Python setup step is needed. `make docs-deps` pins the
+      # version via DOCS_PYTHON_VERSION.
+      - name: Set up uv
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
         with:
-          python-version: "3.12"
-          cache: pip
-          cache-dependency-path: requirements-docs.txt
+          enable-cache: true
+          cache-dependency-glob: requirements-docs.txt
 
       - name: Install documentation dependencies
         run: make docs-deps
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ tmp
 
 # MkDocs build output
 site/
+
+# MkDocs toolchain venv created by `make docs-deps`
+.venv-docs/
--- a/Makefile
+++ b/Makefile
@@ -340,19 +340,48 @@ e2e-nfd: ## e2e — NFD label-provenance scenario (pinned to a100)
 
 # The docs site is MkDocs Material. CI calls these same targets so a failure
 # reproduces locally with one command.
-PYTHON ?= python3
 MKDOCS ?= mkdocs
+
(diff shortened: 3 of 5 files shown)


ArangoGutierrez:
@giuliocalzo rebased onto main, conflicts solved.

The conflict was in `docs/index.md`: main had simplified the quickstart install block while this branch split the section into two install paths. The merge keeps main's simpler `helm install` inside Path 1, and main's `make cluster-create` note is dropped only because Path 2 below it documents the same thing in more depth.

One fix on top of the rebase. Running the new target twice showed `make docs-deps` is not idempotent: `uv venv` refuses to overwrite an existing environment, so it succeeded once and failed on every re-run, and would fail on the first run for anyone who already had a `.venv-docs`. It now passes `--clear`.

Worth flagging that this PR fixes a real breakage beyond the docs restructure: `make docs` cannot run on main today. `docs-deps` uses `$(PYTHON) -m pip` with `PYTHON ?= python3`, which on macOS is Xcode's Python 3.9.6, and `pymdown-extensions==11.0.1` needs >=3.10, so pip silently filters it out. Pinning the venv to 3.12 with uv is what makes `make docs` build again.

Ready when you are.

giuliocalzo (changes requested):
LGTM can you please follow greptile suggestion?

roma-glushko (dismissed):

roma-glushko (approved):

giuliocalzo (approved):

Issue #680, opened 2026-08-18
[Feature]: wire the seven remaining Grace-Blackwell nvidia-smi -q surfaces that are still stubs or unhandled field IDs


giuliocalzo:
## Problem statement

Seven `nvidia-smi -q` fields that a real GB300 answers are still `N/A` on the `gb300` profile. Individually each is small; together they are the long tail of the Grace-Blackwell surface, and one of them is reached by a stub that `nvidia-smi` calls once per GPU on every single `-q` run — so the mock is being asked for it constantly and declining every time.

None of them needs new simulation logic in the way SRAM ECC (#641) or fabric health (#677) do. They are a config value and an export away, which makes them the cheapest remaining fidelity per unit of work in the audit.

Found by diffing `nvidia-smi -q` from a real 4-GPU GB300 tray (driver `580.173.02`, CUDA 13.0) against the `gb300` profile running in kind, extending the audit behind #644 to Grace-Blackwell.

> Originally filed covering twelve fields. Three surfaces have been split out because each needs a decision or an investigation that this batch should not block on: Conf Compute protected memory, Max Customer Boost Clocks, and GPU Part Number. See the linked issues under _Non-goals_.

### The fields

| `-q` field | Real GB300 | Reached by | Status |
| --- | --- | --- | --- |
| `Capabilities` → `EGM` | `disabled` | `nvmlDeviceGetCapabilities` | generated stub, called 8× per run |
| `Inforom BBX Object Flush` → Latest Timestamp, Latest Duration | `2026/08/17 14:30:48.213`, `21946 us` | `nvmlDeviceGetLastBBXFlushTime` | generated stub, called 8× per run |
| `PCI` → `Atomic Caps Inbound` | `FETCHADD_32 FETCHADD_64 SWAP_32 SWAP_64 CAS_32 CAS_64` | `NVML_FI_DEV_PCIE_INBOUND_ATOMICS_MASK` (229) | field ID unhandled |
| `GPU Recovery Action` | `None` | `NVML_FI_DEV_GET_GPU_RECOVERY_ACTION` (230) | field ID unhandled |
| `Power Smoothing` | `Insufficient Permissions` | `NVML_FI_PWR_SMOOTHING_ENABLED` (251), `NVML_FI_PWR_SMOOTHING_PRIV_LVL` (252) | field IDs unhandled |
[...]

Issue #377, opened 2026-05-29
feat(nvml-mock): Confidential Computing (CC) mode + attestation


giuliocalzo:
**Priority:** Low

**Context.** Hopper/Blackwell CC mode and attestation are part of secure GPU deployments.

**Gap.** No CC-mode state, encrypted-NVLink modeling, or attestation report surface.

**Proposed scope.**
- Mock `nvmlDeviceGetConfComputeGpusReadyState` / CC settings / attestation-report getters returning configurable values.
- E2E: a CC-aware probe reads CC enabled/ready state from the mock.

**Why.** Enables CC-stack test paths; niche today, hence Low.

PR #714, opened 2026-08-24
feat(mocknvml): report Conf Compute protected memory for nvidia-smi -q (#711)

### LINKED ISSUE #711: [Feature]: gb300 reports N/A for Conf Compute Protected Memory Usage, and the CC gate needs a capability-versus-mode decision
## Problem statement

On a real 4-GPU GB300 tray (driver `580.173.02`, CUDA 13.0), `nvidia-smi -q` reports `Conf Compute Protected Memory Usage` → Total, Used, Free as `0 MiB`. On the `gb300` profile all three are `N/A`, because both exports behind them are generated stubs that `nvidia-smi` calls once per GPU on every `-q` run:

```
   8 [NVML-STUB] nvmlDeviceGetConfComputeProtectedMemoryUsage called (NOT IMPLEMENTED)
```

`nvmlDeviceGetConfComputeProtectedMemoryUsage` (`bridge/stubs_generated.go:188`) and `nvmlDeviceGetConfComputeMemSizeInfo` (`:183`).

Split out of #680 because, unlike the rest of that batch, this surface cannot be implemented until we decide which knob gates it — and the gate #680 originally proposed produces the wrong answer.

### The decision to make

Real hardware reports `0 MiB` because Confidential Compute *mode* is off, not because the part cannot do CC. NVML draws that distinction itself: `nvmlSystemGetConfComputeCapabilities` reports what the platform supports, while `nvmlSystemGetConfComputeState` / `...Settings` report whether it is switched on. The mock models neither — all five system-level CC exports are stubs (`bridge/stubs_generated.go:1032-1054`) and real CC behaviour is #377.

Meanwhile `features.confidential_compute: true` on the `gb300`, `gb200`, `b200` and `h100` profiles reads as a capability claim, and it is never consumed: `FeaturesConfig` is declared at `engine/config_types.go:546` and no field of it is referenced anywhere in the engine.

So gating protected memory on `features.confidential_compute`, as #680 originally proposed, would make `gb300` report *nonzero* protected memory — contradicting the reference tray and #680's own acceptance criterion that every field be hardware-plausible on `gb300`. Two ways out:

1. **Answer `0 MiB` unconditionally** while CC mode is unmodeled, and leave `features.confidential_compute` as the capability flag it reads like. Cheapest, matches the reference output exactly, and leaves #377 room to add mode later.
2. **Add a CC mode key** (for example `confidential_compute.mode: off`) distinct from the capability flag, gate protected memory on it, and report nonzero protected memory only when mode is on.

Option 1 is recommended. Option 2 is worth it only if we want the e2e test to assert a non-default value here, and it partly pre-empts #377.

## Proposed solution

Implement both exports off the engine device, reporting `0 MiB` total / used / free for a CC-capable part with mode off. Settle option 1 versus 2 on this issue before implementing, since it determines whether any new config key exists.

## Scope / non-goals

In scope: the two device exports, and whichever config surface the decision above picks.

Non-goals: real Confidential Compute behaviour — attestation, CC-on mode, encrypted transfers — which is #377; the five system-level CC exports, unless option 2 requires the state export to stay self-consistent.

## Audience

Internal NVIDIA teams.

## Acceptance criteria

- [ ] `nvmlDeviceGetConfComputeMemSizeInfo` and `nvmlDeviceGetConfComputeProtectedMemoryUsage` are hand-written in the bridge; their stubs disappear on regeneration.
- [ ] `Conf Compute Protected Memory Usage` reports Total, Used and Free on `gb300` instead of `N/A`, matching the reference tray's `0 MiB` under the chosen option.
- [ ] The capability-versus-mode decision is recorded on this issue and reflected in `docs/configuration.md`.
- [ ] `features.confidential_compute` is either read by the engine or removed from the profiles — it must not stay decorative.
- [ ] Profiles that do not configure this surface still report `N/A`.
- [ ] `nvidia-smi pmon -c 1` does not regress against `main` and `topo -m` still exits `0` — the new exports land in the internal export table path that #630 and #640 both got bitten by.
- [ ] Unit tests cover both exports.
- [ ] e2e assertion added to the `Label("nvidia-smi")` spec, building on the `tests/e2e/go/assertions/nvidiasmi/` scaffolding from #679.

giuliocalzo:
## Summary

Closes #711.

All three rows of the `nvidia-smi -q` `Conf Compute Protected Memory Usage` block read `N/A`, because both NVML getters behind them were generated stubs that `nvidia-smi` calls once per GPU on every `-q` run. They now report `0 MiB` for Total, Used and Free, matching the reference GB300 tray.

**The capability-versus-mode decision (option 1, revised by the captures).** Protected memory answers to CC *mode*, not to whether the part can do CC — and the hardware in-tree says so plainly: all seven captures in `tests/e2e/go/assertions/nvidiasmi/testdata/hardware` report `0 MiB` here on drivers 570 and later, **including the A100, L40S and T4 boards that cannot do Confidential Compute at all**. So gating on `features.confidential_compute`, as #680 proposed, would have made a T4 report `N/A` where a real one reports `0 MiB`, on top of making `gb300` report a non-zero protected region against its own reference tray. Both getters therefore answer zero on every device, unconditionally, and read no config. Nothing partitions protected memory until CC mode is on, which stays unmodelled (#377).

What *does* gate the surface is the driver version: the CC APIs arrived in `525`, so both functions are registered there and a profile pinned below it reports `N/A`, as real hardware of that vintage does. That is the one direction in which these rows legitimately decline, and it replaces the issue's "profiles that do not configure this surface still report N/A" criterion, which only made sense under the capability-gate reading.

`features.confidential_compute` is removed from the `h100`, `b200`, `gb200` and `gb300` profiles and from `FeaturesConfig` rather than left decorative, per the issue's requirement that it be read or removed.

Two smaller notes. `nvmlConfComputeMemSizeInfo_t` was an opaque forward declaration, so the bridge needs its full definition to fill the caller's buffer; a `_Static_assert` pins it at two `unsigned long long`s, because callers allocate from go-nvml's struct and a field added on this side would write past their allocation. And neither getter advances the failure-injection counter — `nvidia-smi` reads both on every `-q`, so ticking here would move when injection trips — while both still surface a lost GPU, matching the other per-query reads that answer from static state.

## Test plan

- [x] `make test` — full unit suite passes, including the new engine tests for both getters (zeros on every device, `ERROR_GPU_IS_LOST` once the device is tripped) and the `FunctionAvailable` cases for the 525 boundary.
- [x] `make test-mocknvml-bridge` — 72 bridge tests pass, including four new `confcompute/gpu{0,1}/{memsize,protected_usage}` cases driven through the built `libnvidia-ml.so.1`, which is the only place a field written at the wrong offset or width would show.
- [x] `make helm-tests` — 170 tests, 15 snapshots. The `configmap` snapshot drops the removed key; the `daemonset` snapshot moved only because the profile ConfigMap checksum annotation changed.
- [x] Both stubs disappear on regeneration: the generated file's count drops from 264 to 262.
- [x] The e2e assertion is asserted against every hardware capture rather than a chosen value, and rejects the pre-fix mock capture already in testdata.
- [x] `golangci-lint --fix` and `go vet` clean. `govulncheck` reports seven findings, all standard-library and all "found in `go1.26.5`, fixed in `go1.26.6`" — a local toolchain mismatch, since the repo pins `GOLANG_VERSION=1.26.6`.
- [ ] CI e2e: the new `Label("nvidia-smi")` spec, plus `pmon -c 1` and `topo -m` for the internal-export-table regression #630 and #640 were both bitten by.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Added
+- mocknvml: the `Conf Compute Protected Memory Usage` block of `nvidia-smi -q`
+  now reports `0 MiB` for Total, Used and Free instead of `N/A`. Both NVML
+  getters behind it were generated stubs that `nvidia-smi` calls once per GPU on
+  every `-q` run, so the mock said the driver could not tell whether any memory
+  is protected, where every real board answers that none is. The values are not
+  configurable and are not gated on the part: all seven real-hardware captures
+  in-tree report `0 MiB` here on drivers 570 and later, including the A100, L40S
+  and T4 boards that cannot do Confidential Compute at all. Nothing partitions
+  protected memory until CC mode is on, which stays unmodelled (#377). The two
+  exports are registered as arriving in driver `525`, so a profile pinned below
+  that reports `N/A` as real hardware of that vintage does. The unread
+  `features.confidential_compute` key is removed from the `h100`, `b200`,
+  `gb200` and `gb300` profiles: gating this surface on it would have made a T4
+  report `N/A` where a real one reports `0 MiB`. (#711)
 - A `--observability` Tilt consumer that runs kube-prometheus-stack and the GPU
   Operator's `dcgm-exporter` over mock GPUs and ships a Grafana dashboard
   in-tree, so the real exporter is scraped and rendered on a cluster with no
--- a/deployments/nvml-mock/helm/nvml-mock/profiles/b200.yaml
+++ b/deployments/nvml-mock/helm/nvml-mock/profiles/b200.yaml
@@ -341,7 +341,6 @@ device_defaults:
     transformer_engine: true
     fp4_support: true
     fp8_support: true
-    confidential_compute: true
     decompression_engine: true
     fifth_gen_tensor_cores: true
 
--- a/deployments/nvml-mock/helm/nvml-mock/profiles/gb200.yaml
+++ b/deployments/nvml-mock/helm/nvml-mock/profiles/gb200.yaml
@@ -378,7 +378,6 @@ device_defaults:
     transformer_engine: true        # 2nd gen Transformer Engine
     fp4_support: true               # FP4 precision support
     fp8_support: true               # FP8 precision support
-    confidential_compute: true      # CC/TEE support
     nvlink_c2c: true                # NVLink Chip-to-Chip (Grace connection)
     decompression_engine: true      # Hardware decompression
     fifth_gen_tensor_cores: true    # 5th generation tensor cores
--- a/deployments/nvml-mock/helm/nvml-mock/profiles/gb300.yaml
+++ b/deployments/nvml-mock/helm/nvml-mock/profiles/gb300.yaml
@@ -381,7 +381,6 @@ device_defaults:
     fp4_support: true               # FP4 precision support
     fp6_support: true               # FP6 precision (Blackwell Ultra)
     fp8_support: true               # FP8 precision support
-    confidential_compute: true      # CC/TEE support
     nvlink_c2c: true                # NVLink Chip-to-Chip (Grace connection)
     decompression_engine: true      # Hardware decompression
     fifth_gen_tensor_cores: true    # 5th generation tensor cores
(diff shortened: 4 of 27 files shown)


ArangoGutierrez (approved):
Thanks for this, gating the two new exports on driver 525 is a nice touch: a profile pinned below that keeps reading N/A the way real hardware of that vintage does, and the _Static_assert on nvmlConfComputeMemSizeInfo_t pins the ABI the bridge writes into.

PR #719, opened 2026-08-24
feat(node-agent): Port PCI Sysfs Simulator

### LINKED ISSUE #688: MEP0003: Implement Node Agent

roma-glushko:
## What This PR Does

Porting PCIBus simulator over to node agent. Cleaning up setup.sh & redundant CLIs.

Defined a new home for shims in the project root. Moved libpcisysfs there. Refactored shim_tests.go.

## Why

A part of MEP0003.

## Checklist


shims/libpcisysfs/testbin/fopen-read.c
roma-glushko, line 1: I have untangled test bin fixtures from tests. Tests are not supposed to compile binaries 😌 It's fine to expect it's done already (by things like make).

shims/libpcisysfs/README.md
roma-glushko, line 1: shims is a new home for all shims/libs that we produce in mokka. Similarly to cmds directory, they will be our entrypoints for all libs that we depends on. Should be clear to enumerate and find out them if we pull them to the project root.

internal/pcisysfs/config.go
roma-glushko, line 1: I have moved pcisysfs render to internal and separated it from libpcisysfs shim.

internal/agent/gpudriver/gpudriver_test.go
roma-glushko, line 1: Added tests as a followup to my previous PR (https://github.com/NVIDIA/k8s-test-infra/pull/705)

internal/agent/pcibus/pcibus.go
```diff
@@ -0,0 +1,99 @@
...
+		}
+	}
+
+	return errors.Join(errs...)
+}
+
+// Apply writes the NFD local-source feature file so NFD can derive
+// feature.node.kubernetes.io/pci-10de.present=true from it.
+func (s *Simulator) Apply(_ context.Context, h *host.Host, _ *agent.State) error {
+	return h.WriteFile(filepath.Join(h.Etc, nfdFeatureFile), []byte(nfdContent), 0o644)
```
faganihajizada, line 93: How is `nodeLabels.pciVendorPresent=false` handled here? Apply always writes the NFD feature file, while this value is only passed to the old nvml-mock container. I verified that the file is still created when the setting is off. Maybe we pass this setting to the node agent and remove or skip the file when disabled.

internal/agent/pcibus/pcibus.go
```diff
@@ -0,0 +1,99 @@
...
+
+// Ready reports whether the last Stage call completed without error.
+func (s *Simulator) Ready() bool { return s.ready.Load() }
+
+// Stage renders the PCI sysfs tree under h.Root and stages libpcisysfs.so.
+// When the state carries no topology the render is a no-op.
+func (s *Simulator) Stage(_ context.Context, h *host.Host, state *agent.State) error {
+	s.ready.Store(false)
+
+	if err := stageSysfs(h, state); err != nil {
```
faganihajizada, line 53: Stage adds the new topology but does not remove old entries. Can we remove devices that are no longer in the config?

giuliocalzo (dismissed):

deployments/nvml-mock/helm/nvml-mock/templates/daemonset.yaml
```diff
@@ -73,13 +73,11 @@ spec:
...
-            # the file and cleanup.sh removes it only when this is "on", so the
-            # mock never deletes an input it did not supply. NFD, not the mock,
-            # creates the label. NFD's PCI source cannot derive it as deployed —
-            # nfd-worker reads a link-time-fixed /host-sys path, so the rendered
-            # tree is not on a path it reads; see setup.sh step 7.
+            # Gates the direct node label (nvidia.com/gpu.present=true) that
+            # setup.sh sets via kubectl as a fast-path before NFD propagates.
+            # The canonical NFD feature file (nvml-mock.features) is written by
+            # the node-agent container, not setup.sh; this flag controls only
+            # the legacy kubectl-label step in setup.sh step 7.
```
ArangoGutierrez: setup.sh no longer reads MOCK_NFD_PCI_LABEL anywhere after this change, so describing it as the gate on step 7's kubectl label is not accurate, and nodeLabels.pciVendorPresent=false now suppresses nothing: pcibus.Apply writes nvml-mock.features unconditionally and the node-agent container is never handed the variable. cleanup.sh:46 still gates its rm on it, so with the value off the file gets created and never removed at teardown, and the helm-unittest case that claims to cover suppression only asserts the rendered env value. Either plumb the gate into the node-agent and honour it in Apply/Revoke, or retire the value and its docs together.

internal/agent/pcibus/stage.go
```diff
@@ -0,0 +1,91 @@
...
+func buildIdentities(state *agent.State) map[string]pcisysfs.PCI {
+	ids := make(map[string]pcisysfs.PCI, len(state.Devices))
+
+	for _, d := range state.Devices {
+		if d.PCIBusID == "" {
+			continue
+		}
+
+		ids[strings.ToLower(d.PCIBusID)] = pcisysfs.PCI{
+			BusID:    d.PCIBusID,
```
ArangoGutierrez: buildIdentities carries BusID and DeviceID but nothing for subsystem_id, and agent.DeviceSpec has no field to hold it, so renderDeviceAttrs falls back to subsystem_vendor = vendor and writes subsystem_device 0x0000 for every device, in the attribute files and in the synthetic config space. The deleted Profile.DeviceIdentities() merged device_defaults.pci.subsystem_id (h100.yaml sets 0x165810DE), so lspci's subsystem line regresses on every shipped profile.

internal/agent/pcibus/stage.go
```diff
@@ -0,0 +1,91 @@
...
+
+	return nil
+}
+
+// buildTopology converts the agent's PCIeTopology to the pcisysfs type.
+// Returns nil when the state carries no root complexes (Render treats nil as no-op).
+func buildTopology(state *agent.State) *pcisysfs.PCIeTopology {
+	rcs := state.NodeShape.Topology.RootComplexes
+
+	if len(rcs) == 0 {
```
ArangoGutierrez, line 66: Returning nil on an empty RootComplexes list drops the flat-default fallback Profile.EffectiveTopology() gave the old renderer: a config with devices but no pcie_topology: block used to get a single pci0000:00 root at NUMA 0 covering every BDF, and now renders no sysfs tree at all. The bundled profiles all declare a topology so only hand-written configs hit it, but it fails as a silent no-op rather than an error.

ArangoGutierrez (commented):
Good consolidation: moving the shim to shims/libpcisysfs and the renderer to internal/pcisysfs finally gives the PCI simulator one home, and folding the render into the node-agent retires a binary and about a hundred lines of setup.sh.

The move itself lands cleanly. The things worth another pass are all places where behaviour that used to live in setup.sh or render-pci-sysfs did not come across with it.

- TestReady_FalseAfterDiscard asserts that Ready() stays true, so the name reads as the opposite of the contract its comment documents. TestReady_SurvivesDiscard would match. (internal/agent/pcibus/pcibus_test.go:151)

internal/agent/pcibus/pcibus.go
roma-glushko: I have retired this env var in this PR https://github.com/NVIDIA/k8s-test-infra/pull/719/commits/fb0fdb9d74710515b466c34c554fefaffe4fb022 🗑️ 

This simplifies the system and doesn't seem to take away any significant capabilities e.g. if there is somebody else who writes the same feature NFD should be okay with that.

deployments/nvml-mock/helm/nvml-mock/templates/daemonset.yaml
roma-glushko: I have retired this env var in this PR https://github.com/NVIDIA/k8s-test-infra/pull/719/commits/fb0fdb9d74710515b466c34c554fefaffe4fb022 🗑️ 

This simplifies the system and doesn't seem to take away any significant capabilities e.g. if there is somebody else who writes the same feature NFD should be okay with that.

internal/agent/pcibus/stage.go
roma-glushko: Addressed in https://github.com/NVIDIA/k8s-test-infra/pull/719/commits/2b939f9206b85a11211cd37ab0e8d931ac999360 ✅ 

Thank you for spotting!

internal/agent/pcibus/stage.go
roma-glushko: The behaviour should be restored in https://github.com/NVIDIA/k8s-test-infra/pull/719/commits/e35a484d87537f296c2eae1fe502c3bcf67399fd ✅

internal/agent/pcibus/pcibus.go
roma-glushko: Great catch! PCI bus and GPU Driver were not accounting for the previous state of their resources (e.g. PCI tree and nvidia char devices accordingly). 

I have added a logic that should make them reconcile the current state properly without leaving any leftover:

https://github.com/NVIDIA/k8s-test-infra/pull/719/commits/65bb67e269925205b226e9be1e76240e16140cdb

PR #724, opened 2026-08-25
fix(local): keep nvml-mock off the control-plane node

giuliocalzo:
## Summary

The nvml-mock chart ships `tolerations: [{operator: Exists}]`, so in every local-dev and CI Kind cluster the DaemonSet landed on the control-plane node as well as the workers. Nothing follows it there — GPU Operator operands, FGO and NFD workers all stop at the `NoSchedule` taint — so the control plane advertised a GPU driver footprint (`/run/nvidia/driver`, `/var/run/cdi/nvidia.yaml`, `nvidia.com/gpu.present`) that no consumer could ever use, and scenarios that pick a mock pod by list position intermittently targeted it. Two e2e helpers already carry comments about working around exactly that.

- Label every Kind worker `mokka.nvidia.com/type=sgpu` (`local/kind/default.kind.yaml`, `local/kind/compute-domain.kind.yaml`) and select on it from `local/nvml-mock.values.yaml`, the shared local-dev baseline.
- Repeat the selector in `local/compute-domain/nvml-mock.values.yaml`: `compute_domain.tiltfile` installs the chart itself and does not layer that baseline. It is load-bearing there anyway — `topology.yaml` assigns a clique per worker and describes no control plane, so a mock pod on it reported a fabric identity the topology never declared.
- Fix the GPU Operator e2e: `verifyGPUOperatorNodeSetup` probed the *control-plane* container for the mock's driver footprint, which only existed because the mock ran there. It now probes the node the specs actually assert on (`gpuOperatorTargetNode`), which is where the Operator's operands live too.

Selector pinning stays additive: Helm deep-merges maps, so the FGO overlay (`run.ai/simulated-gpu-node-pool`) and `--multi-gpu-profile`'s `--set nodeSelector.kubernetes\.io/hostname=worker-N` compose with it rather than replacing it.

## Test plan

- [x] `make lint-fix` — 0 issues (the non-zero exit is the pre-existing `govulncheck` stdlib finding on `main`)
- [x] `make helm-tests` — 173 tests, 9 suites, 15 snapshots pass
- [x] `go vet -tags e2e ./tests/e2e/go/...`
- [x] `helm template` with each overlay combination renders the expected merged selector: baseline → `mokka.nvidia.com/type=sgpu`; compute-domain → same; FGO + `--multi-gpu-profile` → all three keys
- [ ] CI e2e (`e2e`, `dra`, `gpu-operator`, `multi-node`, `nfd`, `nri`) — all jobs create the cluster with `make cluster-create` on the default profile, so the labels are present

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -129,6 +129,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   public NVML APIs.
 
 ### Changed
+- The local-dev and CI Kind clusters no longer run nvml-mock on the control-plane
+  node. The chart tolerates every taint, so the DaemonSet landed there too, but
+  nothing follows it: GPU Operator operands, FGO and NFD workers all stop at the
+  `NoSchedule` taint. That left a node advertising a GPU driver footprint no
+  consumer could use, and scenarios that pick a mock pod by list position
+  intermittently targeted it. Every Kind worker now carries
+  `mokka.nvidia.com/type=sgpu` and `local/nvml-mock.values.yaml` selects on it;
+  the compute-domain overlay repeats the selector because its Tiltfile installs
+  the chart without that baseline, and its `topology.yaml` gives the control
+  plane no clique to report. Pinning stays additive — Helm deep-merges maps, so
+  the FGO pool selector and `--multi-gpu-profile`'s hostname pin compose with
+  it. (#724)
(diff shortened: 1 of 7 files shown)


ArangoGutierrez (approved):
/lgtm

roma-glushko:
Looks good to me 👍

PR #725, opened 2026-08-25
docs: lead with the Kubernetes install and standardize on the mokka namespace

giuliocalzo:
## Summary

- **Promote the Helm install to the top of the quickstart.** It was `Option 3`, behind two local `libnvidia-ml.so` build options, so the fastest path to simulated GPUs was the last thing a reader found. The Go 1.25/GCC prerequisites moved down with the build sections they actually apply to, since the Helm path doesn't need them. The `docker pull` / `kind load` lines are gone too — the chart already pulls `ghcr.io/nvidia/nvml-mock:latest` by default.
- **Standardize every namespace reference on `mokka`.** The install snippets used the release default while `nvml-mock-ctl.md` and the NRI guidance in `helm-chart.md` said `-n nvml-mock`, so copy-pasting between pages produced commands that resolved to nothing. `mokka` is the namespace `local/nvml_mock.tiltfile` provisions and the one the Ginkgo E2E suite queries, so the docs now match the tooling.
- **Note `make cluster-create`** in the quickstart and landing page, for readers who need the CDI-enabled Kind node image that the device plugin, DRA driver, and GPU Operator paths run on.

Docs only; no chart, code, or CI changes.

## Test plan

- [ ] `mkdocs build --strict` passes (heading levels changed in `quickstart.md`; no other page links to those anchors, verified with a repo-wide search for `quickstart.md#`).
- [ ] Follow the quickstart top to bottom on a fresh Kind cluster: `kind create cluster --name mokka`, the `helm install` with `--namespace mokka --create-namespace`, then `kubectl exec -n mokka ds/nvml-mock -- nvidia-smi -L` lists 8 GB300 GPUs.
- [ ] Spot-check a couple of `nvml-mock-ctl` examples against a release installed in `mokka`.

## Notes for reviewers

The `-n mokka` change in the `nri.enabled` values-table row is load-bearing rather than cosmetic: the NRI plugin excludes its own release namespace, so the surrounding text warning against installing into `default` still reads correctly with the new name.

`docs/demo/nv-sentinel/README.md` still uses `nvml-mock-system` and the other demo docs have their own install steps. I left those alone since each demo is self-contained and internally consistent; happy to sweep them in a follow-up if you'd prefer one name everywhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ No physical NVIDIA hardware required.
 kind create cluster --name mokka
 
 helm install nvml-mock oci://ghcr.io/nvidia/k8s-test-infra/chart/nvml-mock \
+    --namespace mokka --create-namespace \
     --set gpu.profile=gb300
 ```
 
--- a/docs/helm-chart.md
+++ b/docs/helm-chart.md
@@ -46,7 +46,7 @@ default. Kind clusters must have containerd NRI enabled; see
 
 ```bash
 helm install nvml-mock oci://ghcr.io/nvidia/k8s-test-infra/chart/nvml-mock \
-  -n nvml-mock --create-namespace \
+  -n mokka --create-namespace \
   --set nri.enabled=true
 ```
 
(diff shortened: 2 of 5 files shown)


ArangoGutierrez (approved):
/lgtm

roma-glushko:
This is great 👍

PR #727, opened 2026-08-25
fix(profiles): model gb200/gb300 as a 4-GPU NVL72 compute tray

giuliocalzo:
## Summary

The `gb200` and `gb300` profiles declared 8 GPUs over 4 PCI root complexes — the shape the 8-GPU baseboard profiles use. That is not what a Grace-Blackwell node reports. An NVL72 rack reaches 72 GPUs as 18 compute trays of 4, and `nvidia-smi` runs per node: the real captures added in #694 show 4 attached GPUs sharing one tray, one slot and one chassis, over 2 Grace CPUs.

```
$ rg '<attached_gpus>' tests/e2e/go/assertions/nvidiasmi/testdata/hardware/qx-gb300.xml
<attached_gpus>4</attached_gpus>
```

A node twice its real size inflates every count a consumer derives from it — allocatable GPUs, GFD's `nvidia.com/gpu.count`, ResourceSlice sizes — and invents a second pair of Grace CPUs and NUMA nodes that no such node has.

Both profiles now ship 4 devices over 2 root complexes (`pci0000:00`, `pci0000:40`), in the chart `profiles/` and the engine `configs/` copies. Two further properties of the captures come with the trim, because a consumer that assumes them away breaks on real hardware rather than here:

- the two GPUs of a superchip report the **same board serial**, so GPUs 0-1 share one and GPUs 2-3 share another;
- `module_id` does **not** follow the device index — the captures report 2, 1, 4, 3 across devices 0..3, and so do the profiles.

The BDFs stay synthetic (`0000:0A/0B/4A/4B`). Real hardware reports `0000:29/3F/9C/B2` on GB200 and per-GPU PCI domains on GB300; adopting those would ripple into the `render-pci-sysfs` domain handling and the fixtures keyed on `0000:0A:00.0`, which is a separate change from getting the node's shape right.

### User-visible

`gb300` is the chart default, so **a default install now exposes 4 GPUs per node rather than 8.** `gpu.count` left empty still derives from the profile's `devices:` list, so select an 8-GPU baseboard profile or set `gpu.count` explicitly for a larger node.

The standalone and node-wide demos now default `GPU_COUNT` from the selected profile's device list instead of a hardcoded 8. This was a latent bug, not just tidiness: `node-wide-injection/run.sh` defaults to the `gb200` profile, and `standalone/demo.sh` asserts the rendered PCI device count equals `GPU_COUNT`, so the hardcoded 8 would have failed its step 9 check.

Oracles and docs updated in lockstep: the `tests/e2e/go/profile` derivation table (4 GPUs, 4 HCAs, 2 PCI roots), the engine profile-consistency tests, the Helm unittests (including the two that pinned the default count at 8) and their snapshots, `docs/helm-chart.md`, `docs/quickstart.md`, `docs/configuration.md` and the top-level README.

## Test plan

- [x] `make test`
- [x] `make helm-tests` (173 tests, 15 snapshots)
- [x] `golangci-lint` clean via `make lint-fix` (its `govulncheck` step fails on pre-existing Go 1.26.5 stdlib advisories, unrelated to this change)
- [x] `bash -n` on both touched demo scripts, and the derived counts verified per profile: `gb200`/`gb300`/`t4` -> 4 devices, the baseboards -> 8
- [ ] e2e legs for the `gb200` and `gb300` profiles in CI

## Notes for reviewers

The mock-output fixtures under `tests/e2e/go/assertions/nvidiasmi/testdata/` (`qx-gb200-healthy.xml` and friends) are 2-GPU captures carrying unique serials, so they no longer match what the profile would emit. Nothing asserts on `serial`, so they remain valid parser fixtures, but they will drift further if regenerated.

### DIFF
--- a/deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml
+++ b/deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml
@@ -125,13 +125,14 @@ tests:
           path: spec.template.spec.containers[0].lifecycle.preStop.exec.command
           value: ["/scripts/cleanup.sh"]
 
-  - it: should set default GPU_COUNT env var to 8
+  # The default profile is gb300, whose device list is one NVL72 compute tray.
+  - it: should set default GPU_COUNT env var to 4
     asserts:
       - contains:
           path: spec.template.spec.containers[0].env
           content:
             name: GPU_COUNT
-            value: "8"
+            value: "4"
 
   - it: should set default DRIVER_VERSION for non-Blackwell profile
     set:
@@ -161,7 +162,7 @@ tests:
     set:
       gpu:
         profile: gb200
-        count: 8
+        count: 4
     asserts:
       - contains:
           path: spec.template.spec.containers[0].env
@@ -278,7 +279,7 @@ tests:
     set:
       gpu:
         profile: gb300
-        count: 8
+        count: 4
     asserts:
       - contains:
           path: spec.template.spec.containers[0].env
@@ -343,7 +344,7 @@ tests:
             name: GPU_COUNT
             value: "4"
 
-  - it: should derive GPU_COUNT from the gb200 profile devices list (8)
+  - it: should derive GPU_COUNT from the gb200 profile devices list (4)
     set:
       gpu:
         profile: gb200
@@ -353,7 +354,7 @@ tests:
           path: spec.template.spec.containers[0].env
           content:
             name: GPU_COUNT
-            value: "8"
+            value: "4"
 
   - it: should mount all required volumes
     asserts:
--- a/deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml
+++ b/deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml
@@ -332,6 +332,22 @@ tests:
             value: "4"
 
   # ── GPU_COUNT derived from the profile devices list (count unset) ──────
+  #
+  # One case per device-list length the shipped profiles carry, so a derivation
+  # replaced by any single constant fails here. The 4-GPU profiles alone would
+  # all agree with a hardcoded 4.
+  - it: should derive GPU_COUNT from the h100 profile devices list (8)
+    set:
+      gpu:
+        profile: h100
+        count: ""
+    asserts:
+      - contains:
+          path: spec.template.spec.containers[0].env
+          content:
+            name: GPU_COUNT
+            value: "8"
+
   - it: should derive GPU_COUNT from the t4 profile devices list (4)
     set:
       gpu:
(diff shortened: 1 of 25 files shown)


deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml
```diff
@@ -343,7 +344,7 @@ tests:
             name: GPU_COUNT
             value: "4"
 
-  - it: should derive GPU_COUNT from the gb200 profile devices list (8)
+  - it: should derive GPU_COUNT from the gb200 profile devices list (4)
```
ArangoGutierrez, line 363: Both derivation cases now expect 4 (t4 just above, gb200 here), and so do both NOTES count assertions, so hardcoding nvml-mock.gpuCount to `{{- 4 -}}` keeps all 173 helm-unittest tests green. That same mutant fails 5 tests on the base, where gb200 derived 8. An h100 derivation case with `count: ""` asserting 8 restores the kill, and freeing `count` on the h100 NOTES case does the same for notes_test.yaml.

Also at deployments/nvml-mock/helm/nvml-mock/tests/notes_test.yaml:26.

ArangoGutierrez (dismissed):
Thanks for chasing this down to the captures rather than to a round number. Pinning gb200/gb300 to what qx-gb200.xml actually reports, shared board serials and the 2, 1, 4, 3 module ids included, makes both profiles a far better stand-in for a real tray.

The docs and oracles move in lockstep here, and the derivation table in tests/e2e/go/profile carries the new shape cleanly.

- The t4 entry two lines below still reads 8x while mock-nvml-config-t4.yaml declares 4 devices, which is the same drift this change is correcting for gb200 and gb300. (pkg/gpu/mocknvml/README.md:121)

roma-glushko (dismissed):
Great spot! This is definitely wrong.

deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml
giuliocalzo: Good catch, and the mutant reproduces — with one nuance: hardcoding `nvml-mock.gpuCount` to `{{- 4 -}}` left every *targeted* assertion green and was caught only incidentally by `should match snapshot with all overrides`, which pins an explicit override. So the derivation itself was no longer pinned by anything purpose-built, exactly as you describe.

Fixed in c1841a6b along the lines you suggested:

- a new `should derive GPU_COUNT from the h100 profile devices list (8)` case with `count: ""` in `daemonset_test.yaml`;
- `count` freed on the h100 NOTES case, so it asserts the derived `GPUs per node: 8` against the default profile's 4.

Verified by mutation on the branch: constant `4` now fails 3 tests (both new assertions plus that snapshot), constant `8` fails 10, and the unmutated suite is 174/174.

The `t4` README line is fixed in the same commit. I checked the rest of that list against the configs while I was there — `a100`/`h100`/`b200`/`l40s` really do declare 8 devices, so `t4` was the only stale entry.

PR #728, opened 2026-08-25
feat(node-agent): Node Agent CDI Simulator

### LINKED ISSUE #688: MEP0003: Implement Node Agent

roma-glushko:
## What This PR Does

Ported the CDI logic as a new node agent Simulator.
Cleaned up nvml-mock scripts from ported logic.

## Why

A part of MEP0003.
## Checklist

---

### DIFF
--- a/cmd/node-agent/main.go
+++ b/cmd/node-agent/main.go
@@ -93,6 +93,7 @@ func runStart(ctx context.Context, cmd *cli.Command) error {
 	if err != nil {
 		return err
 	}
+
 	log := logging.NewLogger(logging.Config{Level: level, Format: format})
 
 	configPath := cmd.String("config")
--- a/cmd/node-agent/main.go
+++ b/cmd/node-agent/main.go
@@ -17,6 +17,7 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	"github.com/NVIDIA/k8s-test-infra/internal/agent"
+	"github.com/NVIDIA/k8s-test-infra/internal/agent/cdi"
 	"github.com/NVIDIA/k8s-test-infra/internal/agent/gpudriver"
 	"github.com/NVIDIA/k8s-test-infra/internal/agent/health"
 	"github.com/NVIDIA/k8s-test-infra/internal/agent/host"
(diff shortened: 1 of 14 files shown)


giuliocalzo (dismissed):

faganihajizada (dismissed):

giuliocalzo:
Non-blocking suggestion: consider giving the CDI simulator a logger and emitting one line per `Apply`, the way I ended up doing it for `rdmaplugin` in #734.

The reasoning is the same for both. `agent.reconcile` logs only failures, so a simulator that succeeds is invisible, and `FileSource` emits an update on subscribe and then only when the config hash changes — so `Apply` runs once at startup and then rarely. One `Info` per `Apply` is therefore bounded, and it's the only evidence the surface was actually published.

That matters more for CDI than for most surfaces: when the spec is missing or stale, the symptom shows up far away, as containerd refusing container creation or injecting nothing, and the first question is whether the agent wrote the spec at all and with how many devices.

Something like:

```go
func New(log *slog.Logger) *Simulator { return &Simulator{log: log} }

func (s *Simulator) Apply(_ context.Context, h *host.Host, state *agent.State) error {
    nvidiaPath := filepath.Join(h.Run, nvidiaSpecFile)
    nriPath := filepath.Join(h.Run, nriSpecFile)
    if err := writeSpec(h, nvidiaPath, buildNvidiaSpec(state)); err != nil {
        return fmt.Errorf("nvidia.yaml: %w", err)
    }
    if err := writeSpec(h, nriPath, buildNRISpec(state)); err != nil {
        return fmt.Errorf("nvml-mock-nri.yaml: %w", err)
    }
    s.log.Info("wrote CDI specs", "simulator", name,
        "nvidia", nvidiaPath, "nri", nriPath, "devices", len(state.Devices))
    return nil
}
```

`rdmaplugin` also carries a `patched` field to distinguish a real write from a no-op, but that only exists because it reads the node before patching. `host.WriteFile` rewrites unconditionally, so there's no equivalent here and I wouldn't add change detection just for the log line.

Happy to fold this into #734 instead if you'd rather keep this PR as a straight port.

giuliocalzo (approved):

ArangoGutierrez (approved):
/lgtm

PR #735, opened 2026-08-25
fix(mocknvml): deliver injected Xid events the way real NVML does

harche:
## What This PR Does

Fixes two places where the mock NVML behaves differently from a real GPU, which made injected Xid errors invisible to health monitors such as the DRA driver's.

1. Xid events now say "this is the whole GPU, not a MIG slice". Real NVML fills `gpuInstanceId` and `computeInstanceId` in `nvmlEventData_t` with `0xFFFFFFFF` when the event is not tied to a MIG instance. The mock wrote `0` and `0`. Clients read that as MIG instance 0 on compute instance 0, which does not exist on a GPU without MIG, so they dropped the event.

2. Device state now survives `nvmlShutdown()` followed by `nvmlInit()`. The mock used to throw away its devices on shutdown and build new ones on the next init, which also reset the failure injector. On real hardware a tripped Xid or ECC fault does not disappear because a process reinitializes NVML. Many clients call init and shutdown around device discovery and then call init again for their event loop, so the failure tripped during discovery was lost before the event loop could see it. The devices are now kept; old handles are still invalid and the API still returns `ERROR_UNINITIALIZED` after shutdown.

Also:
* The bridge test hook (`xidDelivery`) now exposes the instance ids so the sentinel is covered by a test.
* New engine test for state persistence, and the two existing tests that expected the server to be dropped on shutdown are updated.
* CHANGELOG entry under Unreleased / Fixed.

## Why

I found this while writing bats e2e tests for the GPU DRA driver health reporting (kubernetes-sigs/dra-driver-nvidia-gpu#1243) on top of the mock NVML kind setup. With `nvml-mock-ctl fail --mode ecc_uncorrectable --xid 79` the driver never tainted the device or reported it as Unhealthy. The event was either never raised (state lost across the driver's init, shutdown, init sequence) or raised with instance ids the driver treats as an invalid MIG address.

With this change the driver's health path works end to end on the mock: the ResourceSlice gets `gpu.nvidia.com/xid=79:NoSchedule` and the pod's `allocatedResourcesStatus` shows `Unhealthy`.

Note for existing users: an injected failure now stays in effect across NVML reinitialization inside the same process, matching the documented sticky behavior of the injector. Restarting the pod still resets everything, as before.

## Checklist

## AI Usage

Claude Code assisted with the investigation, code, tests and this write up; I reviewed the changes and verified the results myself.

### DIFF
--- a/pkg/gpu/mocknvml/bridge/process_layout_test.go
+++ b/pkg/gpu/mocknvml/bridge/process_layout_test.go
@@ -55,8 +55,8 @@ func TestWriteProcessEntry_StrideAndFields(t *testing.T) {
 		e := unsafe.Add(base, i*procEntrySize)
 		require.Equal(t, want.pid, *(*uint32)(e), "entry %d pid", i)
 		require.Equal(t, want.mem, *(*uint64)(unsafe.Add(e, 8)), "entry %d usedGpuMemory", i)
-		require.Equal(t, uint32(0xFFFFFFFF), *(*uint32)(unsafe.Add(e, 16)), "entry %d gpuInstanceId", i)
-		require.Equal(t, uint32(0xFFFFFFFF), *(*uint32)(unsafe.Add(e, 20)), "entry %d computeInstanceId", i)
+		require.Equal(t, nvmlNoMIGInstanceID, *(*uint32)(unsafe.Add(e, 16)), "entry %d gpuInstanceId", i)
+		require.Equal(t, nvmlNoMIGInstanceID, *(*uint32)(unsafe.Add(e, 20)), "entry %d computeInstanceId", i)
 
 		nameBuf := unsafe.Slice((*byte)(unsafe.Add(e, procEntryNameOffset)), procEntryNameMax)
 		got, _, found := bytes.Cut(nameBuf, []byte{0})
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -197,6 +197,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   rejected. (#719)
 
 ### Fixed
+- mocknvml: Xid critical-error events are now attributed to the whole GPU the
+  way real NVML does — `nvmlEventData_t.gpuInstanceId`/`computeInstanceId`
+  carry the `0xFFFFFFFF` "not a MIG instance" sentinel instead of `0`/`0`,
+  which consumers (e.g. the DRA driver's health monitor) treat as MIG GI 0 /
+  CI 0 and drop on a GPU without MIG enabled.
+- mocknvml: device state (tripped failure injection, ECC counters, undelivered
+  Xid events) now survives a full `nvmlShutdown()`/`nvmlInit()` cycle, as it
+  does on real hardware. Previously the devices were recreated on re-init, so
+  a failure tripped during one init cycle was silently forgotten before a
+  client that inits NVML again (e.g. for an event-set wait loop) could observe
+  it. This also means the `after_calls` counter and the seeded `probability`
+  roll sequence continue across a re-init within one process rather than
+  starting over. Handles issued before the shutdown remain invalid, and their
+  addresses are never reused for new handles.
 - mocknvml: `nvmlPciInfo_t.busId` now reports the 8-digit PCI domain real NVML
   uses (`00000000:07:00.0`, `NVML_DEVICE_PCI_BUS_ID_FMT`) while `busIdLegacy`
   keeps the 4-digit one (`0000:07:00.0`). Both were filled with the profile's
--- a/docs/helm-chart.md
+++ b/docs/helm-chart.md
@@ -1281,7 +1281,9 @@ recover.
 
 Each `nvidia-smi` invocation is a fresh process whose call counter starts at
 0, so a narrow query like `--query-gpu=ecc.errors.uncorrected.aggregate.total`
-will only ever issue **one** guarded call per GPU per invocation. To see the
+will only ever issue **one** guarded call per GPU per invocation. (Within a
+single process the counter, like the rest of the device state, also survives
+`nvmlShutdown()` followed by `nvmlInit()`.) To see the
 failure surface from a single short command set `after_calls: 1`, or use a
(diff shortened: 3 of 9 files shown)


harche:
For context, here is how this shows up from a consumer's side. I hit both problems while writing a bats e2e for the DRA driver's KEP 4680 health reporting on top of the mock NVML kind setup.

**Where the driver relies on the behaviour fixed here**

1. The `0xFFFFFFFF` sentinel. The driver maps an Xid event to a device by its instance address. Only `gpuInstanceId == computeInstanceId == 0xFFFFFFFF` is treated as "the whole GPU"; anything else is looked up as a MIG instance and ignored if there is none:
   https://github.com/harche/k8s-dra-driver-gpu/blob/ca278431/cmd/gpu-kubelet-plugin/device_health.go#L444-L470

2. State across shutdown and init. The health monitor calls `Init()` and defers `Shutdown()` during construction, then calls `Init()` again when it registers events and starts its wait loop. Device discovery (which trips the injector via `GetMemoryInfo`) runs in yet another init/shutdown cycle before that:
   https://github.com/harche/k8s-dra-driver-gpu/blob/ca278431/cmd/gpu-kubelet-plugin/device_health.go#L165
   https://github.com/harche/k8s-dra-driver-gpu/blob/ca278431/cmd/gpu-kubelet-plugin/device_health.go#L190

**The e2e that exercises it**

The test writes the override, restarts the kubelet plugin so discovery trips the injector, then expects the ResourceSlice taint `gpu.nvidia.com/xid=79:NoSchedule` and `Unhealthy` in the pod's `allocatedResourcesStatus`:
https://github.com/harche/k8s-dra-driver-gpu/blob/ca278431/tests/bats/test_gpu_health.bats#L110-L190

With the mock at current `main` this test fails; with this branch the whole mock e2e suite passes for me (kind v1.36.4, `gb200` profile, 2 GPUs).

**Small reproducer without Kubernetes**

This only needs the mock library. It mimics the driver: init, one guarded call to trip the injector, shutdown, init again, then wait for the event.

```go
package main

import (
	"fmt"

	"github.com/NVIDIA/go-nvml/pkg/nvml"
)

func main() {
	nvml.Init()
	dev, _ := nvml.DeviceGetHandleByIndex(0)
	dev.GetMemoryInfo() // guarded call: trips the configured failure
	nvml.Shutdown()

	nvml.Init() // what a health monitor does after discovery
	defer nvml.Shutdown()
	dev, _ = nvml.DeviceGetHandleByIndex(0)
	set, _ := nvml.EventSetCreate()
	dev.RegisterEvents(nvml.EventTypeXidCriticalError, set)
	ev, ret := set.Wait(2000)
	fmt.Printf("ret=%v xid=%d gi=%#x ci=%#x\n", ret, ev.EventData, ev.GpuInstanceId, ev.ComputeInstanceId)
}
```

Override used:

```yaml
version: 1
all:
  failure:
    mode: ecc_uncorrectable
    xid:
      code: 79
```

Run with `MOCK_NVML_CONFIG=<driver_root>/config/config.yaml MOCK_NVML_OVERRIDES=/tmp/ov.yaml LD_LIBRARY_PATH=<dir containing libnvidia-ml.so.1>`.

Mock at `main` (3f47697):

```
ret=ERROR_TIMEOUT xid=0 gi=0x0 ci=0x0
```

If the shutdown/init pair is removed, the event does arrive at `main`, but as `gi=0x0 ci=0x0`, which the driver discards:

```
ret=SUCCESS xid=79 gi=0x0 ci=0x0
```

With this PR, in both variants:

```
ret=SUCCESS xid=79 gi=0xffffffff ci=0xffffffff
```

pkg/gpu/mocknvml/bridge/process_layout_test.go
```diff
@@ -55,8 +55,8 @@ func TestWriteProcessEntry_StrideAndFields(t *testing.T) {
 		e := unsafe.Add(base, i*procEntrySize)
 		require.Equal(t, want.pid, *(*uint32)(e), "entry %d pid", i)
 		require.Equal(t, want.mem, *(*uint64)(unsafe.Add(e, 8)), "entry %d usedGpuMemory", i)
-		require.Equal(t, uint32(0xFFFFFFFF), *(*uint32)(unsafe.Add(e, 16)), "entry %d gpuInstanceId", i)
-		require.Equal(t, uint32(0xFFFFFFFF), *(*uint32)(unsafe.Add(e, 20)), "entry %d computeInstanceId", i)
+		require.Equal(t, nvmlNoMIGInstanceID, *(*uint32)(unsafe.Add(e, 16)), "entry %d gpuInstanceId", i)
```
ArangoGutierrez, line 58: Swapping the `uint32(0xFFFFFFFF)` literal for `nvmlNoMIGInstanceID` puts the same symbol on both sides of the assertion, so nothing pins the value any more. Setting the constant to 0, which is the bug this PR fixes, leaves the whole bridge package green. Keeping the literal on the expectation side restores the pin.

Also at pkg/gpu/mocknvml/bridge/events_test.go:191.

ArangoGutierrez (approved):
Thanks for this. The 0xFFFFFFFF instance-id sentinel and keeping device state across nvmlShutdown/nvmlInit are both real fidelity gaps, and the standalone reproducer plus the DRA driver e2e pin down exactly why each one mattered.

The new engine tests fail when the persistence change is reverted, so the behaviour is pinned where it counts. The rest is small and none of it is on the fix itself:

- The constant sits directly under writeProcessEntry's doc comment with no blank line between them, so godoc now reads that comment as documenting nvmlNoMIGInstanceID and leaves writeProcessEntry undocumented. A blank line above the const puts both back. (pkg/gpu/mocknvml/bridge/internal.go:280)
- The type doc still says the C block is released only by Clear(), which this change inverts: Clear now retires blocks and Free is the only path that releases, and Free has no callers in the tree. Worth refreshing that paragraph and the LIFECYCLE bullet so the ownership rule in a CGo file matches what the code does. (pkg/gpu/mocknvml/engine/handles.go:82) https://github.com/NVIDIA/k8s-test-infra/blob/a5b7422b293d24969b7dd24e055a0d6758cb0030/pkg/gpu/mocknvml/engine/handles.go#L80-L88

giuliocalzo (approved):

PR #737, opened 2026-08-26
fix(mocknvml): report the OEM boost ceiling instead of N/A

### LINKED ISSUE #712: [Feature]: Max Customer Boost Clocks is N/A on gb300 and cannot be derived from clocks.graphics_max
## Problem statement

A real 4-GPU GB300 tray (driver `580.173.02`, CUDA 13.0) reports `Max Customer Boost Clocks` → Graphics as `2070 MHz`. The `gb300` profile reports `N/A`.

Split out of #680, which proposed deriving this from the existing `clocks.graphics_max` "requiring no new key". That does not work: `gb300` sets `graphics_max: 2200` (`configs/mock-nvml-config-gb300.yaml:129`), so deriving it would report 2200 MHz where hardware says 2070. Customer boost max is an OEM-defined limit — `nvml.h` describes `NVML_CLOCK_ID_CUSTOMER_BOOST_MAX` as "OEM-defined maximum clock rate" — and it sits genuinely below boost max on real parts. It needs its own key.

### Two exports, both stubbed

The `-q` census shows `nvmlDeviceGetClock` called 8× per run, once per GPU, and it is a stub (`bridge/stubs_generated.go:148`):

```
   8 [NVML-STUB] nvmlDeviceGetClock called (NOT IMPLEMENTED)
```

But `nvmlDeviceGetMaxCustomerBoostClock` is the dedicated API for this row and is *also* a stub (`:372`), which #680 did not list. Whichever one the 580 binary uses, leaving the other stubbed keeps the value unreachable for DCGM and other go-nvml callers, so both should resolve from the same config.

One further trap: `nvmlDeviceGetClock` is a two-dimensional lookup over clock type × clock ID. Implementing it to answer only `CUSTOMER_BOOST_MAX`, while returning `NOT_SUPPORTED` for `CURRENT`, `APP_CLOCK_TARGET` and `APP_CLOCK_DEFAULT`, would be incoherent — the mock already holds every one of those values in its `clocks:` block (`graphics_current`, `graphics_app`, `graphics_app_default`, `sm_*`, `memory_*`, `video_*`). Once the export exists it should answer the whole matrix.

## Proposed solution

- Add `clocks.customer_boost_max`, defaulting to `clocks.graphics_max` when unset so existing profiles are unchanged, and set it to `2070` on `gb300`.
- Implement `nvmlDeviceGetMaxCustomerBoostClock` and `nvmlDeviceGetClock`, both hand-written, both resolving from the `clocks:` block: the full clock-type × clock-ID matrix from the existing keys, and `CUSTOMER_BOOST_MAX` from the new one.

Alternatives considered: deriving from `graphics_max` with no new key, as originally proposed in #680 — rejected, it cannot reproduce the reference value and would make the e2e assertion tautological.

## Scope / non-goals

In scope: the new clock key, both exports, and the full clock matrix `nvmlDeviceGetClock` implies.

Non-goals: settable clocks (`nvmlDeviceSetApplicationsClocks` and friends); clock-throttle reason modeling, which already exists; per-profile clock retuning beyond the `gb300` value this issue needs.

## Audience

Internal NVIDIA teams.

## Acceptance criteria

- [ ] `nvmlDeviceGetClock` and `nvmlDeviceGetMaxCustomerBoostClock` are hand-written in the bridge; their stubs disappear on regeneration.
- [ ] `nvmlDeviceGetClock` answers `GRAPHICS`, `SM`, `MEM` and `VIDEO` across `CURRENT`, `APP_CLOCK_TARGET`, `APP_CLOCK_DEFAULT` and `CUSTOMER_BOOST_MAX` from the `clocks:` block, returning `NOT_SUPPORTED` only for combinations the config does not carry.
- [ ] `Max Customer Boost Clocks` → Graphics reports `2070 MHz` on `gb300` — a value deliberately distinct from `graphics_max`, so the test cannot be satisfied by deriving it.
- [ ] `clocks.customer_boost_max` falls back to `clocks.graphics_max` when unset; profiles that set neither still report `N/A`.
- [ ] `nvidia-smi pmon -c 1` does not regress against `main` and `topo -m` still exits `0` — the new exports land in the internal export table path that #630 and #640 both got bitten by.
- [ ] Unit tests cover both exports, including that a `customer_boost_max` differing from `graphics_max` is honoured.
- [ ] e2e assertion added to the `Label("nvidia-smi")` spec, building on the `tests/e2e/go/assertions/nvidiasmi/` scaffolding from #679.
- [ ] `docs/configuration.md` documents `clocks.customer_boost_max` and its fallback.

giuliocalzo:
## What This PR Does

`Max Customer Boost Clocks` read `N/A` on every profile. Both NVML entry points that can answer the row were generated stubs: the dedicated `nvmlDeviceGetMaxCustomerBoostClock`, and `nvmlDeviceGetClock`, which `nvidia-smi` calls once per GPU on every `-q` run. The mock therefore claimed the driver could not report an OEM boost ceiling, where every real board reports one.

Both are now hand-written, and their stubs disappear on regeneration (`make gen` reports 160 implementations and 260 stubs, down from 158 and 262).

`nvmlDeviceGetClock` answers its whole clock-type x clock-id matrix rather than only the combination this row needs, so callers reaching it for current or application clocks stop getting `NOT_SUPPORTED` too. It routes to the getter that owns each value instead of re-reading the `clocks:` block, so the generic entry point and the dedicated getters cannot disagree — including about which combinations are unsupported.

## Why

Closes #712, but not the way that issue proposed, and the difference is the main thing to review.

The issue asked for a new `clocks.customer_boost_max` key set to `2070` on `gb300`, on the premise that the OEM ceiling "sits genuinely below boost max on real parts". The seven hardware captures in-tree contradict that premise. A100, H100, L40S, T4, B200, GB200 and GB300 all report `max_customer_boost_clocks` equal to `max_clocks`, and all four GPUs of the reference GB300 tray report 2070 MHz for both. The 2070-vs-2200 gap the issue read as a customer-boost limit was the `gb300` profile's `graphics_max` being wrong.

So the ceiling derives from `clocks.graphics_max` with no new key. `nvml.h` does leave room for an OEM to set the ceiling lower, but no board we have captured does, and a second key would only have given profiles two values to keep in step by hand.

That makes correcting the profiles part of the fix rather than a side quest. The three Blackwell profiles move to their captures — `b200` 1965, `gb200` 2062, `gb300` 2070 — which also fixes `Max Clocks`, `Applications Clocks` and `Default Applications Clocks` on them. The application clocks move with the maximum because a real board reports them equal, and leaving them behind would have claimed an application clock above the GPU's own ceiling; `supported_clocks` is capped for the same reason.

### On the tautology risk the issue raised

Deriving from `graphics_max` does make a check of "the row equals the configured value" a restatement of the config, which is why the issue rejected the approach. The e2e check pins the equality against the `<max_clocks>` reading beside it as well, so a profile that ever needs the two to differ has to change the check and say why, and a unit test asserts the equality holds across every hardware capture. The profile derivation table also pins `graphics_max` per board, so a profile edited away from its capture fails there.

`pmon` and `topo -m` are re-checked alongside the new row, since a new export on the path `nvidia-smi` walks per GPU is what broke `pmon` in #630 and #640.

## Follow-up

`video_max` and the memory clocks on these three profiles are still off against their captures. Out of scope here and left for a separate issue.

## Checklist

- [x] Tests pass — `make test`, 40 packages, and `make helm-tests`
- [x] Documentation updated — `docs/configuration.md` covers how the row resolves
- [x] CHANGELOG.md updated


roma-glushko (approved):
Looks good to me 👍

ArangoGutierrez:
> ### Greptile Summary
> The PR implements the previously stubbed NVML customer-boost and generic clock queries and aligns the Blackwell clock profiles with captured hardware values.
> 
> * Routes generic clock requests through the corresponding dedicated engine getters.
> * Reports the graphics maximum as the captured OEM customer-boost ceiling.
> * Updates B200, GB200, and GB300 maximum, application, and supported clock values.
> * Adds engine, profile, hardware-capture, and standalone end-to-end coverage.
> * Updates generated stubs, Helm snapshots, documentation, and the changelog.
> 
> ### Confidence Score: 5/5
> The PR appears safe to merge.
> 
> No blocking failure remains.
> 
> ### Important Files Changed
> Filename	Overview
> pkg/gpu/mocknvml/engine/device.go	Implements the generic clock matrix and derives the graphics customer-boost ceiling from the configured graphics maximum.
> pkg/gpu/mocknvml/bridge/device.go	Exports the two newly implemented NVML entry points through the existing validated device-handle lookup pattern.
> pkg/gpu/mocknvml/engine/device_test.go	Covers supported and unsupported clock combinations, missing configuration, and invalid enum sentinels.
> tests/e2e/go/assertions/nvidiasmi/customer_boost.go	Validates each GPU’s reported customer-boost ceiling against both profile configuration and the adjacent maximum-clock reading.
> tests/e2e/go/scenario_standalone_test.go	Adds standalone end-to-end coverage for customer-boost reporting and adjacent nvidia-smi commands.
> pkg/gpu/mocknvml/configs/mock-nvml-config-gb300.yaml	Aligns GB300 graphics, SM, application, and supported maximum clocks with the reference capture.
> ### Flowchart
> ```mermaid
> %%{init: {'theme': 'neutral'}}%%
> flowchart LR
>     A[nvidia-smi or NVML caller] --> B[nvmlDeviceGetClock]
>     A --> C[nvmlDeviceGetMaxCustomerBoostClock]
>     B --> D[ConfigurableDevice.GetClock]
>     D --> E[Current clock getter]
>     D --> F[Application clock getters]
>     D --> G[Customer boost getter]
>     C --> G
>     G --> H[clocks.graphics_max]
>     H --> I[Max Customer Boost Clocks output]
> ```
> Reviews (2): Last reviewed commit: ["Merge branch 'main' into feat/issue-712-..."](https://github.com/nvidia/k8s-test-infra/commit/0ede0dd475713c9bc651f237891896c479bf119d) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=57417861)

Is there a way we can have Greptile label a PR if Greptile thinks is safe to merge?

ArangoGutierrez (approved):
/lgtm

Issue #689, opened 2026-08-20
[Feature]: serve the mock PCI tree without shadowing /sys/devices


giuliocalzo:
### Context

#674 serves the rendered PCI tree to Go consumers by bind-mounting `sys/devices` and `sys/bus/pci/devices` onto the kernel paths. Both mounts are needed together because the PCI entries are relative symlinks into `../../../devices/pciDDDD:BB`, and `/sys/devices` has to be mounted whole: narrowing it to the profile's root complexes needs a mountpoint the runtime cannot create on a read-only sysfs.

That works, and it is what makes `nvidia.com/gpu.mode` correct. The cost is everything else under `/sys/devices` that a served container loses:

- `/sys/devices/system/cpu` and the rest of the host's device classes.
- The container's own namespaced sysfs. `/sys/class/net/<iface>` is a symlink into `devices/virtual/net`, which dangles after the mount.
- `/sys/devices/virtual/dmi/id`, worked around in #674 by having the renderer reproduce kind's bind-mount targets.

Raised in review by @ArangoGutierrez (https://github.com/NVIDIA/k8s-test-infra/pull/674#discussion_r3813690942) and @roma-glushko (https://github.com/NVIDIA/k8s-test-infra/pull/674#discussion_r3814014315).

Mirroring the node's real `/sys/devices` into ours does not fix it. Copying attribute files freezes live values (CPU frequency, thermal zones, counters) into stale ones, and no copy of the *node's* tree can restore a container's namespaced sysfs, because what belongs under `devices/virtual/net` is the pod's own net namespace view, which the node does not have.

### Proposal

Stop shadowing `/sys/devices`. Mount only `/sys/bus/pci/devices` — a path every node's sysfs already has — and have the renderer emit symlink targets that escape sysfs into the overlay the mock already mounts, e.g.

```
/sys/bus/pci/devices/0000:07:00.0 -> ../../../../opt/nvml-mock/sys/devices/pci0000:00/0000:07:00.0
```

A consumer reading `vendor`, `class` or `numa_node` through the symlink resolves into the overlay and gets the same files it gets today. Nothing else in the container's `/sys` is touched: CPU topology, the DMI directory and the pod's namespaced sysfs all stay as the kernel populated them, and the renderer would no longer need to reproduce kind's DMI mount targets.

### Open questions

[...]

PR #742, opened 2026-08-27
fix(node-agent): serve the mock PCI tree at the kernel paths

### LINKED ISSUE #673: [Bug]: rendered PCI sysfs tree never reaches Go consumers, so GPU Operator labels gpu.mode and gpu.machine unknown

### LINKED ISSUE #681: [Bug]: nvidia.com/gpu.machine cannot be mocked under kind — the node image shadows /sys/class/dmi/id/product_name
Split out of #673, whose PCI half is fixed by #674. Mocking the machine type turns out to be a separate problem with a cause of its own, so it is tracked here.

## What happens

GPU Feature Discovery derives `nvidia.com/gpu.machine` from its `--machine-type-file`, which defaults to `/sys/class/dmi/id/product_name`. On a mock node that label never reflects the simulated platform.

Under kind it cannot, whatever the mock renders. The node image fakes the machine type for itself and for every container it starts:

- `/usr/local/bin/entrypoint` (`fix_product_name`): `echo 'kind' > /kind/product_name`, then `mount -o ro,bind /kind/product_name /sys/class/dmi/id/product_name` on the node.
- `/kind/bin/mount-product-files.sh`, registered as a `createContainer` hook in `/etc/containerd/cri-base.json` (so hook #0 for *every* container), repeats that bind into each container's rootfs, plus `product_uuid` at both `/sys/class/dmi/id/product_uuid` and `/sys/devices/virtual/dmi/id/product_uuid`.

Both mounts land on the file `/sys/class/dmi/id` resolves into, `/sys/devices/virtual/dmi/id/product_name` — the same file the mock renders into the tree #674 bind-mounts over `/sys/devices`. kind's hook runs after the container's mounts are set up, so kind wins: GFD reads `kind`, and a container reading either path sees `kind` too.

Both of kind's guards test the *host* kernel's DMI, so this bites exactly where CI runs (Linux) and not on Docker Desktop, whose linuxkit VM exposes no DMI and therefore no `/sys/class/dmi` for GFD to read either — the label is `unknown` there.

## Expected

A mock node labels `nvidia.com/gpu.machine` with the profile's platform (`NVIDIA-GB200-NVL72`, `DGXA100`, …), on Linux and macOS alike, and the e2e suite asserts it.

## Options

1. Render the machine type to a path kind does not shadow, mount it into served containers, and point GFD at it with `GFD_MACHINE_TYPE_FILE`. Works on any host; costs one env var in the GPU Operator values.
2. Keep GFD on its default path and accept `kind` under kind. No mocking, so nothing to assert.

## Notes

#674 renders `virtual/dmi/id/product_name` and `product_uuid` as *mirrors* of the node's own attributes, which is required for a different reason: shadowing `/sys/devices` removes the bind-mount targets kind's hook needs, and `mount(8)` cannot create one on a read-only sysfs, so every served pod fails to start with `mount: .../sys/class/dmi/id/product_uuid: mount point does not exist`. That mirroring is deliberately not a machine-type mock.

giuliocalzo:
## What This PR Does

Makes the simulated PCI tree and machine type visible to Go consumers, so a mock node labels itself the way a real one does.

- **Serves the rendered PCI tree at the kernel paths.** The agent's `nvidia.com/gpu` CDI spec bind-mounts `sys/devices` and `sys/bus/pci/devices` read-only, as a pair. Fixes `nvidia.com/gpu.mode=unknown`.
- **Reproduces the node's DMI attributes inside the tree.** Serving `/sys/devices` replaces what `/sys/class/dmi/id` resolves into, so without this a container reads `ENOENT` where the node has values — and under `kind`, whose node image bind-mounts its product files into every container, the pod fails to start outright.
- **Renders only the GPUs NVML reports.** `gpu.count` caps the device list without touching `pcie_topology`, so a capped node used to render PCI entries that read as GPUs with no NVML device behind them.
- **Serves a machine-type file and points GFD at it via NRI.** Fixes `nvidia.com/gpu.machine=unknown`.

Trade-off: `/sys/devices` is served whole, so a served container no longer sees the host's other device classes, CPU topology among them. Narrowing it needs mountpoints the runtime cannot create on a read-only sysfs; #689 tracks removing it.

Known gap: the CDI path still needs `GFD_MACHINE_TYPE_FILE` in the operator's values, because the runtime applies the spec's mounts but drops its env — filed as #747. NRI needs no override.

## Why

Fixes #673. `pcibus` rendered the tree into the agent's overlay, reachable only through the `libpcisysfs.so` `LD_PRELOAD` shim. That covers libc consumers like `lspci`, but Go bypasses it — `os.Open` issues `openat` directly — so GPU Feature Discovery and the NVIDIA DRA driver read the node's real `/sys` and found no mock GPUs. GFD resolved each GPU's BDF from NVML, failed to read its class, and labelled the node `nvidia.com/gpu.mode=unknown`; the DRA driver omitted `dra.k8s.io/pcieRoot` for the same reason.

Fixes #681. `nvidia.com/gpu.machine` read `unknown` for a separate reason: GFD's default source for it is a DMI path no mock can own under `kind`, and one that does not exist at all on hosts without DMI.

## Checklist

- [x] Tests pass (`go test -v -race ./...`) — plus `make helm-tests` (175 tests, 15 snapshots) and `go vet` including under the e2e build tag
- [x] Linter passes (`make lint-fix`) — `golangci-lint` reports 0 issues; the target's `govulncheck` stage fails on Go 1.26.5 stdlib advisories fixed in 1.26.6, pre-existing and reached only through code this PR does not touch
- [x] Documentation updated (if applicable) — `docs/helm-chart.md`

Verified with `tilt ci -- --gpu-operator --gpu-profile gb300` on `kind`: both workers reach `gpu.mode=compute` and `gpu.machine=NVIDIA-GB300-NVL`, and all four GPUs resolve at the kernel paths from inside a served container. Not verified locally: `kind`'s DMI mount-target hook, which cannot fire on Docker Desktop (no DMI) and is covered by Linux CI; and `dra.k8s.io/pcieRoot`, which needs a `--dra` run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -370,6 +370,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   profile that declares no `bus_id`, where it previously left the last
   profile's devices in place — invisible until the tree started being served,
   and then a container enumerating GPUs the node no longer simulates. (#673)
+- node-agent: the rendered PCI tree now follows the devices NVML reports rather
+  than `pcie_topology` as the profile writes it. `gpu.count` (`GPU_COUNT`) caps
+  the device list without touching the layout, so a capped node used to render
+  BDFs with no NVML device behind them — phantom NVIDIA 3D controllers, every
+  attribute plausible, harmless only while nothing could read the tree. Declared
+  BDFs no device claims are dropped along with any root left empty, and a device
+  no root claims is adopted by the first one, since an unplaced GPU cannot be
+  resolved at all where a misplaced one only misreports its `pcieRoot`. (#673)
 - node-agent: `nvidia.com/gpu.machine` no longer reads `unknown`. GFD derives it
   from `--machine-type-file`, whose default `/sys/class/dmi/id/product_name` no
   mock can own under kind: the node image writes `kind` there and re-binds it
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -345,6 +345,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   roll sequence continue across a re-init within one process rather than
   starting over. Handles issued before the shutdown remain invalid, and their
   addresses are never reused for new handles.
+- node-agent: the rendered PCI sysfs tree now reaches Go consumers. It was
+  readable only through `MOCK_PCI_ROOT` and the `libpcisysfs.so` `LD_PRELOAD`
+  shim, which covers libc consumers such as `lspci` but not Go: `os.Open` issues
+  `openat` directly, the shim never sees it, and the process reads the node's
+  real `/sys`, where the mock GPUs do not exist. GPU Feature Discovery and the
+  NVIDIA DRA driver are both Go, so GFD resolved each mock GPU's BDF from NVML,
+  failed to read its class, and labelled the node
+  `nvidia.com/gpu.mode=unknown`. The `nvidia.com/gpu` CDI spec the agent writes
+  now bind-mounts `sys/devices` and `sys/bus/pci/devices` read-only onto the
+  kernel paths. The two go together, because the PCI entries are relative
+  symlinks into `../../../devices/pciDDDD:BB`: mounting one alone yields entries
+  that list and whose every attribute read returns `ENOENT`, which is
+  indistinguishable from no mount at all. `/sys/devices` is mounted whole —
+  narrowing it to the profile's root complexes needs a mountpoint the runtime
+  cannot create on a read-only sysfs, and container creation fails outright
+  rather than degrading — so a served container does not see the host's other
+  device classes, CPU topology among them (#689). The `pcibus` simulator also
+  mirrors the node's `product_name` and an empty `product_uuid` stand-in into
+  `sys/devices/virtual/dmi/id`: that is where `/sys/class/dmi/id` resolves, and
+  kind bind-mounts the node's product files there for every container, so
+  without the mirror the shadowing removes a target `mount(8)` cannot recreate
+  and every served pod fails to start. This is not a machine-type mock, which
+  remains blocked under kind (#681). Rendering also empties the tree for a
+  profile that declares no `bus_id`, where it previously left the last
+  profile's devices in place — invisible until the tree started being served,
+  and then a container enumerating GPUs the node no longer simulates. (#673)
 - mocknvml: `nvmlPciInfo_t.busId` now reports the 8-digit PCI domain real NVML
   uses (`00000000:07:00.0`, `NVML_DEVICE_PCI_BUS_ID_FMT`) while `busIdLegacy`
   keeps the 4-digit one (`0000:07:00.0`). Both were filled with the profile's
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -366,11 +366,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `sys/devices/virtual/dmi/id`: that is where `/sys/class/dmi/id` resolves, and
   kind bind-mounts the node's product files there for every container, so
   without the mirror the shadowing removes a target `mount(8)` cannot recreate
-  and every served pod fails to start. This is not a machine-type mock, which
-  remains blocked under kind (#681). Rendering also empties the tree for a
+  and every served pod fails to start. Rendering also empties the tree for a
   profile that declares no `bus_id`, where it previously left the last
   profile's devices in place — invisible until the tree started being served,
   and then a container enumerating GPUs the node no longer simulates. (#673)
+- node-agent: `nvidia.com/gpu.machine` no longer reads `unknown`. GFD derives it
+  from `--machine-type-file`, whose default `/sys/class/dmi/id/product_name` no
+  mock can own under kind: the node image writes `kind` there and re-binds it
+  into every container after the container's own mounts are set up, and hosts
+  without DMI (Docker Desktop) have no such path at all. The agent now writes
+  the machine type to `driver/config/machine-type`, served at
+  `/etc/nvml-mock/machine-type` by the CDI mount that already carries
+  `config.yaml`, and the chart's GPU Operator values point
+  `GFD_MACHINE_TYPE_FILE` at it. The value is the profile's GPU product name for
+  want of a platform name in the profiles, so the label reads
+  `NVIDIA-GB300-NVL`, matching `gpu.product`, rather than the
+  `NVIDIA-GB300-NVL72` a real compute tray reports. (#681)
 - mocknvml: `nvmlPciInfo_t.busId` now reports the 8-digit PCI domain real NVML
   uses (`00000000:07:00.0`, `NVML_DEVICE_PCI_BUS_ID_FMT`) while `busIdLegacy`
   keeps the 4-digit one (`0000:07:00.0`). Both were filled with the profile's
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -345,49 +345,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   roll sequence continue across a re-init within one process rather than
   starting over. Handles issued before the shutdown remain invalid, and their
   addresses are never reused for new handles.
-- node-agent: the rendered PCI sysfs tree now reaches Go consumers. It was
-  readable only through `MOCK_PCI_ROOT` and the `libpcisysfs.so` `LD_PRELOAD`
-  shim, which covers libc consumers such as `lspci` but not Go: `os.Open` issues
-  `openat` directly, the shim never sees it, and the process reads the node's
-  real `/sys`, where the mock GPUs do not exist. GPU Feature Discovery and the
-  NVIDIA DRA driver are both Go, so GFD resolved each mock GPU's BDF from NVML,
-  failed to read its class, and labelled the node
-  `nvidia.com/gpu.mode=unknown`. The `nvidia.com/gpu` CDI spec the agent writes
-  now bind-mounts `sys/devices` and `sys/bus/pci/devices` read-only onto the
-  kernel paths. The two go together, because the PCI entries are relative
-  symlinks into `../../../devices/pciDDDD:BB`: mounting one alone yields entries
-  that list and whose every attribute read returns `ENOENT`, which is
-  indistinguishable from no mount at all. `/sys/devices` is mounted whole —
-  narrowing it to the profile's root complexes needs a mountpoint the runtime
-  cannot create on a read-only sysfs, and container creation fails outright
-  rather than degrading — so a served container does not see the host's other
-  device classes, CPU topology among them (#689). The `pcibus` simulator also
-  mirrors the node's `product_name` and an empty `product_uuid` stand-in into
-  `sys/devices/virtual/dmi/id`: that is where `/sys/class/dmi/id` resolves, and
-  kind bind-mounts the node's product files there for every container, so
-  without the mirror the shadowing removes a target `mount(8)` cannot recreate
-  and every served pod fails to start. Rendering also empties the tree for a
-  profile that declares no `bus_id`, where it previously left the last
-  profile's devices in place — invisible until the tree started being served,
-  and then a container enumerating GPUs the node no longer simulates. (#673)
-- node-agent: the rendered PCI tree now follows the devices NVML reports rather
-  than `pcie_topology` as the profile writes it. `gpu.count` (`GPU_COUNT`) caps
-  the device list without touching the layout, so a capped node used to render
-  BDFs with no NVML device behind them — phantom NVIDIA 3D controllers, every
-  attribute plausible, harmless only while nothing could read the tree. Declared
-  BDFs no device claims are dropped along with any root left empty, and a device
-  no root claims is adopted by the first one, since an unplaced GPU cannot be
-  resolved at all where a misplaced one only misreports its `pcieRoot`. (#673)
-- node-agent: `nvidia.com/gpu.machine` no longer reads `unknown`. GFD derives it
-  from `--machine-type-file`, whose default `/sys/class/dmi/id/product_name` no
-  mock can own under kind: the node image writes `kind` there and re-binds it
-  into every container after the container's own mounts are set up, and hosts
-  without DMI (Docker Desktop) have no such path at all. The agent now writes
-  the machine type to `driver/config/machine-type`, served at
-  `/etc/nvml-mock/machine-type` by the CDI mount that already carries
-  `config.yaml`, and the chart's GPU Operator values point
-  `GFD_MACHINE_TYPE_FILE` at it. The value is the profile's GPU product name for
-  want of a platform name in the profiles, so the label reads
+- node-agent: the rendered PCI sysfs tree now reaches Go consumers. Only the
+  `libpcisysfs.so` `LD_PRELOAD` shim redirected the kernel paths to it, and Go
+  bypasses the shim — `os.Open` issues `openat` directly — so GPU Feature
+  Discovery read the node's real `/sys`, found none of the mock GPUs, and
+  labelled the node `nvidia.com/gpu.mode=unknown`; the NVIDIA DRA driver omitted
+  `dra.k8s.io/pcieRoot` for the same reason. The agent's `nvidia.com/gpu` CDI
+  spec now bind-mounts `sys/devices` and `sys/bus/pci/devices` read-only over
+  the kernel paths, as a pair, since the PCI entries are relative symlinks into
+  `../../../devices/pciDDDD:BB`. `/sys/devices` is served whole, so a served
+  container no longer sees the host's other device classes, CPU topology among
+  them; #689 tracks narrowing it. The node's DMI attributes are reproduced
+  inside the tree, without which kind's product-file bind-mounts lose their
+  target and every served pod fails to start. (#673)
+- node-agent: the rendered tree now follows the devices NVML reports rather than
+  `pcie_topology` as the profile writes it. `gpu.count` caps the device list
+  without touching the layout, so a capped node rendered PCI entries that read
+  as GPUs — NVIDIA vendor ID, GPU device class — with no NVML device behind
+  them. Declared BDFs no device claims are dropped along with any root left
+  empty, and a device no root claims is adopted by the first. A profile
+  declaring no `bus_id` now empties the tree instead of leaving the previous
+  profile's GPUs served. (#673)
+- node-agent: `nvidia.com/gpu.machine` no longer reads `unknown`. GFD's default
+  source for it, `/sys/class/dmi/id/product_name`, is a path no mock can own
+  under kind — the node image writes `kind` there and re-binds it into every
+  container — and hosts without DMI have no such path at all. The agent now
+  writes the machine type to `driver/config/machine-type`, and the NRI plugin
+  points `GFD_MACHINE_TYPE_FILE` at it, so an install needs no GPU Operator
+  override. Deployments using the CDI path still need one, because the runtime
+  applies the spec's mounts but drops its env. The value is the GPU product name
+  for want of a platform name in the profiles, so the label reads
   `NVIDIA-GB300-NVL`, matching `gpu.product`, rather than the
   `NVIDIA-GB300-NVL72` a real compute tray reports. (#681)
 - mocknvml: `nvmlPciInfo_t.busId` now reports the 8-digit PCI domain real NVML
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -375,9 +375,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   container — and hosts without DMI have no such path at all. The agent now
   writes the machine type to `driver/config/machine-type`, and the NRI plugin
   points `GFD_MACHINE_TYPE_FILE` at it, so an install needs no GPU Operator
-  override. Deployments using the CDI path still need one, because the runtime
-  applies the spec's mounts but drops its env (#747). The value is the GPU product name
-  for want of a platform name in the profiles, so the label reads
+  override. Deployments using the CDI path still need one, because the toolkit
+  resolving `nvidia.com/gpu` applies the spec's mounts and drops its env (#747).
+  The value is the GPU product name for want of a platform name in the
+  profiles, so the label reads
   `NVIDIA-GB300-NVL`, matching `gpu.product`, rather than the
   `NVIDIA-GB300-NVL72` a real compute tray reports. (#681)
 - mocknvml: `nvmlPciInfo_t.busId` now reports the 8-digit PCI domain real NVML
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -373,7 +373,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   writes the machine type to `driver/config/machine-type`, and the NRI plugin
   points `GFD_MACHINE_TYPE_FILE` at it, so an install needs no GPU Operator
   override. Deployments using the CDI path still need one, because the runtime
-  applies the spec's mounts but drops its env. The value is the GPU product name
+  applies the spec's mounts but drops its env (#747). The value is the GPU product name
   for want of a platform name in the profiles, so the label reads
   `NVIDIA-GB300-NVL`, matching `gpu.product`, rather than the
   `NVIDIA-GB300-NVL72` a real compute tray reports. (#681)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -363,9 +363,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   without touching the layout, so a capped node rendered PCI entries that read
   as GPUs — NVIDIA vendor ID, GPU device class — with no NVML device behind
   them. Declared BDFs no device claims are dropped along with any root left
-  empty, and a device no root claims is adopted by the first. A profile
-  declaring no `bus_id` now empties the tree instead of leaving the previous
-  profile's GPUs served. (#673)
+  empty. A device no root claims is still rendered, since one missing from the
+  tree is one no consumer can resolve, but under a root its own address implies
+  and reporting `numa_node` `-1` — the kernel's own "no proximity information" —
+  rather than borrowing the locality of a root the profile never put it in. A
+  profile declaring no `bus_id` now empties the tree instead of leaving the
+  previous profile's GPUs served. (#673)
 - node-agent: `nvidia.com/gpu.machine` no longer reads `unknown`. GFD's default
   source for it, `/sys/class/dmi/id/product_name`, is a path no mock can own
   under kind — the node image writes `kind` there and re-binds it into every
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -353,7 +353,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `dra.k8s.io/pcieRoot` for the same reason. The agent's `nvidia.com/gpu` CDI
   spec now bind-mounts `sys/devices` and `sys/bus/pci/devices` read-only over
   the kernel paths, as a pair, since the PCI entries are relative symlinks into
-  `../../../devices/pciDDDD:BB`. `/sys/devices` is served whole, so a served
+  `../../../devices/pciDDDD:BB`. Tearing the tree down empties those two
+  directories in place rather than removing them, so a container already holding
+  the mount is not left on a replaced inode reading an empty tree.
+  `/sys/devices` is served whole, so a served
   container no longer sees the host's other device classes, CPU topology among
   them; #689 tracks narrowing it. The node's DMI attributes are reproduced
   inside the tree, without which kind's product-file bind-mounts lose their
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -370,7 +370,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   tree is one no consumer can resolve, but under a root its own address implies
   and reporting `numa_node` `-1` — the kernel's own "no proximity information" —
   rather than borrowing the locality of a root the profile never put it in. A
-  profile declaring no `bus_id` now empties the tree instead of leaving the
+  `bus_id` that is not an address in the kernel's `DDDD:BB:DD.F` form is dropped
+  rather than placed: it named the default root, which every shipped profile
+  declares, so the device inherited that root's NUMA node — and since a `bus_id`
+  becomes a directory name, one carrying `/` or `..` rendered outside the tree.
+  A profile declaring no `bus_id` now empties the tree instead of leaving the
   previous profile's GPUs served. (#673)
 - node-agent: `nvidia.com/gpu.machine` no longer reads `unknown`. GFD's default
   source for it, `/sys/class/dmi/id/product_name`, is a path no mock can own
--- a/deployments/nvml-mock/helm/nvml-mock/templates/daemonset.yaml
+++ b/deployments/nvml-mock/helm/nvml-mock/templates/daemonset.yaml
@@ -255,6 +255,14 @@ spec:
             - name: host-fabric-state
               mountPath: {{ printf "/host%s" .Values.fabricmanager.stateDir }}
             {{- end }}
+            # Read-only, and read from rather than written to: serving the
+            # rendered tree at /sys/devices replaces the directory
+            # /sys/class/dmi/id resolves into, so the agent reproduces the
+            # node's DMI attributes inside the tree to keep the runtime's
+            # bind-mount targets alive.
+            - name: host-sys
+              mountPath: /host/sys
+              readOnly: true
           livenessProbe:
             httpGet:
               path: /healthz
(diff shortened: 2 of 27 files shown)


internal/agent/state.go
```diff
@@ -25,6 +25,27 @@ type State struct {
...
+//
+// It is the condition under which a PCI sysfs tree exists on the node, so
+// anything that serves that tree to a container gates on it. A bind mount whose
+// source is missing fails container creation for the whole pod, which is why
+// the answer has to come from the same state the renderer reads rather than
+// from a filesystem probe.
+func (s *State) HasPCITopology() bool {
+	if len(s.NodeShape.Topology.RootComplexes) > 0 {
+		return true
+	}
```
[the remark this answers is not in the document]
giuliocalzo: Good catch, and the mirror case is worse: `GPU_COUNT` truncates `State.Devices` while `pcie_topology` is compiled from the profile whole, so a node running gb300 with `gpu.count=2` rendered all four profile BDFs. The two extra entries look like real GPUs — `vendor` falls back to `0x10de` and `class` to a 3D controller — so a consumer enumerating the tree saw GPUs NVML denies exist. Harmless only while nothing could read the tree.

Fixed in 1f38fba3 by reconciling the layout against the devices, in one place both the renderer and the mount gate read (`State.PCITopology`):

- declared BDFs no active device claims are dropped, along with any root they leave empty;
- a device no root claims is adopted by the first surviving root — coverage beats placement, since an unplaced GPU cannot be resolved at all where a misplaced one only misreports its `pcieRoot`;
- `HasPCITopology` now answers from that same reconciliation instead of a parallel root-presence check, so the predicate cannot drift from what gets rendered.

Pinned against the real profile in `TestCompileState_TopologyTracksACappedDeviceCount`, which asserts both halves: the compiled state does carry four BDFs for two devices, and the rendered tree carries only the two.

ArangoGutierrez (dismissed):
/lgtm

deployments/nvml-mock/helm/nvml-mock/templates/daemonset.yaml
```diff
@@ -175,6 +175,11 @@ spec:
               mountPath: /host/run/nvidia
             - name: host-nfd-features
               mountPath: /host/etc/kubernetes/node-feature-discovery/features.d
+            # Debugging aid: lets `kubectl exec` compare the node's own sysfs
```
roma-glushko: This should be removed if debug is done. setup.sh is no longer working with picsysfs.

local/gpu-operator/gpu-operator.values.yaml
```diff
@@ -58,11 +58,17 @@ devicePlugin:
...
+#
+# The machine type has to come from a file of the mock's own: GFD's default,
+# /sys/class/dmi/id/product_name, reads "kind" under kind and does not exist on
+# hosts without DMI. The agent writes this one, served by the CDI config mount.
 gfd:
   enabled: true
   env:
     - name: NVIDIA_DRIVER_ROOT
       value: "/var/lib/nvml-mock/driver"
+    - name: GFD_MACHINE_TYPE_FILE
```
roma-glushko, line 75: In the real env, who is responsible to this override? Are we going to ask our end users to apply gpu operator override like that just to make mokka work? 

I have mentioned that in the previous review, I think NRI should do that. It's a part of simulation logic that should just work if mokka is installed.

CHANGELOG.md
```diff
@@ -297,6 +297,51 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   roll sequence continue across a re-init within one process rather than
   starting over. Handles issued before the shutdown remain invalid, and their
   addresses are never reused for new handles.
+- node-agent: the rendered PCI sysfs tree now reaches Go consumers. It was
```
roma-glushko: This one is TOO MUCH information. Let's make it concise.

CHANGELOG.md
```diff
@@ -297,6 +297,51 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  kind bind-mounts the node's product files there for every container, so
+  without the mirror the shadowing removes a target `mount(8)` cannot recreate
+  and every served pod fails to start. Rendering also empties the tree for a
+  profile that declares no `bus_id`, where it previously left the last
+  profile's devices in place — invisible until the tree started being served,
+  and then a container enumerating GPUs the node no longer simulates. (#673)
+- node-agent: the rendered PCI tree now follows the devices NVML reports rather
+  than `pcie_topology` as the profile writes it. `gpu.count` (`GPU_COUNT`) caps
+  the device list without touching the layout, so a capped node used to render
+  BDFs with no NVML device behind them — phantom NVIDIA 3D controllers, every
```
roma-glushko: What's NVIDIA 3D controllers?

internal/agent/cdi/spec_test.go
```diff
@@ -139,6 +140,62 @@ func TestNvidiaSpecFabricManagerDisabled(t *testing.T) {
...
+		if m.ContainerPath == path {
+			return m, true
+		}
+	}
+	return cdiMount{}, false
+}
+
+// GFD and the DRA driver are Go, so libpcisysfs.so never sees their openat
+// calls: only a real mount at the kernel path reaches them.
+func TestNvidiaSpecServesPCISysfsAtKernelPaths(t *testing.T) {
```
roma-glushko, line 185: Should be marked as Parallel()

internal/agent/gpudriver/gpudriver_test.go
```diff
@@ -99,6 +99,51 @@ func TestWriteEngineConfig_EmptyConfigRawErrors(t *testing.T) {
...
 }
 
+// GFD reads its machine type from a file, so the mock has to serve one: under
+// kind the DMI path it defaults to is either absent or owned by the node image.
+func TestWriteMachineType_ServesTheProductName(t *testing.T) {
+	h := testHost(t)
+	state := testState(t)
+	state.Devices[0].Name = "NVIDIA GB300 NVL"
+
+	require.NoError(t, writeMachineType(context.Background(), h, state))
```
roma-glushko:
```suggestion
	require.NoError(t, writeMachineType(t.Context(), h, state))
```

internal/agent/gpudriver/gpudriver_test.go
```diff
@@ -99,6 +99,51 @@ func TestWriteEngineConfig_EmptyConfigRawErrors(t *testing.T) {
...
+}
+
+// A missing file leaves GFD on its own default, which is the better failure:
+// an empty one would label the node with the empty string.
+func TestWriteMachineType_NoFileWithoutAProductName(t *testing.T) {
+	h := testHost(t)
+	state := testState(t)
+	state.Devices[0].Name = ""
+
+	require.NoError(t, writeMachineType(context.Background(), h, state))
```
roma-glushko:
```suggestion
	require.NoError(t, writeMachineType(t.Context(), h, state))
```

internal/agent/pcibus/stage.go
```diff
@@ -13,18 +14,76 @@ import (
...
-const defaultRootComplexID = "pci0000:00"
+const (
+	// kernelDMIRelPath is the kernel's DMI directory relative to /sys;
+	// /sys/class/dmi/id is a symlink to it.
+	kernelDMIRelPath = "devices/virtual/dmi/id"
+	// mockDMIRelPath is the same directory inside the rendered tree.
+	mockDMIRelPath = pcisysfs.SysDevicesRelPath + "/virtual/dmi/id"
+)
+
+// dmiMountTargets are the attributes kind's mount-product-files.sh hook
```
roma-glushko: This talks about kind, but is it relevant to any k8s cluster? Let's not forget that a regular EKS/GCP/Azure cluster is our legitimate deployment target.

CHANGELOG.md
```diff
@@ -297,6 +297,51 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  attribute plausible, harmless only while nothing could read the tree. Declared
+  BDFs no device claims are dropped along with any root left empty, and a device
+  no root claims is adopted by the first one, since an unplaced GPU cannot be
+  resolved at all where a misplaced one only misreports its `pcieRoot`. (#673)
+- node-agent: `nvidia.com/gpu.machine` no longer reads `unknown`. GFD derives it
```
roma-glushko: @giuliocalzo This seems to be the main change. But PR description goes over implementation details instead of clearly outlining and summarizing what was changed. I would suggest to tweak the prompt that writes description to make it more straight to the point. I would be easier to understand what the PR contains.

roma-glushko (dismissed):

deployments/nvml-mock/helm/nvml-mock/templates/daemonset.yaml
giuliocalzo: Removed in 02e1033b — it was a debugging aid and nothing in that container reads it. The agent container keeps its `/host/sys` mount, since `stageDMI` reads the node’s DMI attributes from there to reproduce them inside the rendered tree. The helm test now asserts the asymmetry rather than the pair.

CHANGELOG.md
giuliocalzo: Cut in 02e1033b. The two `#673` entries were largely the same story told twice, so they are now one entry plus a short one for the topology reconciliation, roughly half the length. The reasoning that was in there survives in the code comments, where it is closer to the thing it explains.

CHANGELOG.md
giuliocalzo: Jargon on my part — it is the PCI device class, `0x0302`. The entry now says the leftover entries "read as GPUs — NVIDIA vendor ID, GPU device class" instead, since the point is only that nothing tells a consumer they are fake.

internal/agent/cdi/spec_test.go
giuliocalzo: Added in 02e1033b, along with the other two PCI sysfs tests in that file. Worth noting `spec_test.go` has no `t.Parallel()` anywhere else (nor do the `pcibus`/`gpudriver` suites) — happy to sweep the file or the package in a follow-up if you want it consistent rather than only on the new tests.

internal/agent/gpudriver/gpudriver_test.go
giuliocalzo: Applied in 02e1033b, to all three new `writeMachineType` tests. The rest of the file still uses `context.Background()` (about twenty call sites); I left those alone to keep this diff to the point, but I can convert them in a follow-up.

internal/agent/pcibus/stage.go
giuliocalzo: Fair — the comment described the harshest failure and skipped the general one. Reworded in 02e1033b: reproducing the DMI attributes matters on any cluster, because serving the tree at `/sys/devices` replaces the directory `/sys/class/dmi/id` resolves into, so an EKS/GKE/AKS container would read `ENOENT` where the node has values. What is specific to `kind` is only that it fails harder than a missing value: the node image bind-mounts its own product files into every container, `mount(8)` cannot create a target on a read-only sysfs, and the pod dies on "mount point does not exist".

local/gpu-operator/gpu-operator.values.yaml
giuliocalzo: Agreed, and done in 02e1033b: the NRI plugin now injects `GFD_MACHINE_TYPE_FILE` pointing at the overlay’s `driver/config/machine-type`, right beside the `MOCK_NVML_CONFIG` it already injects. It goes through `setDefaultEnv`, so a cluster that pins its own file keeps it. With `nri.enabled` nobody has to touch GPU Operator values.

One constraint worth flagging, because it is why the override is still in the overlays rather than deleted. NRI is opt-in (`nri.enabled: false`), and the default path is CDI — where the runtime applies the spec’s mounts but drops its env. I retested that when this came up: `MOCK_NVML_CONFIG` set in `containerEdits.env` never arrives either, and the mock library only works there because `/etc/nvml-mock/config.yaml` is its compiled-in default. So on the CDI path there is no channel today that carries an env var into GFD, and `local/gpu-operator` and `tests/e2e` both drive that path.

So it "just works" wherever NRI runs, and the two overlays keep the override with a comment saying why. If you want the override gone everywhere, the CDI env drop is the thing to fix — I have not filed an issue for it yet; say the word and I will, since it is a latent trap for anything else that assumes `containerEdits.env` works.

CHANGELOG.md
giuliocalzo: Rewritten in the description now. "What This PR Does" is four bullets — serve the tree at the kernel paths, reproduce the DMI attributes, render only the GPUs NVML reports, serve the machine type through NRI — each with the label it fixes, plus the `/sys/devices` trade-off and the CDI env gap called out separately. The mechanism detail is gone from the description; it lives in the code comments and the changelog.

local/gpu-operator/gpu-operator.values.yaml
giuliocalzo: Filed as #747, with the reproduction and the note that `MOCK_NVML_CONFIG` is dropped the same way. The overlays, the docs and the CDI spec's env block now reference it, so the override says which defect it is working around — and the spec's env block carries a warning not to add a key that would actually depend on that channel.

internal/agent/state.go
```diff
@@ -26,6 +27,115 @@ type State struct {
...
+			orphans = append(orphans, bdf)
+		}
+	}
+
+	switch {
+	case len(orphans) == 0:
+		return roots
+	case len(roots) == 0:
+		return []RootComplex{{ID: DefaultRootComplexID, DeviceBDFs: orphans}}
+	default:
```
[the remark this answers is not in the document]
giuliocalzo: You are right, and the reasoning I put in that comment was wrong: I traded locality accuracy for coverage as if they were the only two options, when the tree can carry the device without asserting where it sits.

Fixed in 980f8466. An unplaced device now renders under the root its own address implies (`pciDDDD:BB` from the BDF), and joins a declared root only where the address names that root — there membership is the profile's own arithmetic, not a guess, so inheriting its NUMA node is legitimate. Otherwise it gets a root of its own reporting `numa_node` `-1`, which is not an invented value: it is what Linux writes for a device it has no proximity information for, so DRA and the device plugin read "unknown" through the encoding they already handle rather than a specific wrong node.

It is still rendered, though. A GPU absent from the tree is one no consumer can resolve from the BDF NVML hands it, which is the failure this whole path exists to fix — so the choice is between "here, locality unknown" and "confidently in the wrong place", and only the second is unrecoverable for a consumer.

Pinned at both levels: `TestPCITopology_UnplacedDeviceHasUnknownNUMA` and `TestPCITopology_UnplacedDeviceJoinsTheRootItsAddressNames` for the reconciliation, and `TestStageSysfs_UnplacedDeviceRendersUnknownNUMA` for the rendered `numa_node` file reading `-1`.

faganihajizada (changes requested):
Cross-linking #689, since the mount this PR adds lands on the same surface. The retarget idea there has no analogue here: /sys/bus/pci/devices holds kernel-made symlinks whose targets can be pointed into the overlay, while /sys/module/<name> is the data. What does carry over is the objection to shadowing. On a kind node, /sys/module holds 111 entries against 5 lines in /proc/modules, so a served container loses 106 built-in module directories, 94 of them with parameters/. kind's own injection is unaffected: its entrypoint binds a fixed list of three DMI paths and nothing under /sys/module, so unlike #674 this needs no reproduction of kind's mount targets.

ArangoGutierrez (commented):
Serving the tree at the kernel paths is the right fix, and reconciling the rendered layout against the devices NVML actually reports closes a gap every capped node was hitting.

The two-mount pair and the prune-on-empty path are the parts most likely to get copied later, so they are worth pinning down exactly.

- Two of Greptile's P1s on this file do not match the code: adopt renders a device no root claims rather than dropping it, and it never appends one to the first surviving root. It matches rootIDForBDF against the existing root IDs and otherwise opens a new root at numaNodeUnknown, which TestPCITopology_RendersADeviceNoRootClaims and TestPCITopology_UnplacedDeviceHasUnknownNUMA both pin. (internal/agent/state.go:99)

internal/agent/cdi/spec_test.go
```diff
@@ -139,6 +140,68 @@ func TestNvidiaSpecFabricManagerDisabled(t *testing.T) {
...
+// The two mounts are one feature: the entries are relative symlinks into
+// ../../../devices/pciDDDD:BB, so half the pair reads like no mount at all.
+func TestNvidiaSpecPCISysfsMountsAreEmittedAsAPair(t *testing.T) {
+	t.Parallel()
+
+	spec := buildNvidiaSpec(pciState())
+
+	_, devices := mountByContainerPath(spec, "/sys/bus/pci/devices")
+	_, sysDevices := mountByContainerPath(spec, "/sys/devices")
+	require.Equal(t, devices, sysDevices, "one half of the pair without the other")
```
ArangoGutierrez: mountByContainerPath returns (cdiMount, bool), so `_, devices :=` binds the found flag and the assertion compares two booleans, which passes when neither mount is emitted. Deleting the pciSysfsMounts append in spec.go leaves this test green while TestNvidiaSpecServesPCISysfsAtKernelPaths fails, so it never guards the pair invariant it names. Comparing the two cdiMount values, or requiring both oks true first, makes it discriminate.

internal/agent/cdi/spec.go
```diff
@@ -147,6 +156,38 @@ func buildNvidiaSpec(state *agent.State) cdiSpec {
...
+// ../../../devices/pciDDDD:BB, so serving it alone yields entries whose every
+// attribute read returns ENOENT.
+//
+// /sys/devices goes whole, hiding the host's other device classes from served
+// containers (#689): narrowing it needs mountpoints the runtime cannot create
+// on a read-only sysfs, and profiles declare root complexes the node lacks.
+//
+// Gated on the same condition pcibus renders under, since a mount with a
+// missing source fails container creation for the whole pod.
+func pciSysfsMounts(state *agent.State) []cdiMount {
```
ArangoGutierrez, line 182: pcibus.Discard does os.RemoveAll on h.Root/sys, so an agent restart replaces the directory inodes this spec names as mount sources. A consumer container that outlives the restart keeps its mount on the removed directory and reads an empty /sys/devices, putting gpu.mode back to unknown until the pod is recreated. gpudriver.Discard already avoids exactly this for the driver/config mount by removing leaves only.

docs/helm-chart.md
```diff
@@ -697,8 +695,8 @@ pcie_topology:
         - "0000:90:00.0"
 ```
 
-`render-pci-sysfs` validates the block at startup and fails the
-DaemonSet under `set -e` if it finds a typo:
+The simulator validates the block and reports a staging failure through the
```
ArangoGutierrez: Nothing under internal/agent or internal/pcisysfs validates pcie_topology, and engine.validateYAMLConfig checks only version, driver version and duplicate device indices, so a typo never reaches /healthz. The rules listed underneath now describe the opposite of what this PR ships: placeDeclared silently drops a BDF no device claims, and its !placed check silently dedupes one listed under two roots.

docs/helm-chart.md
```diff
@@ -707,7 +705,71 @@ DaemonSet under `set -e` if it finds a typo:
...
+**Trade-off:** `/sys/devices` is mounted whole, which hides the host's other
+device classes — CPU topology among them — from served containers. Narrowing it
+to the profile's root complexes is not possible: the runtime would have to create
+the mountpoint, and sysfs is read-only in a container, so container creation
+fails outright rather than degrading. Profiles routinely declare root complexes
+the node does not have. [#689](https://github.com/NVIDIA/k8s-test-infra/issues/689)
+tracks removing the trade-off.
+
+A workload that needs the node's real device tree belongs in
+`nri.excludedNamespaces`, or should not request a GPU through the
```
ArangoGutierrez: nri.excludedNamespaces only reaches the NRI plugin's --excluded-namespaces flag, and that plugin injects nvml-mock.nvidia.com/gpu, which carries no sysfs mounts. The /sys/devices mount comes from the nvidia.com/gpu spec the container toolkit resolves, so excluding a namespace from NRI leaves it in place.

internal/pcisysfs/render.go
```diff
@@ -35,18 +46,26 @@ type Options struct {
...
+			return nil
+		}
 		return errors.New("pcisysfs render: Output is required")
 	}
 
+	// A profile declaring no PCI devices means an empty tree, not the previous
+	// profile's: leftovers would go on being served at the kernel paths as if
+	// the node still simulated those GPUs.
+	if empty {
+		return prune(o.Output, &PCIeTopology{})
```
ArangoGutierrez, line 61: The Options.Topology field doc still says Render is a no-op for a nil or empty topology, which is now the destructive path. The Output doc directly below it was updated for this same change.

roma-glushko (dismissed):

internal/agent/cdi/spec.go
giuliocalzo: Confirmed, and it is the worse half of the pair: `Discard` removed `h.Root/sys` itself, so both mount sources came back as new inodes and a container that outlived the restart was left holding the removed ones — reading an empty tree that no restage could ever reach, with `gpu.mode` back to `unknown` until the pod is recreated.

Fixed in d3c1da2d. `pcisysfs.Clear` empties `sys/devices` and `sys/bus/pci/devices` in place instead, so the mount surface never moves. It is not `gpudriver`'s leaf list, though: the leaves here are a rendered tree whose shape changes with the profile, so `Clear` reuses the renderer's own prune with an empty topology and then takes the entries prune deliberately spares — the staged DMI among them, since a teardown owns all of it where a profile change does not.

Pinned in `TestDiscard_KeepsTheDirectoriesTheCDISpecMounts` with `os.SameFile` across the teardown, and in `TestClear_EmptiesTheServedDirectoriesInPlace`, so the distinction between emptying and replacing cannot quietly regress into `RemoveAll` again.

internal/agent/cdi/spec_test.go
giuliocalzo: You are right, and I checked it the way you described: deleting the `pciSysfsMounts` append leaves this test green. `mountByContainerPath` returns `(cdiMount, bool)`, so `_, devices :=` bound the flag and the assertion compared `false` to `false`.

Comparing the two `cdiMount` values would not work — the pair is two different mounts, `/sys/devices` and `/sys/bus/pci/devices`, so they are never equal. What the test actually names is that neither is emitted without the other, so in d3c1da2d it asserts that in both directions: a state with a rendered tree must serve both, one without must serve neither.

Verified it discriminates now by emitting only `PCIDevicesRelPath` and re-running: `--- FAIL ... /a_rendered_tree_serves_both: the hierarchy those symlinks point into`.

docs/helm-chart.md
giuliocalzo: Both true, and the second is on me: I added reconciliation in this PR and left the rules above it describing the opposite. There is no validation anywhere — `engine.validateYAMLConfig` checks version, driver version and duplicate device indices, and nothing under `internal/agent` or `internal/pcisysfs` looks at `pcie_topology` at all — so a typo reaches `/healthz` as success.

Rewritten in d3c1da2d to say that nothing validates the block and then describe what reconciliation silently absorbs: a BDF no device claims is dropped along with any root it empties (the same path `gpu.count` takes), a BDF under two roots stays under the first, and a device no root lists is still rendered under the root its address implies with `numa_node -1`.

The two format rules went with them, since nothing enforces those either — `rootIDForBDF` falls back to the default root for a BDF it cannot split rather than rejecting it.

docs/helm-chart.md
giuliocalzo: Correct — I wrote an escape hatch that does not exist. `nri.excludedNamespaces` only reaches the plugin's `--excluded-namespaces`, and the plugin injects `nvml-mock.nvidia.com/gpu`, which carries no sysfs mounts. The `/sys/devices` mount comes from the `nvidia.com/gpu` spec the toolkit resolves, so an excluded namespace still gets it the moment a pod requests a GPU.

Corrected in d3c1da2d: not requesting `nvidia.com/gpu` is the only way out, and the text now says why `nri.excludedNamespaces` is not one.

internal/pcisysfs/render.go
giuliocalzo: Fixed in d3c1da2d. `Topology` now says nil or empty renders an empty tree rather than leaving the previous one served, which is what the `Output` doc below it already said.

internal/agent/state.go
```diff
@@ -26,6 +27,170 @@ type State struct {
...
+	index := make(map[string]int, len(roots)+len(orphans))
+	for i, rc := range roots {
+		index[rc.ID] = i
+	}
+
+	for _, bdf := range orphans {
+		id := rootIDForBDF(bdf)
+		if i, ok := index[id]; ok {
+			roots[i].DeviceBDFs = append(roots[i].DeviceBDFs, bdf)
+			continue
```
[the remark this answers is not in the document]
giuliocalzo: The observation is accurate, unlike the two earlier P1s on this file: `pci0000:00` holding `0000:07:00.0` is exactly the shape our own profiles use, so a device that root forgot renders under `pci0000:07`, which is not a root complex that node has. Keeping it as a known limitation rather than fixing it, for two reasons.

There is nowhere better to put it. The symlink has to resolve somewhere, and the three options are: leave the device out of the tree, which is the failure this PR exists to fix — a GPU NVML reports that no consumer can resolve from its BDF; fold it into a declared root, which is the previous behaviour this same reviewer flagged, and which hands a consumer a specific wrong root *and* a real NUMA node it will not re-derive; or name a root after the device's own bus and report `numa_node -1`. Only the third says "I do not know where this sits" in an encoding consumers already handle, and it is self-consistent: the symlink target matches where the device is rendered.

It also only fires for a profile that contradicts itself — a BDF in `devices[]` that its own `pcie_topology` omits. No shipped profile does that, and `TestPCITopology_UnplacedDeviceJoinsTheRootItsAddressNames` pins that a device whose address names a declared root joins it rather than getting a synthetic one, so the synthetic root is the last resort and not the common path.

What is fair is that it happens silently. That is the same gap @ArangoGutierrez raised on the docs, where I had claimed validation that does not exist; d3c1da2d documents the behaviour, and a warning at stage time for a topology that omits an active BDF would be the real fix. Happy to add it here if reviewers want it rather than in the follow-up.

giuliocalzo:
@faganihajizada — want to make sure I act on the right thing, because I think your review may be aimed at a mount this PR does not add.

This PR mounts exactly two paths, both from the `nvidia.com/gpu` CDI spec: `/sys/devices` and `/sys/bus/pci/devices`. It does not touch `/sys/module`, which is a sibling of `/sys/devices` rather than a child, so a served container keeps all 111 of those entries and their `parameters/` intact. If the 106-directory loss you measured comes from work of your own that mounts `/sys/module`, I agree the objection applies there and that #689's retarget does not generalise — `/sys/bus/pci/devices` entries are kernel-made symlinks whose targets can be pointed into the overlay, and `/sys/module/<name>` is the data itself, as you say.

Where the objection does land on this PR is `/sys/devices`, which is mounted whole. A served container loses CPU topology, and `/sys/class/net/<iface>` dangles because it is a symlink into `devices/virtual/net`. That is called out as a trade-off in the changelog and in `docs/helm-chart.md`, with #689 as the tracking issue, and #689 already carries the retarget proposal along with the open questions on it — whether the DRA driver still extracts the PCIe root from a path outside `/sys/devices`, and whether the overlay root is reachable from the CDI spec.

Your DMI point matches what this PR does, for the same reason: kind binds three DMI paths, and those live at `devices/virtual/dmi/id`, so they fall under this mount and had to be reproduced. That reproduction is exactly what #689 would let us delete.

So: is the changes-requested asking me to land the #689 retarget in this PR rather than after it, or was it meant for a different change? Happy to do the former, but it is a design swap with an unverified dependency on the DRA driver's `deviceattribute`, so I would rather not fold it in silently.

faganihajizada (dismissed):

internal/agent/state.go
```diff
@@ -31,6 +32,170 @@ type State struct {
...
+// kernel's pciDDDD:BB form. A BDF too malformed to split is left to the default
+// root, since the renderer needs somewhere to put it either way.
+func rootIDForBDF(bdf string) string {
+	domain, rest, ok := strings.Cut(bdf, ":")
+	if !ok {
+		return DefaultRootComplexID
+	}
+	bus, _, ok := strings.Cut(rest, ":")
+	if !ok {
+		return DefaultRootComplexID
```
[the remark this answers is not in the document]
giuliocalzo: Valid, and fixed in 23a56d11.

You are right about the mechanism: `rootIDForBDF` fell back to `DefaultRootComplexID` for a `bus_id` it could not split, every shipped profile declares `pci0000:00`, so `adopt` found it in the index and appended the device to that real root. The malformed address inherited a NUMA node the profile never assigned it — precisely the borrowed locality the earlier fix set out to remove. The fallback predated that design and I did not revisit it.

Checking it turned up a second consequence of the same root cause. A `bus_id` also becomes a path component: `renderRootComplex` joins it under `Output`, so a value carrying separators or parent references renders outside the tree. Before the fix, `bus_id: "../../../../escaped"` had `Render` create a directory beside the overlay and return `nil`:

```
Render err: <nil>
ESCAPED: wrote outside Output at /var/folders/.../001/escaped
```

It needs a crafted `customConfig`, which is already trusted input, so no privilege boundary is crossed — but the agent is a privileged DaemonSet writing into a host path, so it is worth closing rather than leaving.

Both problems are the same gap: nothing checked that a hand-authored `bus_id` is an address. Taking your first option, since a string that is not an address is one no consumer can look up either:

- `validBDF` requires the kernel's `DDDD:BB:DD.F` form, and `activeBDFs` leaves anything else out of the topology entirely. This also covers the 8-digit domain from `nvmlPciInfo_t.busId`; the profile field carries the 4-digit form, as `busIdLegacy` does.
- `rootIDForBDF` loses its fallback along with the reason it existed, and now documents that it requires a validated BDF.
- The renderer keeps a backstop of its own, rejecting a root-complex ID or BDF that is not a single path component, so the agent's filter is not the only thing between a profile and the host filesystem.

Covered by `TestPCITopology_DropsAMalformedBDF`, `TestPCITopology_DropsABDFThatWouldEscapeTheTree` and `TestRender_RejectsANameThatEscapesOutput`. `docs/helm-chart.md` now lists the drop with the other reconciliation outcomes, so it is not silent-and-undocumented.

faganihajizada (approved):

PR #745, opened 2026-08-27
chore(greptile): tune AI review output to the project's review standards

ArangoGutierrez:
## What This PR Does

Adds a `.greptile/` configuration directory so the Greptile bot reviews this
repository the way a maintainer would.

- `config.json` turns off the generated wrapper around each review (tool
  header, confidence score, changed-files table, rendered flowchart, footer),
  sets `strictness: 3` with `commentTypes: ["logic"]`, and carries eight
  glob-scoped project rules.
- `rules.md` states what Greptile should not raise, listing the linters that
  already run on every pull request.
- `files.json` points the reviewer at `AGENTS.md`, `CONTRIBUTING.md`,
  `.golangci.yml` and `docs/architecture.md`.
- `hack/verify-greptile-config.sh` plus `make verify-greptile` validate the
  configuration.

Nothing about merge gating changes. `statusCheck` is on so a review leaves a
visible check, but it is not added to the required-checks list, and
auto-approval is off, so a human still holds the approve bit.

## Why

Greptile output on a recent pull request carried a summary header, a
confidence score, a changed-files table and a flowchart around a single
substantive finding. That finding was good and named the symbols involved;
the wrapper around it was what made it easy to miss.

The `strictness` and `commentTypes` settings keep Greptile off ground
`golangci-lint` already covers. That linter set is wide here (revive,
staticcheck with all checks, testifylint with enable-all, cyclop, nestif,
bodyclose, contextcheck and others), so a second probabilistic pass over the
same ground yields duplicate comments and trains reviewers to skim.

The eight rules encode failures that build, vet, test and lint were all green
for at the time. The clearest is the bridge rule: a go-nvml bump that adds an
NVML entry point needs `make gen` and a hand-added C type in `nvml_types.h`,
and a stale `stubs_generated.go` still links under `-buildmode=c-shared`, so
the break appears only when a consumer dlopens the library with `RTLD_NOW`.

The verifier exists because Greptile fails open on a rule scope that stops
matching. After a package move the rule quietly stops applying and nothing
reports it. The check fails when a scope glob or a context path matches
nothing. It earned its place while this branch was being written: it caught
three ignore patterns pointing at gitignored directories that can never appear
in a diff, and one (`**/*_generated.go`) that would have hidden
`pkg/gpu/mocknvml/bridge/stubs_generated.go` from the rule that needs to see it.

## Testing

Run against the final commit:

- `make test` passes.
- `make lint` passes: golangci-lint reports 0 issues, govulncheck reports no
  vulnerabilities the code calls.
- `make verify-greptile` passes 49 of 49 checks.
- `make gen` produces no diff, so generated code is unaffected.
- The verifier was mutation-tested. A dead scope glob, invalid JSON, an
  inserted em dash and an invalid severity each turn it red, and the
  unmutated tree stays green.

Behaviour is not yet verified against a live review. Greptile reads
configuration from the source branch of a pull request, so this pull request
is itself the first test of the settings.

## Breaking Changes

None. No existing source is modified; the Makefile gains one standalone
target.

## Checklist

## Open question for reviewers

`ignorePatterns` is documented as a string in `.gitignore` syntax in the
`greptile.json` reference but is not typed in the `.greptile/` reference. If
Greptile comments on a file under `vendor/`, the field wants an array instead
and this is a one-line fix.

### DIFF
--- a/.greptile/config.json
+++ b/.greptile/config.json
@@ -1,0 +1,96 @@
+{
+  "strictness": 3,
+  "commentTypes": [
+    "logic"
+  ],
+  "triggerOnUpdates": true,
+  "triggerOnDrafts": false,
+  "statusCheck": true,
+  "hideFooter": true,
+  "shouldUpdateDescription": false,
+  "ignorePatterns": "vendor/\ntests/mocknvml/vendor/\nLICENSES/\n**/zz_generated*.go\n",
+  "summarySection": {
+    "included": true,
+    "collapsible": false,
+    "defaultOpen": true
+  },
+  "issuesTableSection": {
+    "included": false
+  },
+  "confidenceScoreSection": {
+    "included": false
+  },
+  "sequenceDiagramSection": {
+    "included": false
+  },
+  "instructions": "Write about the code, never about the review. Do not describe your own pass over the change: no \"I checked\", no \"the changes look good\", no \"nothing blocking remains\", no \"this pull request appears safe to merge\", and no counting of your own findings. Do not name the tooling or any configuration file behind the review.\n\nState each finding in one or two sentences: what breaks, then what to do about it. Name the symbol, invariant, or test involved, because that specificity is what makes a comment worth reading and what makes it read as a person who looked at the change. Do not put a severity word or tag in the comment text. Do not repeat an inline comment in the summary; the summary carries only findings that have no inline anchor.\n\nDo not use emojis, em dashes, or filler adjectives. When nothing blocks the change, open with one short sentence naming what this change specifically improves. A sentence that would fit any pull request adds nothing.",
+  "rules": [
+    {
+      "id": "nvml-bridge-types",
+      "rule": "When a change adds a new nvml entry point to vendor/github.com/NVIDIA/go-nvml/pkg/nvml/nvml.go, pkg/gpu/mocknvml/bridge/stubs_generated.go must be regenerated with `make gen`, and pkg/gpu/mocknvml/bridge/nvml_types.h must gain the matching C type definitions by hand with full field layouts rather than opaque forward declarations. The generator does not touch that header. A stale stubs_generated.go omits the //export directives for the new entry points, and the shared library still links because -buildmode=c-shared does not pass --no-undefined, so build, vet, test and lint all stay green. The failure appears only when a consumer dlopens the library with RTLD_NOW.",
+      "scope": [
+        "pkg/gpu/mocknvml/bridge/**",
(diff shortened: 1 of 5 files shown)


.greptile/config.json
```diff
@@ -0,0 +1,96 @@
...
+      "id": "workflow-tag-paths-filter",
+      "rule": "In a workflow triggered on push, do not combine a tags filter with a paths filter in the same on.push block. A tag push carries no file diff, so the paths filter suppresses the run and the tag-triggered job never fires.",
+      "scope": [
+        ".github/workflows/**"
+      ],
+      "severity": "high"
+    },
+    {
+      "id": "comment-intent",
+      "rule": "A code comment must explain why the code does what it does, where that reason is not obvious from the code itself. Flag a comment that restates what the following line does, and flag a comment that narrates the change being made rather than the long-lived behaviour a future reader needs.",
```
roma-glushko, line 79: The man after my own heart ❤️

giuliocalzo (approved):

.greptile/config.json
```diff
@@ -0,0 +1,96 @@
...
+    {
+      "id": "test-discrimination",
+      "rule": "Each assertion must compare against a literal or a value derived independently of the code under test. Recomputing the implementation's own logic inside the test asserts nothing. Flag a test that would still pass if the body of the function it covers were replaced with an empty return or a zero value, and flag a table case whose expected value is produced by calling the subject.",
+      "scope": [
+        "**/*_test.go"
+      ],
+      "severity": "high"
+    },
+    {
+      "id": "helm-chart-version-coupling",
```
roma-glushko, line 62: We should encourage to write helm unit tests for substantial changes to the helm chart. Make sense to make it as a rule?

.greptile/files.json
```diff
@@ -0,0 +1,22 @@
...
+      "path": "CONTRIBUTING.md",
+      "description": "Contribution workflow and the commit and pull request conventions this repository expects."
+    },
+    {
+      "path": ".golangci.yml",
+      "description": "The exact linter set that already runs on every pull request. Treat every finding these linters produce as already reported, and do not raise it again in review.",
+      "scope": ["**/*.go"]
+    },
+    {
+      "path": "docs/architecture.md",
```
roma-glushko, line 17: I think it should read all docs dir. The docs are our shared context between internal, external contributors and agents.

.greptile/config.json
roma-glushko, line 1: We should also encourage to write changelog for changes to make it consistent.

PR #746, opened 2026-08-27
fix(mocknvml): return ERROR_GPU_IS_LOST from nvmlEventSetWait for lost devices

harche:
## What
When a mock GPU is `lost` / `fallen_off_bus`, `nvmlEventSetWait` now returns
`NVML_ERROR_GPU_IS_LOST` right away, the same as a real driver after Xid 79.
Before, the wait only ever returned `TIMEOUT`.

A GPU that's configured as immediately lost (a bare `mode:` block, no
`after_calls`/`probability`) fails the wait on its own — so a client that only
loops on `nvmlEventSetWait` (the DRA driver's health monitor) sees the lost GPU
without having to call another device function first, matching real hardware.
When a normal device call has already tripped the GPU, the configured Xid is
still delivered once before the lost errors begin. The wait never trips the
injector and never advances an `after_calls`/`probability` gate.

## Why
The DRA driver decides a GPU's health from what this wait returns. Without the
`GPU_IS_LOST` return, its "GPU is gone → taint it" path couldn't be tested
against the mock.

## Testing
- `go test ./pkg/gpu/mocknvml/...` — all pass (Linux, go 1.26).
- dra-driver `test_gpu_health.bats` (mock NVML, 2-GPU kind) — all 6 pass.
- End-to-end on kind: injecting `fallen_off_bus` at runtime taints all GPUs
  `gpu.nvidia.com/gpu-lost`, marks the pod's claim Unhealthy, and keeps it
  Unhealthy past the kubelet health timeout via the heartbeat; clearing the
  override recovers to Healthy.


giuliocalzo (approved):
/lgtm

ArangoGutierrez (approved):
Thank you very much for your second contribution! Looking forward for more!

PR #751, opened 2026-08-28
NRI Plugin Refactoring

roma-glushko:
## What This PR Does

Refactoring NRI plugin to:
- move it's codebase to interna/nri package
- use `urfave/cli` as CLI framework
- use a unified, common logging setup
- separate containerd machinery from the actual injection logic
- break down injection logic into a clear multi-step pipeline that is easy to unit test

Since we have three components that behave like servers (Node Agent, Control Plane, NRI), I have promoted the health server logic into its own package reused in all three.

## Why

It's stemmed from MEP0003. I'm keeping NRI in line with the recent restructuring and revisiting it's codebase to make sure it's clearly organized.

## Checklist

---


cmd/nri-plugin/main.go
```diff
@@ -0,0 +1,291 @@
...
+
+func newCLI() *cli.Command {
+	defaults := nri.DefaultConfig()
+	return &cli.Command{
+		Name:  "nri-plugin",
+		Usage: "Mokka NRI plugin — injects the nvml-mock overlay into containers as they are created",
+		// Grouped in the order the plugin uses them: how the process behaves,
+		// how it attaches to the runtime, then one group per injection step.
+		Flags: slices.Concat(
+			processFlags(),
```
roma-glushko, line 40: Too many flags, had to group them like this 😃

deployments/nvml-mock/helm/nvml-mock/values.yaml
```diff
@@ -125,6 +125,9 @@ allocationWatcher:
 
 nri:
   enabled: false
+  logging:
```
ArangoGutierrez, line 128: nodeAgent.logging and controlPlane.logging both declare level and format enums in values.schema.json, but nri.logging does not, so `--set nri.logging.level=verbose` templates cleanly and then the plugin exits 1 on ParseLevel. Copying that schema block under properties.nri keeps a typo failing at template time instead of on every node.

internal/agent/agent.go
```diff
@@ -59,21 +62,28 @@ func New(cfg Config) *Agent {
...
+func (a *Agent) Liveness() health.Probe { return *a.live.Load() }
+
+// Readiness aggregates every simulator's readiness and attributes the result
+// per simulator, so a red /readyz names which one is not serving.
+func (a *Agent) Readiness() health.Probe {
+	probe := health.Probe{OK: true, Components: make(map[string]health.Probe, len(a.simulators))}
 	for _, sim := range a.simulators {
-		m[sim.Name()] = sim.Ready()
+		ready := sim.Ready()
+		if !ready {
```
ArangoGutierrez, line 81: Deleting these two lines leaves the agent and health suites green, so /readyz would answer 200 for a node whose simulator is not ready. TestReadyz_OneNotReady covered exactly this rule before the move; a case in agent_test.go driving Readiness with one simulator down would put it back.

ArangoGutierrez (dismissed):
Thanks for this. Splitting adjust.go into one file per injection step makes the pipeline much easier to follow, and the step order, the three-arm GPU switch and every fail-open branch come through the move unchanged.

The behaviour moved with the code; what is left to tidy sits around the refactor rather than inside it.

- The shipped binary changes name (nvml-mock-nri to nri-plugin), all fourteen env-var bindings move to the MOKKA_NRI_ prefix, the --plugin-name default becomes mokka-nri-plugin, and pkg/nri/nvmlmock stops being importable, with nothing recorded under [Unreleased]. Anyone configuring the plugin through the environment falls back to defaults after this with no signal. (CHANGELOG.md:8) https://github.com/NVIDIA/k8s-test-infra/blob/e5b2a4c8b57d7ebb448bfaad6a1d93fe0d3fb359/CHANGELOG.md#L6-L10
- The DefaultConfig comment says it returns the identity the chart deploys with, but the chart passes --plugin-name=nvml-mock from values.yaml while the default here is mokka-nri-plugin. Rewording it keeps the next reader from trusting the default to match a real deployment. (internal/nri/config.go:27)

giuliocalzo (dismissed):

deployments/nvml-mock/helm/nvml-mock/values.yaml
roma-glushko: This should help https://github.com/NVIDIA/k8s-test-infra/pull/751/commits/a5d8a51b4f7c8ffe77858e3eefeaab4edb676fae ✅

internal/agent/agent.go
roma-glushko: Covered the test gap in https://github.com/NVIDIA/k8s-test-infra/pull/751/commits/419e5478a15edc4e88e821f2f5bae89167c2a1f9 ✅

giuliocalzo (approved):

PR #748, opened 2026-08-28
feat(node-agent): Mokka Node Agent IB simulator

roma-glushko:
## What This PR Does

- Porting InfiniBand simulator to Node Agent
- Move IB shims to the shims dir
- Move IB daemon code to `internal/ib/*` package
- Add a set of path sugar methods to Host struct to simplify path building across simulators
- Revisited bundle-ib-tools.sh script
- Dropped mock-ib CLI
- Dropped manual register peers mode used in tests in order to simplify the system
- Implemented IB sysfs tree reconcilation
- Moved common filesystem operations into `internal/fsutils` package

This is an interim refactoring for IB. With Mokka Control Plane, most of the IB daemon code can be delegated to the Control Plane.

## Why

Related to MEP0003.

## Checklist

- [ ] Commits are signed off (`git commit -s`)
- [ ] Tests pass (`go test -v -race ./...`)
- [ ] Linter passes (`make lint-fix`)
- [ ] New code has SPDX license headers
- [ ] Documentation updated (if applicable)
- [ ] CHANGELOG.md updated (if user-facing change)

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

PR #756, opened 2026-08-31
chore!: remove vendor/ and resolve dependencies through the Go proxy

### LINKED ISSUE #753: Drop revendoring packages in favor of the Go proxy
Instead of committing a bulk of 3rd-party directly into our codebase, we should use DGXC Go proxy in order to ensure supply chain security.

References:
- https://github.com/NVIDIA/aicr/issues/2372
- https://github.com/NVIDIA/aicr/pull/2388

giuliocalzo:
## What This PR Does

- drop vendor cache from our git repository
- installes NVIDIA DGXC Proxy action which is completely secretless and based on OIDC
- uses NVIDIA DGXC Golang Proxy to pull in dependencies through in dockerfiles and CI pipelines
- Public `proxy.golang.org` is used during development

## Why

Closes #753.

## Trade-off

Image builds and `make gen` now need egress to `proxy.golang.org`, where `vendor/` let them run offline. Bounded by every dependency being a public module and the proxy being CDN-distributed. Reverting is `git revert` plus `go mod vendor`; nothing here is one-way.

## Checklist

- [x] Linter passes (`make lint-fix`) — see the `govulncheck` toolchain note above
- [x] Documentation updated (if applicable) — CONTRIBUTING, `docs/development.md`, `pkg/gpu/mocknvml/README.md`, plus comments that cited now-dead `vendor/` paths


cmd/generate-bridge/nvmlsource.go
```diff
@@ -0,0 +1,83 @@
...
+	return input, header, nil
+}
+
+// moduleDir reports where a module is unpacked in the local cache, downloading
+// it first if necessary.
+//
+// `go mod download -json` rather than `go list -m -f '{{.Dir}}'`: the latter
+// reports an empty Dir for a module that is not yet in the cache, which on a
+// cold cache would silently yield a relative path of "pkg/nvml/nvml.go".
+func moduleDir(module string) (string, error) {
```
roma-glushko: Let's just assume it's already downloaded and do downloads in makefile/docker prior to running bridge-gen. Otherwise, we do what's supposed to be done by task runner.

pkg/network/mockib/render/ping_integration_test.go
```diff
@@ -53,7 +53,7 @@ func TestIbping_Loopback_Integration(t *testing.T) {
 
 	repoRoot := filepath.Join(wd, "..", "..", "..", "..")
 	daemonBin := filepath.Join(t.TempDir(), "mock-ib")
-	build := exec.Command("go", "build", "-mod=vendor", "-o", daemonBin, "./cmd/mock-ib")
```
roma-glushko: This is not actual anymore. I have just merged my IB refactoring (https://github.com/NVIDIA/k8s-test-infra/pull/748) which totally removes any binary building in tests 😌

Makefile
```diff
@@ -115,50 +115,31 @@ build: ## Build all CLIs
...
 	@make -C shims/libpcisysfs
 
 .PHONY: test
 test: ## Run unit tests with race detection and coverage
-	@$(GO_CMD) test -v -race -coverprofile=coverage.out -covermode=atomic $$(go list ./... | grep -v vendor)
-
-.PHONY: vendor
-vendor: ## Refresh top-level go.mod / vendor / verify
-	@echo "Refreshing go.mod.."
-	@go mod tidy
```
roma-glushko, line 137: I agree that the `vendor` command is not actual anymore, but we should move go mod  commands to lint* commands:

```
@go mod tidy
@go fmt ./...
@go vet ./...
```

They are still needed to verify go.mod consistency.

Makefile
```diff
@@ -115,50 +115,31 @@ build: ## Build all CLIs
...
-
-.PHONY: vendor-check
-vendor-check: vendor ## Fail if go.mod / go.sum / vendor are out of sync with HEAD
-	@git diff --quiet HEAD -- go.mod go.sum vendor
+	@$(GO_CMD) test -v -race -coverprofile=coverage.out -covermode=atomic ./...
 
 .PHONY: modules
-modules: | .mod-tidy .mod-vendor .mod-verify ## Tidy / vendor / verify every sub-module
+modules: | .mod-tidy .mod-verify ## Tidy / verify every module
 .mod-tidy:
```
roma-glushko: Not sure if we need this complicated command if simple go mod tidy should do the trick. We don't have any other go.mod files in the project.

Makefile
```diff
@@ -115,50 +115,31 @@ build: ## Build all CLIs
...
 	done
 
-.mod-vendor:
-	@for mod in $$(find . -name go.mod -not -path "./testdata/*" -not -path "./third_party/*" -not -path "./deployments/*"); do \
-		echo "Vendoring $$mod..."; ( \
-			cd $$(dirname $$mod) && go mod vendor \
-			) || exit 1; \
-	done
-
 .mod-verify:
```
roma-glushko: The same goes here.

CONTRIBUTING.md
```diff
@@ -29,13 +29,13 @@ cd k8s-test-infra
 go version  # must be >= 1.25
 
 # Run unit tests
-go test -v -race $(go list ./... | grep -v vendor)
+go test -v -race ./...
```
roma-glushko: we should advertise make commands not bare go commands. This is related to all commands in this file.

CONTRIBUTING.md
```diff
@@ -29,13 +29,13 @@ cd k8s-test-infra
...
 
 # Run linter
 golangci-lint run -v --timeout 5m
```
roma-glushko: The same is here. Use make lint-fix

ArangoGutierrez:
why draft?

giuliocalzo:
> why draft?

still working on it ;)

ArangoGutierrez:
/ok to test [07d056b](https://github.com/roma-glushko/k8s-test-infra/commit/07d056b3fd61e32537101531992dfb49b2d0cbab)

ArangoGutierrez (approved):
/lgtm

ArangoGutierrez (commented):
Thanks for this. Dropping 1.78M lines of vendor and putting a digest-checked proxy kit in its place is a big cleanup, and minting the credential per job over OIDC instead of parking a static token in repo secrets is the right shape.

The proxy path itself is carefully built: the token stays in the GOAUTH helper rather than GITHUB_ENV, check-goproxy.sh rejects both a pipe separator and a leading public entry, and enforced mode on the release build makes the coverage claim self proving. What the vendor removal left behind elsewhere is where the work remains. A few smaller notes:

- The action publishes a goproxy output and this hands BuildKit a hand-copied literal of the same URL instead, repeated again in nvml-mock-publish.yaml and in the probe's expected value. A kit upgrade that moves the default host or repo key would leave the build arg pointing somewhere the GOAUTH helper prefix does not cover. Giving the configure step an id and referencing its output keeps the two in step. (.github/workflows/nvml-mock-e2e-go.yaml:124)
- This block is the natural home for the Go version the job resolves two steps up into steps.golang.outputs.version, which today feeds only setup-go. Without it the published image builds on the Dockerfile default of 1.26.6 while hack/golang-version.sh reports 1.26.7, and the e2e build already passes GOLANG_VERSION. The new GOTOOLCHAIN=local also removes the auto-upgrade that would have absorbed a future go.mod bump past that default. (.github/workflows/nvml-mock-publish.yaml:95)
- lint picks up the vendor scan and hack/check-depproxy-kit.sh while lint-fix picks up only go mod tidy, so this help text no longer holds. CONTRIBUTING.md now sends contributors to lint-fix, which means editing the digest-pinned action.yml in place gives a clean local run and a red CI. (Makefile:87)
- docs/pilot/FINDINGS.md does not exist in this repo, and the vendored kit cites it nine times, five of them inside ::error:: output a contributor debugging a 403 will actually read. The file is digest-pinned so it cannot be corrected in place; a short mapping in UPSTREAM.md saying where those F-numbers live would keep the pointers useful. (.github/actions/setup-dgxc-goproxy/action.yml:249)

REUSE.toml
```diff
@@ -18,6 +18,6 @@ SPDX-FileCopyrightText = "Copyright 2024-2026 NVIDIA CORPORATION"
 SPDX-License-Identifier = "Apache-2.0"
 
 [[annotations]]
-path = ["go.sum", "vendor/**", ".gitignore", ".editorconfig"]
+path = ["go.sum", ".gitignore", ".editorconfig"]
```
ArangoGutierrez, line 21: The vendor sweep reached this file but not .greptile/config.json, which still scopes the nvml-bridge-types rule on vendor/github.com/NVIDIA/go-nvml/** and still ignores vendor/ and tests/mocknvml/vendor/. make verify-greptile goes from 49/49 on the base commit to 46/49 here, and the rule that catches a go-nvml bump landing without a bridge regen now matches zero files, because such a bump only touches go.mod and go.sum.

Makefile
```diff
@@ -62,6 +62,20 @@ tools: ## Install static checkers & other binaries
 
 .PHONY: lint
 lint: tools gen-check ## Lint the source code
+	@echo "🧹 Tidying go.mod.."
+	@go mod tidy
+	@git diff --exit-code -- go.mod go.sum || { \
+		echo "ERROR: go.mod / go.sum are out of sync. Run 'make lint-fix' and commit the result."; \
+		exit 1; }
+	@echo "🧹 Checking that no vendor directory has reappeared.."
+	@found=$$(find . -type d -name vendor -not -path "./tmp/*"); \
```
ArangoGutierrez, line 71: This find only prunes ./tmp/*, so it descends into .worktrees/, the directory .gitignore reserves for the sibling worktrees this repo's workflow uses. Run from a checkout that has any, it returns 83 paths, 81 of them .worktrees/*/vendor from branches cut before this lands, and make lint then fails with a vendor-mode error naming directories that are not part of the build. Pruning .worktrees as well, or driving the check off git ls-files, makes it assert what it means.

Makefile
```diff
@@ -71,20 +85,35 @@ lint: tools gen-check ## Lint the source code
...
 	@echo "🧹 Vetting.."
 	@go vet ./...
 	@echo "🛡️ govulncheck.."
 	@$(BIN_DIR)/govulncheck -tags=e2e,integration ./...
 
+.PHONY: deps-verify
+deps-verify: ## Re-download every module into a throwaway cache and re-verify it against go.sum
+	@echo "🛡️ Verifying modules against go.sum.."
+	@cache=$$(mktemp -d); \
+	GOMODCACHE="$$cache" GOFLAGS=-mod=readonly go mod download; \
```
ArangoGutierrez, line 101: SHELLFLAGS is -o pipefail -ec, so set -e is live in every recipe and a failing go mod download aborts right here: lines 102 to 104 never run. The throwaway GOMODCACHE from mktemp -d is left behind on exactly the path the cleanup exists for, and the status capture is dead code. A trap on EXIT before the download, or wrapping the download in if !, restores both.

Makefile
```diff
@@ -62,6 +62,20 @@ tools: ## Install static checkers & other binaries
 
 .PHONY: lint
 lint: tools gen-check ## Lint the source code
+	@echo "🧹 Tidying go.mod.."
+	@go mod tidy
```
ArangoGutierrez, line 66: This gate tidies and diffs the root module only, and it replaces modules-check, which looped every find . -name go.mod and also ran go mod verify. tests/mocknvml/go.mod is the second module in the tree and nothing at this head tidies or verifies it, since dependency-integrity.yaml filters on the root go.mod and go.sum.

PR #761, opened 2026-09-01
fix(chart): correct GPU Operator flags in NOTES.txt

ArangoGutierrez:
## Problem

The chart's post-install NOTES print a GPU Operator command that does not work:

```
helm install gpu-operator nvidia/gpu-operator \
  --set driver.enabled=false \
  --set toolkit.enabled=true \
  --set devicePlugin.enabled=true
```

A user following it hit `Init:Error` on `nvidia-container-toolkit-daemonset`, with the
validator reporting `open /run/nvidia/driver/sys/bus/pci/devices: no such file or
directory` and pointing at `DISABLE_DEV_CHAR_SYMLINK_CREATION`, which is one of the
nine operand environment settings the `--set` form omits entirely.

Three separate defects in that block:

1. `toolkit.enabled=true` contradicts the configuration we actually verify.
   `tests/e2e/gpu-operator-values.yaml` sets `toolkit.enabled=false` and drives CDI.
2. It claimed the integration was UNTESTED. The `e2e-gpu-operator` job has covered it
   across six GPU profiles for some time.
3. It cannot carry what the install needs. The operands require
   `NVIDIA_DRIVER_ROOT=/var/lib/nvml-mock/driver`, and the validator needs
   `DRIVER_INSTALL_DIR`, `LD_LIBRARY_PATH`, `DISABLE_DEV_CHAR_SYMLINK_CREATION`,
   `NVIDIA_VISIBLE_DEVICES` and `WITH_WORKLOAD=false`. None fit a short flag list.

Two further problems found while fixing it: `--set validator.cuda.env[0].value="false"`
sends a boolean into ClusterPolicy's string `env` field, and the unquoted `env[0]` is a
zsh glob, so on a default macOS shell the whole command line aborts with
`no matches found` and nothing runs at all.

## Approach

Print a values-file form instead of a flag list, referencing
`tests/e2e/gpu-operator-values.yaml` by absolute URL so it is reachable by a reader who
installed the chart from the OCI registry and has no checkout. Drop the UNTESTED
banner. Add `--wait --timeout 600s` and a verification step, matching how the other two
options end. Mention the Kind containerd prerequisites and point at the demo for setup.

The verification command deliberately scans every node rather than `.items[0]`: the
Operator's operands do not tolerate the control-plane taint, so a single-node check
returns empty on a healthy install.

## Testing

- `make helm-tests`: 175/175 tests, 15/15 snapshots.
- Four assertions added to `notes_test.yaml`, because the previous assertion matched the
  substring `OPTION C: GPU Operator` and therefore passed against the old, broken block.
  Each new assertion was mutation-checked: restoring the parent blob makes them fail
  individually, and a mutant that removes only `--wait` from this section (leaving
  OPTION B's) fails exactly one of them.
- The `helm install ... -f <raw URL>` form was confirmed to fetch (rc=0), with a 404 on a
  bogus sibling path as the control.

## Breaking changes

None. Documentation output only.

## Stack

This is PR 1 of 3. It is the base of the stack and can merge on its own.

### DIFF
--- a/deployments/nvml-mock/helm/nvml-mock/templates/NOTES.txt
+++ b/deployments/nvml-mock/helm/nvml-mock/templates/NOTES.txt
@@ -50,15 +50,34 @@ OPTION B: NVIDIA DRA Driver (requires DRA-enabled cluster)
 
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
 
-OPTION C: GPU Operator (UNTESTED — contributions welcome)
+OPTION C: GPU Operator
 
-  helm install gpu-operator nvidia/gpu-operator \
-    --set driver.enabled=false \
-    --set toolkit.enabled=true \
-    --set devicePlugin.enabled=true
+  The operator's driver and container-toolkit operands are replaced by
+  nvml-mock, so both are disabled and CDI carries the devices instead.
+  The overlay below carries that, plus the driver root the device plugin,
+  GFD and dcgm-exporter each need in their environment.
 
-  ⚠ This integration has no E2E test coverage. If you get it working,
-  please contribute the test back.
+  Cluster prerequisites (KIND): the node needs CDI enabled in containerd
+  and the nvidia runtime handler registered.
+  Walkthrough, including that containerd setup:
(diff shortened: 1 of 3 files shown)


giuliocalzo (dismissed):

roma-glushko (approved):
I would keep it DRY by simply leaving a link to the corresponding page in our documentation. Otherwise, we have two place to keep in sync.

PR #771, opened 2026-09-01
docs(demo): back the NVSentinel demo with the Percona MongoDB store

### LINKED ISSUE #757: [Feature]: Consider using NVSentinel's built in Percona DB for multi-arch support rather than a standalone database pod
### Problem statement

NVSentinel enables multiarch database support using percona as documented in https://docs.nvidia.com/nvsentinel/runbooks/mongo-db-bitnami-to-percona-migration/, it would be good to have that as the default option in the demo

### Proposed solution

Update NVSentinel values to use percona daatabase deployment

### Scope / non-goals

_No response_

### Audience

Operators / cluster admins

### Pre-submission checks

giuliocalzo:
## What This PR Does

Replaces the NVSentinel demo's hand-rolled standalone MongoDB with the `mongodb-store` subchart's Percona Server for MongoDB path:

- `nvsentinel-values.yaml`: sets `mongodb-store.useBitnami: false` + `usePerconaOperator: true`, and trims the Percona defaults for a laptop-sized Kind cluster — a single-member replica set (`unsafeFlags.replsetSize: true`), smaller CPU/memory requests, a 2Gi PVC, and no `mongodb_exporter` sidecar (there is no Prometheus Operator here).
- Deletes `mongodb.yaml` (239 lines of StatefulSet, cert-manager plumbing, and a replica-set init Job) and the `run.sh` steps that applied it, created the `MONGODB_URI` Secret, and waited on the external-MongoDB setup hook.
- Bumps `NVSENTINEL_VERSION` to `v1.21.0`, which is where the Percona flags live.
- Installs the mock into the `mokka` namespace, matching `docs/demo/standalone/demo.sh` and the namespace the rest of the docs standardized on, instead of this demo's own `nvml-mock-system`.
- Raises `HOT_TEMP_C` so the pinned temperature clears every profile's slowdown threshold (see below).
- README: swaps the "Why standalone MongoDB" section for one explaining the Percona choice, and updates the requirements and step list.

## Why

Closes #757.

The standalone MongoDB existed only because the default Bitnami store is amd64-only (`bitnamilegacy/*`) and its containers run Bitnami-specific startup scripts, so it cannot come up on Apple Silicon. The Percona path solves that upstream — every image on it is published for both `linux/amd64` and `linux/arm64`:

| Image | Platforms |
| --- | --- |
| `percona/percona-server-mongodb:8.0.12-4` | amd64, arm64 |
| `percona/percona-server-mongodb-operator:1.23.0` | amd64, arm64 |
| `percona/percona-backup-mongodb:2.11.0` | amd64, arm64 |
| `ghcr.io/rtsp/docker-mongosh:2.5.2` | amd64, arm64 |
| `docker.io/lachlanevenson/k8s-kubectl:v1.25.4` | amd64, arm64 |

Using the supported path also means the demo stops diverging from what NVSentinel documents in the [Bitnami to Percona migration runbook](https://docs.nvidia.com/nvsentinel/runbooks/mongo-db-bitnami-to-percona-migration/).

## The `HOT_TEMP_C` fix

While validating on arm64 the demo stopped detecting the fault, and the cause turned out to be unrelated to the datastore. `GpuThermalMarginWatch` fails only once the T.Limit margin drops *below* the per-GPU slowdown offset, so pinning the GPU at exactly its slowdown threshold leaves a margin of `0 °C` and the watch stays healthy. The old default of 90 °C was tuned for h100 (slowdown 87 °C), but gb300 — now the chart's default profile — slows down at exactly 90 °C, so the demo silently stopped tripping. DCGM confirmed it: the hot GPU read `TMPTR=90, GTLIMIT=0` while its cool neighbours read `~34`. The temperature now clears any profile's threshold, and the README explains that the comparison is strict.

## Verification

`helm template` against the v1.21.0 chart renders clean: no coalesce warnings, no Bitnami images.

On a live arm64 Kind cluster the Percona store comes up (a single-member `mongodb-rs0-0`) and the whole namespace reaches Ready. Phase 1 was confirmed by hand: heating the GPU past its slowdown threshold produced `GpuThermalMarginWatch=True: ErrorCode:GPU_TEMP_HW_SLOWDOWN_VIOLATION ... thermal margin -3°C below HW slowdown T.Limit` and NVSentinel cordoned the node, which also proves the events travel through the Percona datastore end to end.

Draft until a full unattended run covers Phase 2 (uncordon on cooldown).

## Checklist

- [ ] Linter passes (`make lint-fix`) — no Go changes
- [ ] CHANGELOG.md updated (if user-facing change) — demo-only change


lalitadithya (approved):

ArangoGutierrez (approved):
/lgtm

PR #773, opened 2026-09-01
fix(mocknvml): reset GPUs through nvidia-smi --gpu-reset instead of segfaulting

giuliocalzo:
## Summary

`nvidia-smi --gpu-reset` (`-r`) died with a SIGSEGV on every invocation, so an operator reaching for the most common GPU remediation got a bare exit 139 with no output.

**The fault was ours, not `nvidia-smi`'s.** The internal export-table catch-all for per-device calls ends by writing a zero count through `arg1`, and on slot 197 — reached only while resetting — `arg1` carries no count. That write sits *ahead* of the slot's debug print, so it produced no log line of its own: the trace stopped mid-sequence, which made the crash look like `nvidia-smi` walking into a post-reset re-attach sequence and dying on driver state a mock cannot supply. It isn't, and no driver ioctl is involved. Returning before that write is the whole fix — `nvidia-smi` then completes the reset entirely in-process.

That makes the reset worth *serving* rather than refusing. Slot 20 now calls into Go and clears the device's injected overrides under the same flock `nvml-mock-ctl` takes — the same mutation as `nvml-mock-ctl reset --gpu <n>`. A real reset clears the transient error state that made someone reach for it; on the mock, that state is exactly what was injected. So a remediation controller or runbook that already uses the standard GPU reset works unmodified, and the NVSentinel demo's heal-and-uncordon step can be driven either way.

Scope follows `nvidia-smi` rather than being decided in the mock: it calls the slot once per targeted GPU, so `-i <n>` clears one device and a bare `--gpu-reset` walks every device.

Two consequences of it being per-device are documented rather than worked around (`docs/nvml-mock-ctl.md`):

- State injected with `--gpu all` lives in a shared bucket no per-device reset clears. Verified to match `nvml-mock-ctl reset --gpu <n>` exactly, so this is inherited CLI semantics; `reset --gpu all` is still the way to clear it.
- A GPU with nothing injected is reset without writing the file or taking the lock, so a reset on a healthy GPU cannot fail on a read-only overrides mount.

## Test plan

- [x] `go test ./...` — 39 packages, 0 failures
- [x] `go vet -tags e2e ./tests/e2e/go/...` clean
- [x] `golangci-lint run` clean, including the cgo bridge linted on Linux
- [x] 7 new unit tests for the reset's file logic (scoping, healthy no-op, missing/corrupt file, shared `all:` bucket)
- [x] `GpuResetProblems` asserts the per-GPU report and trailer, not just a non-crashing exit — a reset that silently cleared nothing would still exit 0
- [x] New `assertGpuResetViaNvidiaSmi` runtime-control scenario injects a fault and heals it through `nvidia-smi`, the only coverage of these slots doing real work

Verified against a live kind cluster (gb300 profile, 4 GPUs):

```
$ nvml-mock-ctl fail --gpu 1 --mode lost
$ nvidia-smi -i 1 --query-gpu=temperature.gpu --format=csv,noheader
[GPU is lost]
$ nvidia-smi -r -i 1
GPU 00000000:0B:00.0 was successfully reset.
All done.                                        exit=0
$ nvidia-smi -i 1 --query-gpu=temperature.gpu --format=csv,noheader
63
```

Regression sweep: 19 other `nvidia-smi` views are byte-identical old-vs-new, while `--gpu-reset` goes from exit 139 to exit 0. A separate sweep confirms both reset slots are reset-exclusive — 18 other views reach neither.

### DIFF
--- a/pkg/gpu/mockctl/reset.go
+++ b/pkg/gpu/mockctl/reset.go
@@ -1,0 +1,74 @@
+// Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mockctl
+
+// Resetting one device's overrides as a single operation, for callers that have
+// no command dispatcher to hang it off. nvml-mock-ctl reaches Doc.Reset through
+// its generic load-mutate-validate-write flow; the mock NVML library serves
+// `nvidia-smi --gpu-reset` from a CGo callback with no such flow around it, and
+// needs the whole read-modify-write in one call.
+
+import (
+	"strconv"
+)
+
+// ResetDevice returns one device to its healthy baseline by removing its
+// override bucket, taking the same flock as every other writer of this file so a
+// concurrent injection cannot interleave with the reset.
+//
+// The index is deliberately not a Target: a Target can say "all", and the shared
+// `all:` bucket is not resettable per device. Doc.Reset leaves that bucket alone
+// for a device target (so state injected with `--gpu all` survives, and needs
+// `Doc.Reset(Target{All: true})` to clear), and a caller resetting a single GPU
+// must not be able to ask for something this cannot honour.
+//
+// A device with nothing injected is reset without touching the file at all — no
+// write and no lock. That is not just an optimisation: resetting a healthy device
+// is the common case, and skipping the write keeps it working where the overrides
+// are not writable. Only a reset that genuinely has state to clear can fail.
+func ResetDevice(path string, index int) error {
+	if path == "" {
+		// No config, so no injected state to clear: the device is already at its
+		// baseline and the reset has nothing to do.
+		return nil
+	}
+	if has, err := deviceHasOverrides(path, index); err != nil || !has {
+		return err
+	}
+
+	unlock, err := LockOverride(path)
+	if err != nil {
+		return err
+	}
+	defer unlock()
+
+	// Re-read under the lock: the peek above raced with any concurrent writer.
+	doc, err := Load(path)
+	if err != nil {
+		return err
+	}
+	doc.Reset(Target{Index: index})
+	return WriteAtomic(path, doc)
+}
+
+// deviceHasOverrides reports whether the document carries a bucket for the
+// device. A missing or empty file is not an error: it means nothing was injected.
+func deviceHasOverrides(path string, index int) (bool, error) {
+	doc, err := Load(path)
+	if err != nil {
+		return false, err
+	}
+	_, ok := doc.Devices[strconv.Itoa(index)]
+	return ok, nil
+}
--- a/pkg/gpu/mockctl/reset_test.go
+++ b/pkg/gpu/mockctl/reset_test.go
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package main
+package mockctl
 
 import (
 	"os"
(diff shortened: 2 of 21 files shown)


pkg/gpu/mocknvml/bridge/gpu_reset.go
```diff
@@ -0,0 +1,70 @@
...
+		return 0
+	}
+	index, ret := dev.GetIndex()
+	if ret != nvml.SUCCESS {
+		debugLog("[RESET] device %p reports no index (ret=%d)\n", handle, ret)
+		return 0
+	}
+	// engine.ConfigOverridePath resolves the same file the engine reads, so the
+	// reader and this writer can never disagree on which one is authoritative.
+	if err := mockctl.ResetDevice(engine.ConfigOverridePath(), index); err != nil {
```
ArangoGutierrez, line 64: ResetDevice writes overrides.yaml, but CDI binds the config directory to /etc/nvml-mock with `ro` while pointing MOCK_NVML_OVERRIDES at it, and NRI rbinds the overlay `ro`, so inside an injected consumer container LockOverride's O_CREATE on the .lock file hits EROFS and `nvidia-smi -r` returns NVML_ERROR_OPERATING_SYSTEM exactly when the GPU has state to clear. The mount comment says the directory is bound so nvml-mock-ctl's writes are visible inside the container, which is a read path, and both new e2e specs exec into the nvml-mock DaemonSet pod where the hostPath is writable, so the route a remediation controller actually takes stays uncovered.

Also at docs/nvml-mock-ctl.md:397.

pkg/gpu/mockctl/reset_test.go
```diff
@@ -0,0 +1,126 @@
...
+func TestResetDevice_EmptyPathSucceeds(t *testing.T) {
+	require.NoError(t, ResetDevice("", 0))
+}
+
+// A per-device reset leaves the shared `all:` bucket in place, matching
+// `nvml-mock-ctl reset --gpu <n>` exactly — the override schema cannot express
+// "drop the shared block for one GPU only". An operator who injected with
+// `--gpu all` needs a Target{All: true} reset; nvidia-smi will report a
+// successful reset while the shared block still masks the device.
+func TestResetDevice_LeavesSharedAllBucket(t *testing.T) {
```
ArangoGutierrez, line 105: The fixture carries no `devices:` map, so ResetDevice returns at the `!has` early exit and Doc.Reset never runs, which leaves this test indistinguishable from TestResetDevice_HealthyDeviceLeavesFileUntouched and unable to pin the behaviour its comment describes. Adding `d.All = nil` to Doc.Reset's device branch keeps the whole mockctl package green; giving the fixture a `devices: "0":` bucket and asserting doc.All survives turns it red on that same mutation.

ArangoGutierrez (commented):
Thanks for chasing this down to the dispatcher rather than working around it: returning before the zero write is the right fix, and pinning the untouched count in the internal-table test is what keeps it from coming back.

The reset reuses the existing write path instead of duplicating it: the same flock every other writer of overrides.yaml takes, the same Doc.Reset the CLI calls, and an early return that skips the write entirely for a device with nothing injected.

- The slot 20 branch sits ahead of the `arg1 == NULL || !mockInternalIsDeviceHandle(arg0)` guard, so an unregistered handle, or an engine still at initCount 0, now yields NVML_ERROR_OPERATING_SYSTEM where the catch-all's own comment says a probe has to see NVML_SUCCESS. Gating the branch on mockInternalIsDeviceHandle would keep a genuinely failed reset distinguishable from a call that was never a device call. (pkg/gpu/mocknvml/bridge/internal.go:112)

pkg/gpu/mockctl/reset_test.go
giuliocalzo: Confirmed and fixed in 1e300ef8. The fixture had no `devices:` map, so `ResetDevice` returned at the `!has` early exit, `Doc.Reset` never ran, and the byte-identical assertion held because the file was never opened for writing — a duplicate of `HealthyDeviceLeavesFileUntouched`.

The fixture now carries a `"0"` bucket alongside the `all:` block, so the reset goes through `Doc.Reset`, and the test asserts both halves: the device bucket is gone and `doc.All` survives intact. Verified against your mutation — adding `d.All = nil` to `Doc.Reset`'s device branch now fails this test, and the package is green again once it is reverted.

pkg/gpu/mocknvml/bridge/gpu_reset.go
giuliocalzo: Confirmed and fixed in 2fbb7512. The chain is exactly as you describe: CDI binds `overlayHostRoot + "/driver/config"` at `/etc/nvml-mock` with `ro` while `MOCK_NVML_OVERRIDES` points into it, NRI rbinds the overlay containing it `ro`, and `LockOverride`'s `O_CREATE` on the `.lock` file then fails with EROFS. Because a device with nothing injected skips the write entirely, the reset succeeded on healthy GPUs and failed on exactly the ones carrying the state an operator wanted cleared.

CDI now binds that directory `rw`. NRI keeps the overlay `ro` and layers a writable bind over just `driver/config`, so the mock library and `nvidia-smi` below it stay immutable in the container — the mount ordering is asserted too, since a `ro` overlay applied second would cover the narrower one. Both new unit tests fail against the previous `ro` options.

On the coverage gap: `setupStandaloneProfile` returns `firstNvmlPod`, so you were right that both existing specs run where the hostPath is writable. There is now an NRI spec that pins a temperature, runs `nvidia-smi -r -i 0` from inside the injected `gpu-agent` container, and requires both a clean exit and the override actually gone. Injection and reset are pinned to one node via `nvmlMockCtlOnNode`, since the overrides are node-local.

`docs/nvml-mock-ctl.md` no longer claims only the healthy half: it states that the reset works from an injected container, why the config directory is writable, and what read-only would have cost.

giuliocalzo:
All three review points addressed; details in the two threads, and this one covers the dispatcher item from the review body.

**`internal.go:112` — slot 20 ahead of the device guard** (26cbae8e). Correct as described: the branch ran before `arg1 == NULL || !mockInternalIsDeviceHandle(arg0)`, so a call carrying an unregistered handle, or one arriving with the engine at initCount 0, failed the lookup and returned `NVML_ERROR_OPERATING_SYSTEM` — the answer that guard's own comment rules out, since `nvidia-smi` aborts `topo -m` on any error from these slots.

The branch is now gated on `mockInternalIsDeviceHandle`, which is the same predicate the guard uses, so those calls fall through to `NVML_SUCCESS` and `NVML_ERROR_OPERATING_SYSTEM` keeps meaning "attempted on a real device and could not run" — the distinction you asked for. `internal/gpu_reset_non_device_handle` covers it and fails with `slot 20 returned 17 for a handle that is not a device` when the gate is removed.

Extracting `hasDedicatedTest` also drops the sweep loop back under the cyclomatic limit it crossed when the two reset slots were excluded from it.

Verification: 75 bridge tests pass (74 before), `go test ./...` clean, `golangci-lint run ./...` reports 0 issues, `gofmt` clean. Each of the three fixes was checked red-green against the exact mutation it guards.

pkg/gpu/mockctl/reset.go
```diff
@@ -0,0 +1,74 @@
...
+// `all:` bucket is not resettable per device. Doc.Reset leaves that bucket alone
+// for a device target (so state injected with `--gpu all` survives, and needs
+// `Doc.Reset(Target{All: true})` to clear), and a caller resetting a single GPU
+// must not be able to ask for something this cannot honour.
+//
+// A device with nothing injected is reset without touching the file at all — no
+// write and no lock. That is not just an optimisation: resetting a healthy device
+// is the common case, and skipping the write keeps it working where the overrides
+// are not writable. Only a reset that genuinely has state to clear can fail.
+func ResetDevice(path string, index int) error {
```
roma-glushko, line 40: Do we need to have it if nvidia-smi does the trick?

roma-glushko (approved):

PR #783, opened 2026-09-02
#740: Removed the nvml-mock container 

### LINKED ISSUE #740: MEP0003: Remove nvml-mock container after porting all logic to node agent
This is a cleanup task to remove all obsolete code related to the nvml-mock once node agent container is fully mirroring its logic.

roma-glushko:
This PR concludes the MVP of the Mokka Node Agent 🎉

## What This PR Does

- Removed the nvml-mock container from the mokka daemonset
- Reset override.yaml files like setup.sh did on node agent startup
- Dropping node labeling in the mokka daemonset
- updated references in docs & demos

## Why

A part of MEP0003.

## Checklist


deployments/nvml-mock/Dockerfile
```diff
@@ -99,4 +99,3 @@ RUN set -eux; \
     rm -rf /tmp/nvidia-utils /tmp/nvidia-utils-580*.deb \
            /usr/share/keyrings/nvidia.gpg /etc/apt/sources.list.d/nvidia.list; \
     apt-get purge -y --auto-remove curl ca-certificates gnupg2 patchelf
```
giuliocalzo, line 108: @roma-glushko as you wipe the setup.sh can you also remove the kubectl installation ?

https://github.com/roma-glushko/k8s-test-infra/blob/0d76717242f17f3030dac98b656f30bbe942b67a/deployments/nvml-mock/Dockerfile#L59-L66

roma-glushko: Great spot!

roma-glushko: Removed in https://github.com/NVIDIA/k8s-test-infra/pull/783/commits/05438d84f65007742000097742d4724ac991243d ✅ 

Thanks again!

tests/e2e/device-plugin-mock.yaml
```diff
@@ -21,13 +21,12 @@ spec:
...
-      # control-plane, where /var/lib/nvml-mock/driver is absent, so NVML init
-      # fails with ERROR_LIBRARY_NOT_FOUND and the rollout never completes.
-      # Single-node clusters label their control-plane too, so it still runs there.
+      # This mirrors the chart's own nodeSelector (local/nvml-mock.values.yaml);
+      # without it the DaemonSet, which tolerates all taints, also lands on the
+      # control-plane, where /var/lib/nvml-mock/driver is absent and NVML init
+      # fails with ERROR_LIBRARY_NOT_FOUND so the rollout never completes.
       nodeSelector:
-        nvidia.com/gpu.present: "true"
+        mokka.nvidia.com/type: sgpu
```
ArangoGutierrez, line 24: mokka.nvidia.com/type=sgpu is set only by the configs under local/kind, which is why the embedded copy stays fine in CI. The two walkthroughs in docs/helm-chart.md build clusters from a plain `kind create cluster` and from kind-multi-node-config.yaml, where no node carries it, so this DaemonSet schedules nothing and the `kubectl wait` on the next line burns its 120s.

tests/e2e/go/scenario_nfd_test.go
```diff
@@ -96,16 +91,11 @@ var _ = Describe("nvml-mock NFD label provenance", Label("nfd"), Ordered, Contin
...
-		// away: the nvml-mock container declares no readinessProbe, demoRelease
-		// sets maxUnavailable=100% so `helm --wait` returns on merely-scheduled
-		// pods, and pod selection waits only for phase Running — so nothing
-		// else orders this negative check after its setup.sh and it can pass
-		// vacuously. The node-agent container, which writes the feature file,
-		// is instead ordered by BeforeAll's WaitDaemonSetReady.
-		assertions.WaitNodeLabelsPresent(ctx, h.Kube, node,
-			[]string{gpuPresentLabel}, nfdLabelTimeout, nfdLabelPoll)
-
+		// Ordered by BeforeAll's WaitDaemonSetReady: node-agent is the only
```
ArangoGutierrez, line 94: /readyz aggregates sim.Ready(), and pcibus.Ready() flips at the end of Stage while the feature file is written in Apply, which reconcile runs after the whole Stage wave and after supervise, so the DaemonSet can report Ready before the file exists. Waiting on the file the next spec already reads would restore the ordering the removed barrier gave you.

deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml
```diff
@@ -833,16 +777,16 @@ tests:
       # node-agent: non-empty is what enables the simulator, and it is the path
       # the cdi simulator mounts into workloads.
       - contains:
-          path: spec.template.spec.containers[1].env
+          path: spec.template.spec.containers[0].env
```
ArangoGutierrez, line 773: Both contains blocks assert the same env var on containers[0], so the second cannot fail for the reason the title gives. With allocationWatcher off there is only one container to wire, so dropping the duplicate and renaming the test would say what it actually checks.

Also at deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml:825.

ArangoGutierrez (commented):
Collapsing the privileged sidecar and the node-patching ClusterRole into the agent is a real cut in blast radius, and the hasDocuments check on rbac.yaml is what keeps it from creeping back.

The port off setup.sh lines up path for path: both override locations, the GPU_COUNT cap and its warning, and the toolkit-ready reasoning kept in troubleshooting rather than deleted with the script.

- Step 4 of the device-plugin quick start still prints nvidia.com/gpu.present and tells the reader to expect true, which cannot happen now on a cluster running neither NFD nor GFD. NOTES.txt got the equivalent fix in this change. (docs/helm-chart.md:131) https://github.com/NVIDIA/k8s-test-infra/blob/4bdfb1aa4da47adade5426c6914b41dadf903015/docs/helm-chart.md#L129-L134

tests/e2e/go/scenario_nfd_test.go
roma-glushko: Switched those simulators to report ready when they are done all staging + publishing job:

https://github.com/NVIDIA/k8s-test-infra/pull/783/commits/945510dc9a3ca2057efe1ff36a9d1af63d143662

deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml
roma-glushko: Addressed in https://github.com/NVIDIA/k8s-test-infra/pull/783/commits/85143e69614f462509976e48c11e2aaf71fc6893 ✅

tests/e2e/device-plugin-mock.yaml
roma-glushko: Should be addressed in https://github.com/NVIDIA/k8s-test-infra/pull/783/commits/ba827b9478fa8e84396d1b2b3aef817b5fae1de3 ✅

giuliocalzo (dismissed):

giuliocalzo (approved):

PR #784, opened 2026-09-02
refactor(mocknvml): describe the host CPU as cpu.type, not grace_superchip

giuliocalzo:
## Summary

The profile schema named one CPU family in a key. `grace_superchip:` was
reachable only by claiming to be a Grace superchip, so a part sitting behind any
other host had no way to describe its pairing.

- `grace_superchip:` becomes a generic `cpu:` block whose `type` names the CPU
  family — `grace` on `gb200` and `gb300`.
- `enabled` is dropped; it said nothing that the presence of the block did not.
- `cpu_cores` and `cpu_memory_gb` become `cores` and `memory_gb`, losing a prefix
  they no longer need.
- A profile that omits the block now makes no claim about its host, which is the
  honest reading for a PCIe or SXM part that ships behind any CPU.

`type` is deliberately free-form, matching how the other profile enums
(`persistence_mode`, `compute_mode`) are handled.

Nothing in the engine ever read these values — they are descriptive metadata on
the same terms as `features:`, and `CPUConfig` carries the same doctrine note.

**Breaking:** a custom profile carrying `grace_superchip:` must be updated; the
key is no longer read.

## Changes

- `pkg/gpu/mocknvml/engine/config_types.go`: `GraceSuperchipConfig` → `CPUConfig`,
  reached through `DeviceConfig.CPU`.
- The four shipped Grace-Blackwell profiles: `gb200`/`gb300` under both
  `deployments/nvml-mock/helm/nvml-mock/profiles/` and `pkg/gpu/mocknvml/configs/`.
- `config_profiles_test.go` asserts `cpu.Type == "grace"` in place of the old
  `enabled` flag.
- Helm snapshots regenerated. The daemonset snapshot moves too, because its
  config checksum annotation hashes the profile.
- Changelog entry under `Unreleased` → `Changed`.

## Test plan

- [x] `go build ./...`
- [x] `go test ./pkg/gpu/mocknvml/...`
- [x] `helm unittest deployments/nvml-mock/helm/nvml-mock` (179 tests, 15 snapshots)
- [x] `make lint-fix` — golangci-lint and `go vet` clean
- [ ] CI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -186,6 +186,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   public NVML APIs.
 
 ### Changed
+- mocknvml: the `grace_superchip:` profile block is now `cpu:`, with `cpu_cores`
+  and `cpu_memory_gb` renamed to `cores` and `memory_gb` and the `enabled` flag
+  replaced by a free-form `type:` (`grace` on the gb200 and gb300 profiles). The
+  old key hardcoded one CPU family into the schema, so a profile for a part
+  behind any other host had no way to describe it. A profile that omits the
+  block makes no claim about its host, which is the honest reading for a PCIe or
+  SXM part. Profiles carrying `grace_superchip:` must be updated — the key is no
+  longer read.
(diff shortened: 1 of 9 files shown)


roma-glushko (approved):
Excellent 👍

PR #786, opened 2026-09-03
Moved the project over to zap

roma-glushko:
## What This PR Does

Using zap instead of slog in Mokka.

## Why

The team has agreed to standardize on zap.

## Checklist

---

### DIFF
--- a/deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml
+++ b/deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml
@@ -899,7 +899,7 @@ tests:
       # The watcher reads the kubelet's pod-resources socket, not the node's
       # sysfs, and the volume is the whole of /sys.
       - notContains:
-          path: spec.template.spec.containers[1].volumeMounts
+          path: spec.template.spec.containers[0].volumeMounts
           content:
             name: node-sysfs
             mountPath: /host/sys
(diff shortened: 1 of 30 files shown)


giuliocalzo (dismissed):

deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml
```diff
@@ -899,7 +899,7 @@ tests:
       # The watcher reads the kubelet's pod-resources socket, not the node's
       # sysfs, and the volume is the whole of /sys.
       - notContains:
-          path: spec.template.spec.containers[1].volumeMounts
+          path: spec.template.spec.containers[0].volumeMounts
```
ArangoGutierrez: This points at containers[0], the node-agent, which the contains assertion three lines up requires the node-sysfs mount on, so the two contradict and helm unittest fails on "should mount the node's sysfs read-only in the agent alone". The watcher is still containers[1] after #783 shifted the indices in this file, and the edit looks like it rode in on the rebase rather than with the zap work.

ArangoGutierrez (commented):
Thanks for taking this on: the slog to zap conversion is consistent across every call site, and cmd/nri-plugin now keeps a handle on its logger so it flushes on exit.

This moves the tree off log/slog onto zap: typed fields at each call site, zap.L() in the packages that used to log through the default logger, observer-based capture in the NRI warning tests, and internal/logging tests that assert the bytes written to stdout and the enabled levels rather than the handler type. The one open item sits in the daemonset helm test and is unrelated to the logging work.

deployments/nvml-mock/helm/nvml-mock/tests/daemonset_test.yaml
roma-glushko: Should be fixed ✅

giuliocalzo (approved):

PR #787, opened 2026-09-03
Added more logging to simulators & NRI

roma-glushko:
## What This PR Does

- adding more log statements to the important lifecycle events
- constitute the logging rules in agents.md

## Why

Making sure we have enough information logged in Mokka Daemon and NRI.

## Checklist

---

### DIFF
--- a/internal/agent/cdi/cdi.go
+++ b/internal/agent/cdi/cdi.go
@@ -63,7 +63,7 @@ func (s *Simulator) Discard(_ context.Context, _ *host.Host) error { return nil 
 
 // Apply writes /run/cdi/nvidia.yaml and /run/cdi/nvml-mock-nri.yaml.
 func (s *Simulator) Apply(_ context.Context, h *host.Host, state *agent.State) error {
-	zap.L().Debug("applying simulator", zap.String("simulator", name))
+	zap.L().Info("applying simulator", zap.String("simulator", name))
 	s.ready.Store(false)
 
 	if err := writeSpec(filepath.Join(h.Run, nvidiaSpecFile), buildNvidiaSpec(state)); err != nil {
@@ -80,7 +80,7 @@ func (s *Simulator) Apply(_ context.Context, h *host.Host, state *agent.State) e
 
 // Revoke removes both CDI specs.
 func (s *Simulator) Revoke(_ context.Context, h *host.Host) error {
-	zap.L().Debug("revoking simulator", zap.String("simulator", name))
+	zap.L().Info("revoking simulator", zap.String("simulator", name))
 	s.ready.Store(false)
 
 	return errors.Join(
--- a/internal/agent/cdi/cdi.go
+++ b/internal/agent/cdi/cdi.go
@@ -12,9 +12,10 @@ import (
 	"path/filepath"
 	"sync/atomic"
 
-	"github.com/NVIDIA/k8s-test-infra/internal/fsutil"
-
+	"go.uber.org/zap"
 	"sigs.k8s.io/yaml"
+
+	"github.com/NVIDIA/k8s-test-infra/internal/fsutil"
 
 	"github.com/NVIDIA/k8s-test-infra/internal/agent"
 	"github.com/NVIDIA/k8s-test-infra/internal/agent/host"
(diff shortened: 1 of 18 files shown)


internal/agent/cdi/cdi.go
```diff
@@ -62,6 +63,7 @@ func (s *Simulator) Discard(_ context.Context, _ *host.Host) error { return nil
 
 // Apply writes /run/cdi/nvidia.yaml and /run/cdi/nvml-mock-nri.yaml.
 func (s *Simulator) Apply(_ context.Context, h *host.Host, state *agent.State) error {
+	zap.L().Debug("applying simulator", zap.String("simulator", name))
```
giuliocalzo: @roma-glushko  can we have Info ? in this way we see the simulator log performing its own action

roma-glushko: Are we interested to see that though outside of local dev env? In the local env we can have the log level debug in order to see all logs.

internal/agent/pcibus/pcibus.go
```diff
@@ -61,12 +64,19 @@ func (s *Simulator) Stage(_ context.Context, h *host.Host, state *agent.State) e
 		return err
 	}
 
+	s.ready.Store(true)
```
ArangoGutierrez: Ready() on this type is documented as "the PCI surfaces and NFD feature file are published", and only Apply writes that file, so setting ready at the end of Stage makes /readyz report pcibus ready before nvml-mock.features exists. It also turns three existing assertions red: TestStage_RendersTopology (pcibus_test.go:54), TestStage_NopWhenNoTopology (:68) and TestReady_TrueAfterApply (:194) all require Ready()==false right after Stage, and golang / check is failing on them. Leaving readiness owned by Apply keeps the contract those tests encode.

internal/agent/pcibus/pcibus.go
```diff
@@ -61,12 +64,19 @@ func (s *Simulator) Stage(_ context.Context, h *host.Host, state *agent.State) e
...
+	zap.L().Debug("simulator staged", zap.String("simulator", name))
 	return nil
 }
 
 // Discard empties the rendered PCI sysfs tree and removes the staged shim.
 // pcibus owns both, so clearing absent or partially staged paths is safe.
 func (s *Simulator) Discard(_ context.Context, h *host.Host) error {
+	if !s.ready.Load() {
```
ArangoGutierrez: agent.Run tears down with revoke() then discard(), and pcibus.Revoke stores ready=false, so this guard makes Discard a no-op on every shutdown: the rendered PCI sysfs tree and the staged libpcisysfs.so survive on the host mount that outlives the pod. A probe running Stage, Apply, Revoke, Discard leaves 0000:07:00.0 under pcisysfs.PCIDevicesRelPath, and removing just these three lines clears it. ib and imex can carry the same guard only because their ready is owned by Stage rather than by Apply/Revoke. The docstring two lines up already says pcibus owns both paths so clearing partially staged ones is safe.

Greptile flagged this too.

ArangoGutierrez (changes requested):
The per-simulator stage and discard logs plus the skip reason coming out of inject.Adjust make it much easier to tell what the node agent actually did on a live node, and the edge-triggered fabricmanager enabled/disabled pair is a nice touch. Two changes to pcibus readiness rode along with the logging and need another pass, both inline.

- Two slips in the new log-level rule that contributors will be reading: "Debug level is or diagnosing" wants "is for diagnosing", and "Error level is for The current operation failed" reads like a paste seam. (AGENTS.md:47)

internal/agent/pcibus/pcibus.go
```diff
@@ -61,12 +64,19 @@ func (s *Simulator) Stage(_ context.Context, h *host.Host, state *agent.State) e
...
+		return nil
+	}
```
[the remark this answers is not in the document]
roma-glushko: Should be better now:

https://github.com/NVIDIA/k8s-test-infra/pull/787/commits/000dd0795e2131a36b002eb695e21c39afedda74

internal/agent/cdi/cdi.go
roma-glushko: Turned the lifecycle log severity into INFO: https://github.com/NVIDIA/k8s-test-infra/pull/787/commits/a5a772cba56d28b056b8f9f0c87eeec27021b5d9 ✅

internal/agent/pcibus/pcibus.go
roma-glushko: Should be fixed now: https://github.com/NVIDIA/k8s-test-infra/commit/000dd0795e2131a36b002eb695e21c39afedda74

internal/agent/pcibus/pcibus.go
roma-glushko: Should be already fixed ✅

giuliocalzo (approved):

ArangoGutierrez (approved):

PR #788, opened 2026-09-03
Moved Host to Simulator deps

roma-glushko:
## What This PR Does

- moving Host as a dependency to Simulator's constructors vs having it in the interfaces
- Use Host.*Path() methods everywhere

## Why

Trying to improve readability of the node agent.

## Checklist

---


ArangoGutierrez (changes requested):
Nice cleanup. Handing the Host to each simulator at construction instead of
threading it through every Stage/Apply/Discard/Revoke call is the right
direction, and it drops the agent package's dependency on the host package
entirely. I checked all seven simulators function by function and the move
itself is faithful: `h` becomes `s.host`, `filepath.Join(h.Root, rel)` becomes
`h.RootPath(rel)` (same value), no error handling changed, no stage reordered,
no dependency lost. Assertion counts are identical in all ten touched test
files, so nothing got loosened to make the move compile.

One thing snuck in that is not part of the move. gpudriver.go:66 adds
`s.ready.Store(true)` at the end of Stage. The base does not have it, and the
Ready() doc comment two lines above still says readiness means the published
`/run/nvidia/driver` symlink exists, which only Apply creates. That matters
because Ready() feeds /readyz, which the DaemonSet uses as its readinessProbe:
reconcile skips the whole Apply wave when any simulator's Stage fails
(agent.go:167), so if, say, pcibus.Stage errors, gpudriver would now report
ready indefinitely with no symlink on disk.

The suite already catches this. TestStage_WritesAllSurfaces asserts
`require.False(t, sim.Ready())` right after Stage, and on root Linux with the
NVML shim installed it fails on this branch and passes on the base commit. CI
does not see it because that test is behind skipUnlessRootLinux plus
skipUnlessNVMLLib and the runner is neither, so it skips.

Could you drop that line? If marking a staged driver ready is actually what you
want, it deserves its own commit with the Ready() comment and the assertion at
gpudriver_test.go:261 updated to match, plus coverage that runs outside the
root-only path.

Two small ones while you are in there:

- gpudriver.go:86 reads "Discard removes only the paths Stage wrote.  writes."
  Looks like a leftover from an in-place edit.
- pcibus/stage.go:84 uses `h.RootPath()` with no arguments where `h.Root` would
  read better. Same value, purely cosmetic.

Last thought: the t.Context() commit is unrelated to the Host move and is only
partly applied (fabricmanager and pcibus are converted, gpudriver is 5 of 21,
the rest untouched). No objection to the change, but it might be cleaner as its
own PR so this one stays a pure move.

giuliocalzo (dismissed):

ArangoGutierrez (approved):
Thanks for the quick turnaround. Handing each simulator its Host at construction drops internal/agent's dependency on the host package entirely, and the lifecycle signatures read much better without the parameter threaded through them.

The readiness store is out of gpudriver.Stage, so Ready() means the published /run/nvidia/driver symlink again, and the Discard comment reads cleanly now. Every path helper lands on the namespace it replaced, h.Etc to EtcPath in pcibus Apply and Revoke included, and all eight simulators take the same h from main.go. Assertion and test counts are unchanged across the thirteen touched test files, so nothing was loosened to make the move compile.

- TestRun_ReturnsImmediatelyWhenNotFull evaluates t.Context() inside the spawned goroutine, the only one of the ten goroutine sites in this diff that does not hoist ctx to a local first. It is harmless today since Run returns nil for ModeOff and ModeSysfs before it reads ctx, but hoisting it above the go statement keeps the T access on the test goroutine and matches the other sites. (internal/agent/ib/ib_test.go:97)

PR #789, opened 2026-09-03
chore: cleanup unused code + moved libcuda to shims

roma-glushko:
## What This PR Does

- Removed unused pkg/kubernetes
- Moved pkg/gpu/mockcuda shim to the shims directory.

pkg/diagnostics stays in place as it's used by NICO.

The CUDA shim is needed by DRA, so it seems like we should keep it as is for now.

## Checklist


ArangoGutierrez (changes requested):
FIX CI and I'll LGTM

deployments/nvml-mock/Dockerfile
```diff
@@ -49,7 +49,7 @@ RUN make build
 FROM debian:bookworm-slim
 ARG TARGETARCH
 COPY --from=builder /src/pkg/gpu/mocknvml/libnvidia-ml.so.* /usr/local/lib/
-COPY --from=builder /src/pkg/gpu/mockcuda/libcuda.so.* /usr/local/lib/
+COPY --from=builder /src/shims/libcuda/libcuda.so.* /usr/local/lib/
```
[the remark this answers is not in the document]
roma-glushko: check one more time. I kept libcuda as it's harder to remove it than I thought.

enhancements/meps/0003-node-agent/README.md
```diff
@@ -230,7 +230,6 @@ func (c *DeviceDriver) Stage(ctx context.Context, host host.Host, state *node.St
     g, gctx := errgroup.WithContext(ctx)
     g.Go(func() error { return c.materializeCharDevs(gctx, host, state.Devices) })
     g.Go(func() error { return c.stageNVMLShim(gctx, host, state.Software) })
-    g.Go(func() error { return c.stageCUDAShim(gctx, host, state.Software) })
     g.Go(func() error { return c.stageNvidiaSMI(gctx, host, state.Software) })
```
ArangoGutierrez, line 233: The Stage sketch lost its stageCUDAShim goroutine and is down to six, but gpudriver.go still launches it. Line 141 of this file says the simulator runs 7 things and lists the CUDA shim among them, the component table at 218 still lists it, and the coverage row this PR edits still reads "gpudriver stages the shim". Restoring the line keeps the sketch matching the code.

tests/e2e/go/scenario_gfd_test.go
```diff
@@ -78,23 +73,13 @@ func deployGPUFeatureDiscovery(ctx SpecContext, h *harness.Harness, node string)
 	Expect(h.Kube.Apply(ctx, assets.GFDManifest)).To(Succeed(), "apply GFD manifest")
 	Expect(h.Kube.DeletePodsByLabel(ctx, gfdNamespace, gfdSelector)).To(Succeed(), "restart GFD pods")
 	assertions.WaitDaemonSetReady(ctx, h.Kube, gfdNamespace, gfdName, config.ReadyTimeout(), config.PollInterval())
-	assertions.WaitNodeLabelsPresent(ctx, h.Kube, node, validatorGFDRequiredLabels(), config.ReadyTimeout(), config.PollInterval())
+	assertions.WaitNodeLabelsPresent(ctx, h.Kube, node, gfdRequiredLabels(), config.ReadyTimeout(), config.PollInterval())
```
ArangoGutierrez: WaitNodeLabelsPresent only checks the three keys exist, so this passes on labels a previous profile left behind and on whatever product or memory value GFD reports. assertions.WaitGFDLabels with ExpectedGFDLabels(p.GFDProductName(), p.MemoryMiB(), p.ExpectedGPUs()) compares against the profile and is already used that way in scenario_gpu_operator_test.go. Now that this is the scenario's only assertion, that seems worth the switch.

ArangoGutierrez (dismissed):
Thanks for this, retiring the vectorAdd validator spec drops a scenario that could never have gone green, and libcuda now lives with the other shims.

libcuda moves in beside libibmock and libmockfs under shims/, with the Dockerfile building and copying it from the new path, so the shipped .so is unchanged; the e2e scenario keeps the GFD label assertions and loses the vectorAdd job that could not pass. Dropping the unused pkg/kubernetes also demotes node-feature-discovery/api/nfd to an indirect require, which go.mod now reflects.

- This still routes pkg/gpu/mockcuda/** to area/mockcuda, a path the rename deletes, and no rule covers shims/, so area/mockcuda can no longer be applied to anything. Repointing it at shims/libcuda/** also needs the deep-equal literal in .github/actions/repo-automation/test/config.test.js updated, since that test asserts the parsed file. (.github/repo-automation/areas.yml:9) https://github.com/NVIDIA/k8s-test-infra/blob/ad91c00a813938777f97d84c6f288596f52c7865/.github/repo-automation/areas.yml#L7-L11
- The new example cites nvidia.com/gpu.present=true in the device-plugin job, but the only caller is scenario_nfd_test.go asserting feature.node.kubernetes.io/pci-10de.present, and the device plugin runs as a DaemonSet. Citing the NFD caller would match what the helper is actually used for. (tests/e2e/go/assertions/resources.go:102)

deployments/nvml-mock/Dockerfile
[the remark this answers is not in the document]
ArangoGutierrez: The reword is already in the file at this head: the Unreleased entry reads "The image still builds and ships libcuda.so and the agent still stages it, so a node is unchanged", and names the move from pkg/gpu/mockcuda to shims/libcuda. No entry says the library stopped being built or shipped.

enhancements/meps/0003-node-agent/README.md
roma-glushko: That's not a big deal as MEP0002 shows just an example not a final state of any simulator at any given point of time.

tests/e2e/go/scenario_gfd_test.go
roma-glushko: Should be switched in https://github.com/NVIDIA/k8s-test-infra/pull/789/commits/3d2700f22aa5598cc1783d6755a2aec73212aaf8

giuliocalzo (approved):

PR #794, opened 2026-09-04
Make GPU reset remediation work against mock GPUs

### LINKED ISSUE #759: [Feature]: Consider making nvidia-smi available in the GPU operator's standard install location
### Problem statement

A common remediation that is performed in response to a fault is GPU reset. The interesting part here is that the GPU reset itself works when run inside the DCGM pod like shown below:

```
# nvidia-smi --gpu-reset
GPU 00000000:01:00.0 was successfully reset.
GPU 00000000:02:00.0 was successfully reset.
GPU 00000000:03:00.0 was successfully reset.
GPU 00000000:04:00.0 was successfully reset.
GPU 00000000:05:00.0 was successfully reset.
GPU 00000000:06:00.0 was successfully reset.
GPU 00000000:07:00.0 was successfully reset.
GPU 00000000:08:00.0 was successfully reset.
All done.
```

but when run via an automation script https://github.com/NVIDIA/NVSentinel/blob/main/gpu-reset/gpu_reset.sh backed by a controller https://github.com/NVIDIA/NVSentinel/blob/main/janitor/pkg/controller/gpureset_controller.go, it fails with 

```
(2026-09-01 08:00:06) INFO: Using DRIVER_ROOT=/run/nvidia/driver
(2026-09-01 08:00:06) INFO: Testing nvidia-smi invocation: chroot /run/nvidia/driver nvidia-smi --version
chroot: failed to run command 'nvidia-smi': No such file or directory
```

I suspect this has something to do with the paths where the binaries are present being different between the two systems

### Proposed solution

TBD

### Scope / non-goals

_No response_

### Audience

Operators / cluster admins

### Pre-submission checks

giuliocalzo:
## What This PR Does

Makes NVIDIA's own GPU-reset remediation path work against mock GPUs, and adds a demo phase that drives it end to end.

NVIDIA's `gpu_reset.sh` never calls `nvidia-smi` directly — every call goes through `chroot "$DRIVER_ROOT" nvidia-smi`, so `DRIVER_ROOT` has to name a whole filesystem: the binary, its shared libraries and a dynamic loader. On real hardware the default of `/run/nvidia/driver` is exactly that, because the driver container r-bind-mounts its entire container root there. The mock stages driver surfaces rather than a filesystem, so the reset failed before `nvidia-smi` ever started.

This PR sidesteps the `chroot` rather than trying to satisfy it. [NVIDIA/NVSentinel#1813](https://github.com/NVIDIA/NVSentinel/pull/1813) made the janitor's chroot target configurable, so the demo sets `resetJob.driverRoot: "/"`, which turns the `chroot` into a no-op and leaves the script running whichever `nvidia-smi` the container already has. The `gpu-reset` image has none of its own — it is a CUDA runtime image plus the script — so the demo gives the reset Job a `nvidia.com/gpu` request, and CDI answers it with the mock's `nvidia-smi`, its `libnvidia-ml.so.1`, and a writable bind of the mock's config directory, which is what lets the reset clear state instead of only exiting 0.

The request covers every GPU the node advertises rather than one. The mock derives a container's visible GPUs from the `/dev/nvidia*` nodes it was handed and filters only on a partial set, so a Job holding a subset could not reach the UUID the janitor tells it to reset. The count is read from the node's allocatable so every `GPU_PROFILE` works, not just the 8-GPU ones.

Two changes:

- **A reset that cannot resolve its overrides now fails.** `ResetDevice` used to return success when no config path resolved, so a reset could report success having cleared nothing. A GPU reset is the remediation an operator reaches for first, and a silent no-op there is worse than an error.
- **Demo Phase 3** injects an uncorrectable ECC error, waits for fault-remediation to write a `GPUReset`, and runs NVIDIA's real `gpu-reset` image as the janitor's Job. It runs by default; `GPU_RESET=false` skips it.

## Why

Fixes [#759](https://github.com/NVIDIA/k8s-test-infra/issues/759).

The demo deliberately uses NVIDIA's unmodified reset image, because the janitor judges success purely from the Job's exit status — an image whose entrypoint merely exited 0 would drive the CR green without touching a GPU. Two things guard against reading that green state too generously: the script's first act is an `nvidia-smi --version` preflight under `set -e`, and the phase asserts the injected fault is actually gone rather than merely reported gone.

An earlier revision of this PR took the other route, staging an ELF interpreter and glibc closure into the driver root to make it chroot-able. That is now dropped in favour of the upstream configuration knob: roughly 330 lines of loader-closure staging and its e2e assertions are gone, and the no-chroot reset path is already covered by the existing `assertGpuResetFromInjectedContainer` spec, which asserts a reset from a CDI-injected container clears the override rather than merely exiting 0.

## Status

Draft, because `resetJob.driverRoot` landed after NVSentinel v1.23.0 was cut, so `NVSENTINEL_VERSION` pins a release that does not exist yet. The README says so and points at `GPU_RESET=false` for running the thermal phases against an older version.

## Verification

Unit tests and `golangci-lint` pass; the demo's Phase 3 has not been run end to end, because no released chart accepts `driverRoot` yet. What was verified statically:

- `helm show values oci://ghcr.io/nvidia/nvsentinel --version v1.23.0` has no `driverRoot`, while NVSentinel `main` does — hence the pin and the draft.
- The janitor renders `resetJob.resources` through `toYaml` and parses arbitrary resource names, and `helm template` confirms `--set-string` with an escaped key lands as `nvidia.com/gpu: "8"`.
- The mock's CDI spec injects `/usr/bin/nvidia-smi`, `/usr/lib64/libnvidia-ml.so.1` and a writable config bind, which is what a `driverRoot: "/"` reset needs.
- `detectVisibleDevicesAt` filters only when some device nodes are present and others absent, so requesting the node's full set leaves every GPU resettable.

The remaining assumption is how CDI device nodes and the Job's host `/dev` bind mount layer in practice. Requesting the full GPU set makes both outcomes equivalent, but it wants one live run once a release carries `driverRoot`.

## Checklist

- [x] Linter passes (`make lint-fix`) — `golangci-lint` clean; `govulncheck` reports pre-existing stdlib advisories that reproduce unchanged on `main`


tests/e2e/go/assertions/nvidiasmi/exec.go
```diff
@@ -268,3 +268,70 @@ func SnapshotFromPod(ctx context.Context, k *kube.Client, pod kube.PodRef) (Snap
...
+// "was successfully reset" over an untouched override — so the injected state
+// is checked before and after instead.
+func GpuResetThroughChroot(ctx context.Context, k *kube.Client, pod kube.PodRef, p profile.Profile) {
+	ginkgo.GinkgoHelper()
+
+	// The same query is the reference reading and the chrooted one, so the
+	// comparison holds on every profile without a table of expected values.
+	// Both are taken before anything is injected: an override in flight during
+	// the pair would show up as a disagreement that is only a refresh boundary.
+	query := []string{"nvidia-smi", "--query-gpu=index,temperature.gpu,pci.bus_id", "--format=csv,noheader"}
```
ArangoGutierrez: Including temperature.gpu makes this byte-for-byte comparison timing-dependent. tests/e2e/ci/nvml-mock.values.yaml enables dynamicMetrics and overrides only seed and utilization, so the chart baseline temperature block (variance_c 3, ramp_c 15, ramp_period_sec 120) stays live in both processes: roughly 2.5s of drift between the two nvidia-smi runs moves the ramp term a degree, and any difference in NVML call order shifts the variance draw off the shared per-device PCG by up to 3. Either one surfaces as "did not resolve this node's config", which points at the wrong thing. Comparing index and pci.bus_id only keeps the discrimination, since the count check and the PCI domain already separate the compiled-in defaults.

internal/agent/gpudriver/chroot_test.go
```diff
@@ -0,0 +1,174 @@
...
+	t.Parallel()
+	skipUnlessNvidiaSMI(t)
+
+	h := testHost(t)
+	state := testState(t)
+
+	require.NoError(t, stageChrootRuntime(context.Background(), h, state))
+
+	injected := filepath.Join(h.Root, "driver/usr/lib64")
+	entries, err := os.ReadDir(injected)
```
ArangoGutierrez, line 170: driver/usr/lib64 cannot exist in this fixture: testHost is a bare t.TempDir(), stageChrootRuntime is the only stager that runs, and chrootDest rejects every destination outside driver/lib and driver/lib64. ReadDir returns ENOENT, the early return fires, and neither require.False ever executes, so the CDI-safety property the test names has nothing holding it. Running stageNVMLShim first, or asserting positively that the glibc copy landed under driver/lib64, would give it something that can fail.

ArangoGutierrez (commented):
Thanks for this, staging the loader and the DT_NEEDED closure is what finally lets an unmodified NVSentinel reset Job reach a mock GPU, and reading PT_INTERP out of the binary rather than deriving it from GOARCH keeps that honest across arm64 and amd64.

The staging path and the config fallback do the real work here; the new tests are thinner than the code they cover.

- The closure test builds its expected set by calling chrootRuntimeClosure itself, so the BFS transitivity is unpinned: dropping the enqueue of each resolved library would shrink the staged set and the expected set together and this stays green. Asserting one independently known member, the PT_INTERP loader plus a second-level dependency, would catch that. (internal/agent/gpudriver/chroot_test.go:134)

tests/e2e/go/assertions/nvidiasmi/exec.go
giuliocalzo: Confirmed and fixed in 354a0d11 — the query is now `index,pci.bus_id`.

I traced the baseline you point at: `_helpers.tpl` merges `temperature (base_c 55, variance_c 3, ramp_c 15, ramp_period_sec 120)` as the chart layer under `chart baseline < per-profile defaults < user overrides`, and the e2e values override only `seed` and `utilization`, so it does stay live. The h100 profile's `dynamic_metrics` block sets `utilization` alone, so nothing displaces it either. That also explains a reading I had taken on a live cluster and not thought about: 60-65 C where the profile's static idle temperature is 34 C.

Agreed that nothing is lost: the count check catches the fallback's device set and `pci.bus_id` catches its PCI domain, while temperature was the same library in both processes and never told the two apart.

internal/agent/gpudriver/chroot_test.go
giuliocalzo: Confirmed and fixed in 354a0d11. Nothing was holding the property, exactly as you describe: `testHost` is a bare temp dir, `stageChrootRuntime` is the only stager that runs, and `chrootDest` rejects every destination outside `driver/lib*`, so `ReadDir` hit ENOENT and the early return fired before either `require.False`.

It now asserts positively that a libc landed under `chrootStageRoots` before checking `driver/usr/lib64`, which is what separates "kept out of the injected directory" from "nothing was staged". The lookup descends instead of listing each root, since libraries land in a multiarch subdirectory, and it tolerates whichever of `driver/lib` and `driver/lib64` this architecture leaves unpopulated.

The test skips without an `nvidia-smi` ELF, so I verified the lookup separately rather than assume it: it finds a nested libc, tolerates the absent root, and returns empty for an unstaged root, so the new `require.NotEmpty` does fire.

roma-glushko:
> A real driver container r-bind-mounts its whole container root there

Can we do the same in Mokka? Sounds like a higher fidelity move to me that replicates better the real driver setup.

giuliocalzo:
> > A real driver container r-bind-mounts its whole container root there
> 
> Can we do the same in Mokka? Sounds like a higher fidelity move to me that replicates better the real driver setup.

I'm making as part of another PR, I prefer to limit the scope of this PR

giuliocalzo:
pending https://github.com/NVIDIA/NVSentinel/pull/1813

Issue #758, opened 2026-09-01
[Feature]: Would it be possible to simulate the syslog?


lalitadithya:
### Problem statement

A lot of the GPU errors are surfaced in the journal/syslog. It would be good to have a way to wire in a simulated journal/syslog

### Proposed solution

TBD

### Audience

Operators / cluster admins

### Pre-submission checks

- [x] I searched existing issues / Discussions and this is not a duplicate.
- [x] I understand triage cadence is weekly and Unprioritized SLA is 7 calendric days.

PR #796, opened 2026-09-04
feat(nvml-mock-ctl): announce injected Xids on the node kernel log

giuliocalzo:
## What This PR Does

An injected Xid now reaches the node's kernel log the way a driver's printk does, in addition to being delivered through the NVML event set:

```text
kernel: NVRM: Xid (PCI:0000:1a:00): 79
```

- **`internal/agent/kernellog`** is the new simulator that writes it. It watches the runtime override document and announces each Xid that appears on it, once per fault: a standing failure is not re-announced, a recovery retracts nothing (kernel logs never do), and re-injecting after a recovery announces again even with the same code.
- **`pkg/gpu/mockctl`** gains the emitter. One write per targeted GPU, since `/dev/kmsg` turns each write into exactly one kernel message. Never `O_CREATE`: on a node without a kernel log, a created regular file would swallow the records and report success.
- **`nvml-mock-ctl`** is unchanged: it writes the override document, as before. `--xid` still records the code; what it no longer does is write the kernel line.
- **Chart**: `nodeAgent.kernelLog.enabled` (default `true`) mounts `/dev/kmsg` and runs `node-agent` privileged. Set to `false` and the DaemonSet sets `MOCK_NVML_KMSG=""`, so the agent stays quiet.
- **nv-sentinel demo**: `run.sh` configures `journald` on one Kind worker (`Storage=persistent`, `ReadKMsg=yes`).
- **Local dev**: the `default` Tilt cluster mounts the same drop-in (`local/kind/journald-node-syslog.conf`) at creation, so a hand-run `fail --xid` is readable with `docker exec <node> dmesg`.

## Why

Addresses #758 (simulate the journal/syslog GPU errors surface in).

The agents this project exists to exercise learn about Xids from the kernel log, not from NVML: NVSentinel's syslog monitor reads the journal and admits only `_TRANSPORT=kernel` entries. An NVML-only injection is invisible to them, so the one fault most worth simulating could not be simulated end to end.

Two details are load-bearing and pinned by tests. The driver names the PCI device without its function and prints hex in lower case, so a real Xid reads `(PCI:0000:1a:00)` for BDF `0000:1A:00.0` — the same spelling the simulated `/sys/bus/pci` tree uses. And the `kernel: ` prefix keeps `NVRM: ` inside the journal's `MESSAGE` field: journald parses a leading `ident: ` out of every kmsg line, so without it the entry arrives as `SYSLOG_IDENTIFIER=NVRM` with the prefix stripped, and patterns anchored on `NVRM: Xid` stop matching.

### Why the agent owns it

The kernel log is host state, like the character devices and the PCI sysfs tree the other simulators write, so it belongs to the agent rather than to a client whose job is to mutate a config file. Owning it there also decouples the announcement from one way of injecting a fault — the allocation watcher, a hand-edited override or a future control plane raise the same line — and it lets the announcement come from the device state the agent staged: already capped by `gpu.count`, and carrying the address each device serves through NVML, neither of which the CLI could do without re-deriving the node's device list from the profile ConfigMap.

The mock NVML engine was the other candidate, and the only one with real timing fidelity, but it runs unprivileged inside every NVML consumer: N processes would each emit, a node with no consumer would emit nothing, and no engine-to-agent channel exists. What no owner can do is announce at the moment of the trip: a deferred failure (`--after-calls N`) trips inside whichever consumer makes the Nth guarded call, an in-memory counter that never reaches disk. The kernel line therefore follows the injection, not the fault, and leads NVML when the failure is deferred.

### Why privileged, and why on by default

Mounting `/dev/kmsg` is not enough. Verified on a live cluster: a pod running as root, with the node's `crw-rw-rw-` device bind-mounted, gets `Operation not permitted`; the same pod with `privileged: true` writes fine. The device cgroup is what blocks it, and no capability lifts that — the capability set is identical in both cases (`drop: [ALL]`, `add: [MKNOD]`). That is a real escalation for a pod that otherwise drops everything, but an Xid that reaches only NVML is invisible to the agents this project exists to exercise, so it is on by default rather than hidden behind a flag nobody would find. `nodeAgent.kernelLog.enabled=false` restores the unprivileged pod for clusters that will not take one (a restricted PodSecurity namespace rejects it) and for nodes with no `/dev/kmsg`, where the `CharDevice` hostPath fails the pod rather than inventing a device. Announcing is always best-effort either way: a node that cannot be written to keeps the NVML injection and logs the skip.

### Why the demo needs a journald drop-in

Stock Kind nodes satisfy neither half of what the syslog monitor needs: journald ships `ReadKMsg=no`, dropping kernel messages, and keeps a volatile journal in `/run/log/journal` where a `/var/log` hostPath cannot see it. Applying the drop-in after node creation takes two restarts, verified on a throwaway cluster: journald declares `CanReload=no`, and it does not create `/var/log/journal` itself — `systemd-journal-flush` is what creates the directory and moves the journal there. The demo enables ingestion on a single worker because Kind nodes share the host's kernel ring buffer, so every node would otherwise report the same fault.

## Verification

- Unit tests for the emitter, the agent simulator (announce once, re-announce after recovery, every targeted device, ignore unrelated override edits, no kernel log) and the CLI; 172 Helm unit tests including the toggle cases; `golangci-lint` clean; `go test -race ./...` green.
- End to end on a fresh Kind cluster with an image built from this branch, `gpu.count=2`: the boot-time drop-in gave the node a persistent journal with `ReadKMsg=yes`; `nvml-mock-ctl fail --gpu 0 --xid 79` produced `MESSAGE: 'NVRM: Xid (PCI:0000:0a:00): 79'`, `_TRANSPORT: kernel`, `SYSLOG_IDENTIFIER: kernel` — the shape NVSentinel matches.
- The behaviours that only exist once the agent owns it, checked on that cluster: a standing failure is not re-announced (line count unchanged over repeated polls), a recovery announces nothing, re-injecting the same code announces again, and `--gpu all` reached only the two GPUs `gpu.count` serves — one line for the device the shared bucket applies to, none for the device whose own failure block already shadows it, exactly as the engine merges them.
- With `nodeAgent.kernelLog.enabled=false`, the upgraded DaemonSet came back unprivileged, an injection wrote nothing to the kernel log, and the agent logged no warning.

## Checklist

- [x] Linter passes (`golangci-lint`: 0 issues). Note: `make lint-fix`'s `govulncheck` step fails on pre-existing Go 1.26.5 standard-library advisories (fixed in 1.26.6), unrelated to this diff.

### DIFF
--- a/cmd/nvml-mock-ctl/faults.go
+++ b/cmd/nvml-mock-ctl/faults.go
@@ -57,6 +57,8 @@ func failCommand() *cli.Command {
 				Name:  "after-calls",
 				Usage: "trip only after N guarded calls have succeeded",
 			},
+			// The node agent watches the override this writes and announces
+			// the Xid on the node's kernel log, as a driver's printk does.
 			&cli.Uint64Flag{
 				Name:  "xid",
 				Usage: "Xid code to surface alongside the failure",
--- a/deployments/nvml-mock/helm/nvml-mock/templates/daemonset.yaml
+++ b/deployments/nvml-mock/helm/nvml-mock/templates/daemonset.yaml
@@ -112,6 +112,14 @@ spec:
             # the CLI can never write to.
             - name: MOCK_NVML_OVERRIDES
               value: /var/lib/nvml-mock/driver/config/overrides.yaml
+            # Where the agent announces an injected Xid, as a driver's printk
+            # would. Unlike the env above this one is the agent's own: it
+            # watches the override document and writes the kernel line, so an
+            # Xid raised by any writer reaches the node's log. Empty says this
+            # pod cannot reach the kernel log -- it is only mounted, and only
+            # writable, when nodeAgent.kernelLog is enabled.
+            - name: MOCK_NVML_KMSG
+              value: {{ ternary "/dev/kmsg" "" .Values.nodeAgent.kernelLog.enabled | quote }}
             {{- if .Values.topology.enabled }}
             - name: MOCK_TOPOLOGY_CONFIG
               value: /etc/nvml-mock/topology/topology.yaml
@@ -149,10 +157,19 @@ spec:
               protocol: TCP
             {{- end }}
           securityContext:
+            {{- if .Values.nodeAgent.kernelLog.enabled }}
+            # Writing /dev/kmsg needs no capability, but it does need past the
+            # device cgroup, which only privileged lifts -- mounting the device
+            # into an unprivileged container still fails with EPERM. Stated
+            # alone: a capability set is inert under privileged, so dropping ALL
+            # here would advertise a bound the container does not have.
+            privileged: true
+            {{- else }}
             allowPrivilegeEscalation: false
             capabilities:
               drop: [ALL]
               add: [MKNOD]
+            {{- end }}
           volumeMounts:
             # Where the simulators write. Every path they stage is rooted here.
             - name: mokka-root
@@ -195,6 +212,12 @@ spec:
             - name: node-sysfs
               mountPath: /host/sys
               readOnly: true
+            {{- if .Values.nodeAgent.kernelLog.enabled }}
+            # The node's kernel ring buffer, for the Xid announcement. Write
+            # only in practice: nothing here reads it back.
+            - name: kernel-log
+              mountPath: /dev/kmsg
+            {{- end }}
           livenessProbe:
             httpGet:
               path: /healthz
@@ -276,6 +299,14 @@ spec:
           hostPath:
             path: /sys
             type: Directory
+        {{- if .Values.nodeAgent.kernelLog.enabled }}
+        # CharDevice, so a node that exposes no kernel ring buffer fails the
+        # pod instead of silently swallowing announcements into a new file.
+        - name: kernel-log
+          hostPath:
+            path: /dev/kmsg
+            type: CharDevice
+        {{- end }}
         {{- if .Values.topology.enabled }}
         - name: cluster-topology
           configMap:
--- a/deployments/nvml-mock/helm/nvml-mock/values.schema.json
+++ b/deployments/nvml-mock/helm/nvml-mock/values.schema.json
@@ -165,6 +165,14 @@
           "pattern": "^[0-9]+(ns|us|ms|s|m|h)$",
           "description": "Go duration string passed to --shutdown-timeout."
         },
+        "kernelLog": {
+          "type": "object",
+          "additionalProperties": false,
+          "description": "Announce injected Xids on the node's kernel log. Closed to unknown keys because a typo here would leave the privileged container it grants in place while reading as if it had been turned off.",
+          "properties": {
+            "enabled": { "type": "boolean" }
+          }
+        },
         "resources": { "type": "object" }
       }
     },
--- a/deployments/nvml-mock/helm/nvml-mock/values.yaml
+++ b/deployments/nvml-mock/helm/nvml-mock/values.yaml
@@ -83,6 +83,24 @@ nodeAgent:
   logging:
     level: info    # debug | info | warn | error
     format: json   # json | plain
+  # Let the agent announce an injected Xid on the node's kernel log, the way a
+  # real driver's printk does, so agents that read kernel logs (and journald
+  # consumers such as NVSentinel's syslog monitor) see it instead of only NVML
+  # clients. It watches the runtime override document, so an Xid raised through
+  # `nvml-mock-ctl fail --xid` or any other writer is announced alike.
+  #
+  # Off by default because turning it on makes the node-agent container
+  # PRIVILEGED: /dev/kmsg sits behind the device cgroup, which rejects the write
+  # from an unprivileged pod even with the device mounted, so there is no lesser
+  # grant that works. A default of true would widen an existing install's
+  # privileges on `helm upgrade` alone, which no value here should do.
+  #
+  # Turn it on to evaluate an agent that reads kernel logs; a cluster whose
+  # PodSecurity admits no privileged pod, or a node with no /dev/kmsg, cannot
+  # (the CharDevice hostPath fails the pod rather than inventing one). Left off,
+  # the announcement is skipped and the NVML injection is unaffected.
+  kernelLog:
+    enabled: false
   # Go duration string: 5s, 1m30s, etc. Teardown is a handful of parallel file
   # removals, and it has to finish inside terminationGracePeriodSeconds.
   shutdownTimeout: 5s
--- a/docs/demo/nv-sentinel/run.sh
+++ b/docs/demo/nv-sentinel/run.sh
@@ -104,6 +104,32 @@ done < <(kind get nodes --name "${CLUSTER_NAME}" | grep -v control-plane | sort)
 [[ "${#WORKERS[@]}" -ge 1 ]] || fail "no worker nodes found"
 info "GPU workers: ${WORKERS[*]}"
 
+# --- Node "syslog" for the syslog health monitor ------------------------------
+# The monitor opens the journal DIRECTORY under a /var/log hostPath and admits
+# only kernel-transport entries, neither of which a stock Kind node offers: it
+# keeps a volatile journal in /run/log/journal and ships ReadKMsg=no, so the
+# Xid the node agent puts on /dev/kmsg is dropped. Both restarts
+# are needed — journald cannot reload, and it does not create /var/log/journal
+# itself; journal-flush is what creates it and moves the journal there.
+#
+# Exactly one node, because the kernel ring buffer is not namespaced: every Kind
+# node on this host shares the machine's, so an Xid injected on any node reaches
+# all of them. Enabling ingestion everywhere would have each node's monitor
+# report the same fault and NVSentinel quarantine the whole cluster.
+SYSLOG_NODE="${WORKERS[0]}"
+info "Enabling persistent kernel-log journaling on ${SYSLOG_NODE}"
+docker exec "${SYSLOG_NODE}" bash -c '
+set -e
+mkdir -p /etc/systemd/journald.conf.d
+cat > /etc/systemd/journald.conf.d/99-node-syslog.conf <<EOF
+[Journal]
+Storage=persistent
+ReadKMsg=yes
+EOF
+systemctl restart systemd-journald
+systemctl restart systemd-journal-flush
+'
+
 # --- Label GPU workers + install nvidia-container-toolkit / CDI ---------------
 for node in "${WORKERS[@]}"; do
   info "Labeling ${node} with ${GPU_NODE_LABEL}"
@@ -149,6 +175,10 @@ info "Loading image into Kind"
 kind load docker-image "${IMAGE_NAME}" --name "${CLUSTER_NAME}"
 
 # --- Install nvml-mock (pinned to the GPU workers) ----------------------------
+# nodeAgent.kernelLog is what this demo turns on: the chart ships it off, since
+# it makes the node-agent container privileged, and without it an injected Xid
+# reaches NVML alone -- invisible to NVSentinel's syslog monitor, which is the
+# half of NVSentinel this demo exists to show.
 info "Installing nvml-mock (profile=${GPU_PROFILE}) on the GPU workers"
 helm upgrade --install nvml-mock "${REPO_ROOT}/${CHART_PATH}" \
   --kube-context "${KUBE_CONTEXT}" \
@@ -158,6 +188,7 @@ helm upgrade --install nvml-mock "${REPO_ROOT}/${CHART_PATH}" \
   --set "gpu.profile=${GPU_PROFILE}" \
   --set gpu.dynamicMetrics.enabled=true \
   --set-string "nodeSelector.nvml-mock-gpu=true" \
+  --set nodeAgent.kernelLog.enabled=true \
   --wait --timeout 180s
 
 # --- Install the NVIDIA GPU Operator with standalone DCGM ---------------------
--- a/internal/agent/kernellog/kernellog.go
+++ b/internal/agent/kernellog/kernellog.go
@@ -1,0 +1,289 @@
+// Copyright 2026 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+// Package kernellog simulates the kernel side of an Xid: the printk a real
+// driver emits when a GPU faults.
+//
+// It belongs to the agent rather than to nvml-mock-ctl because the kernel log
+// is host state, like the character devices and the PCI sysfs tree the other
+// simulators write, and because the fault is not the CLI's to announce — an
+// injection reaches the mock through the runtime override document, whoever
+// writes it. Watching that document instead of the CLI's own invocation is what
+// lets the allocation watcher, a hand-edited file or a future control plane
+// raise an Xid that the agents watching this node can see.
+package kernellog
+
+import (
+	"context"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"go.uber.org/zap"
+
+	"github.com/NVIDIA/k8s-test-infra/internal/agent"
+	"github.com/NVIDIA/k8s-test-infra/internal/agent/host"
+	"github.com/NVIDIA/k8s-test-infra/pkg/gpu/mockctl"
+	"github.com/NVIDIA/k8s-test-infra/pkg/gpu/mocknvml/engine"
+)
+
+const name = "kernellog"
+
+// defaultInterval matches the TTL of the engine's own override cache, so the
+// kernel line and the NVML event follow an injection within the same second.
+const defaultInterval = time.Second
+
+var (
+	_ agent.Simulator = (*Simulator)(nil)
+	_ agent.Daemon    = (*Simulator)(nil)
+)
+
+// Simulator announces newly injected Xids on the node's kernel log.
+type Simulator struct {
+	path     string // kernel log to write; empty announces nowhere
+	interval time.Duration
+	staged   atomic.Bool
+
+	mu        sync.Mutex
+	overrides string         // runtime override document to watch
+	devices   []device       // in device order, so a fault on several prints in it
+	announced map[int]uint64 // device index -> Xid already on the kernel log
+	warned    map[int]uint64 // device index -> Xid we already failed to write
+	lastErr   string         // deduplicates a persistent read failure
+}
+
+// device is what this simulator needs of a GPU: which bucket of the override
+// document speaks for it, and what to call it in a kernel line.
+type device struct {
+	index int
+	busID string
+}
+
+// Options configures the simulator.
+type Options struct {
+	// Path is the kernel log to write, /dev/kmsg on a node that grants one.
+	// Empty disables the announcement, which is how a deployment that cannot
+	// reach the kernel log says so: the NVML side of an injection stands on its
+	// own, so this is a quiet degradation rather than an error.
+	Path string
+	// Interval is how often the override document is re-read. Zero picks a
+	// default aligned with the engine's own polling.
+	Interval time.Duration
+}
+
+// New returns a kernel log Simulator.
+func New(opts Options) *Simulator {
+	interval := opts.Interval
+	if interval == 0 {
+		interval = defaultInterval
+	}
+
+	return &Simulator{
+		path:      opts.Path,
+		interval:  interval,
+		announced: map[int]uint64{},
+		warned:    map[int]uint64{},
+	}
+}
+
+// Name returns the simulator's stable identifier.
+func (s *Simulator) Name() string { return name }
+
+// Ready reports whether the simulator knows which devices this node serves.
+func (s *Simulator) Ready() bool { return s.staged.Load() }
+
+// Stage records where the injections arrive and which address each device
+// answers to, so a kernel line names the device the way NVML does. It writes
+// nothing: the artifact this simulator produces is an event, not a file.
+func (s *Simulator) Stage(_ context.Context, h *host.Host, state *agent.State) error {
+	zap.L().Info("staging simulator", zap.String("simulator", name))
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	// The document nvml-mock-ctl writes and the engine reads. Derived rather
+	// than configured because both ends already agree on it: the DaemonSet
+	// points MOCK_NVML_OVERRIDES here, and it is the CLI's default besides.
+	s.overrides = h.RootPath("driver/config/overrides.yaml")
+
+	previous := s.devices
+	s.devices = make([]device, 0, len(state.Devices))
+	for _, d := range state.Devices {
+		// A profile is free to leave addressing to the mock, and such a device
+		// still serves the address it inherits from the base mock. Naming it
+		// any other way would leave the kernel line pointing at no GPU.
+		busID := strings.ToLower(d.PCIBusID)
+		if busID == "" {
+			busID = engine.BaseDevicePCIBusID(d.Index)
+		}
+
+		// The boundary the rendered PCI tree already applies to bus_id, for a
+		// sharper reason: the address goes into the node's kernel ring buffer,
+		// which is unnamespaced and read by health agents, so an address
+		// carrying a newline would let whoever can edit the profile put lines
+		// of their own where a remediator reads them. Logged verbatim, which is
+		// safe — zap escapes it — and is what names the offending value.
+		if !agent.ValidBDF(busID) {
+			zap.L().Warn("device has no usable PCI address; its Xids cannot be announced",
+				zap.String("simulator", name), zap.Int("device", d.Index),
+				zap.String("pci", d.PCIBusID))
+			continue
+		}
+		s.devices = append(s.devices, device{index: d.Index, busID: busID})
+	}
+
+	// A profile edit can renumber devices, and an index that now names another
+	// GPU must not inherit what the old one announced, or the new device's
+	// first Xid is read as a fault already reported.
+	s.forgetRenumbered(previous)
+
+	s.staged.Store(true)
+	// Says how many devices can be named, which is what a silent kernel log
+	// comes down to when a profile's addresses do not survive the check above.
+	zap.L().Info("simulator staged", zap.String("simulator", name),
+		zap.Int("devices", len(s.devices)), zap.String("kernelLog", s.path))
+
+	return nil
+}
+
+// forgetRenumbered drops what was announced for any device whose address is not
+// the one it had at the previous staging, including a device that is gone.
+func (s *Simulator) forgetRenumbered(previous []device) {
+	current := make(map[int]string, len(s.devices))
+	for _, d := range s.devices {
+		current[d.index] = d.busID
+	}
+
+	for _, was := range previous {
+		if current[was.index] != was.busID {
+			delete(s.announced, was.index)
+			delete(s.warned, was.index)
+		}
+	}
+}
+
+// Run announces injections until the agent shuts down.
+func (s *Simulator) Run(ctx context.Context) error {
+	if s.path == "" {
+		<-ctx.Done()
+		return nil
+	}
+
+	ticker := time.NewTicker(s.interval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+			s.Poll(ctx)
+		}
+	}
+}
+
+// Reload is a no-op: Stage already delivers every revision.
+func (s *Simulator) Reload(_ context.Context, _ *agent.State) error { return nil }
+
+// Discard stops the announcements. It retracts nothing — a kernel log never
+// takes an Xid back — and forgets what it announced, so the next agent may
+// report a fault that outlives this one.
+func (s *Simulator) Discard(_ context.Context, _ *host.Host) error {
+	zap.L().Info("discarding simulator", zap.String("simulator", name))
+	s.staged.Store(false)
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	s.announced = map[int]uint64{}
+	s.warned = map[int]uint64{}
+
+	return nil
+}
+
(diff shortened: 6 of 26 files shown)


cmd/nvml-mock-ctl/main.go
```diff
@@ -375,9 +381,64 @@ func mutate(cmd, configOverridePath, gpu, mode, links string, afterCalls int, xi
 		return 1
 	}
 	fprintf(stdout, "ok: %s applied to %s\n", cmd, gpuLabel(gpu))
+
+	// Announced only once the override is on disk, so a failed injection never
+	// leaves a phantom Xid in the kernel log. Healthy clears an injection and
+	// kernel logs never retract an Xid, so that mode announces nothing.
+	if cmd == "fail" && xid > 0 && mode != engine.FailureModeHealthy {
+		announceXid(kmsgPath, xid, target, cfg, stdout, stderr)
+	}
```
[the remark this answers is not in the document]
giuliocalzo: Right about the ordering, but neither suggested remedy is reachable from here. The deferred trip is counted inside each NVML consumer's own failure injector — it never reaches disk — so no process outside that consumer can observe or wait for it, and those consumers are the unprivileged ones that deliberately have no `/dev/kmsg` access, which is why the announcement lives in the CLI in the first place. Rejecting the combination would take `--after-calls` away from the documented demo flow, where `N=1` trips on the next guarded call and the drift is sub-second.

So 1b150ee4 makes the lead explicit instead of hiding it: `fail --xid` reports on stderr that `--after-calls N` holds the NVML failure back while the kernel log already has the Xid, and the docs say to omit `--after-calls` when a consumer must not see the Xid before NVML reports the device failed.

cmd/nvml-mock-ctl/main.go
```diff
@@ -375,9 +381,64 @@ func mutate(cmd, configOverridePath, gpu, mode, links string, afterCalls int, xi
...
+		}
+	}
+
+	busIDs := make([]string, 0, len(indexes))
+
+	for _, index := range indexes {
+		if busID := cfg.GetDevicePCIBusID(index); busID != "" {
+			busIDs = append(busIDs, busID)
+		}
+	}
```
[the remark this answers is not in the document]
giuliocalzo: Good catch, this was a real bug — fixed in 1b150ee4.

A profile that omits `pci.bus_id` still serves an address: `createDevicesFromYAML` passes the empty override through and the device keeps the base mock's `PciBusID` (`0000:00:00.0` … `0000:07:00.0`), so NVML reported a BDF while the announcement emitted nothing and warned. `Config.EffectiveDevicePCIBusID` now resolves what the device actually serves — profile first, base mock otherwise — and `targetBusIDs` uses it. The fallback is pinned against `dgxa100.New()` in a test rather than a formula, so a change to the base mock fails the test instead of silently naming the wrong device, and a CLI test covers a config with `system.num_devices` and no device list.

deployments/nvml-mock/helm/nvml-mock/templates/daemonset.yaml
```diff
@@ -149,7 +156,14 @@ spec:
               protocol: TCP
             {{- end }}
           securityContext:
+            {{- if .Values.nodeAgent.kernelLog.enabled }}
+            # Writing /dev/kmsg needs no capability, but it does need past the
+            # device cgroup, which only privileged lifts -- mounting the device
+            # into an unprivileged container still fails with EPERM.
+            privileged: true
```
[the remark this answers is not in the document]
giuliocalzo: Acknowledged and intentional. The privilege is what makes the feature work at all: `/dev/kmsg` sits behind the container device cgroup, so an unprivileged container is refused with `operation not permitted` even as root with the node's world-writable device bind-mounted, and no capability short of privileged lifts that. Isolating the writer in its own container would move the privilege rather than remove it, and would split `nvml-mock-ctl` away from the config volumes it mutates.

Keeping it on by default is deliberate for a tool whose whole purpose is simulating GPU faults on Kind and other test clusters, where an agent that reads kernel logs sees nothing without it. `nodeAgent.kernelLog.enabled=false` restores the restricted context for a cluster that will not take a privileged pod or a node with no `/dev/kmsg`; the DaemonSet then sets `MOCK_NVML_KMSG=""` and `--xid` skips the announcement silently, leaving the NVML injection unaffected. Both the value's comment and `docs/nvml-mock-ctl.md` spell out the trade-off.

deployments/nvml-mock/helm/nvml-mock/values.yaml
```diff
@@ -83,6 +83,20 @@ nodeAgent:
...
+  #
+  # This costs a PRIVILEGED node-agent container: /dev/kmsg sits behind the
+  # device cgroup, which rejects the write from an unprivileged pod even with
+  # the device mounted, so there is no lesser grant that works. Disable it on a
+  # cluster that will not take that (a restricted PodSecurity namespace rejects
+  # the pod outright) or on a node with no /dev/kmsg, since the CharDevice
+  # hostPath fails the pod rather than inventing one. Disabled, the
+  # announcement is skipped and the NVML injection is unaffected.
+  kernelLog:
+    enabled: true
```
ArangoGutierrez: Defaulting kernelLog.enabled to true turns the node agent privileged on the next `helm upgrade` of any existing install, and this chart is published to ghcr for third parties. Every other footprint-widening toggle here (allocationWatcher, nri, topology, controlPlane) is opt-in, so defaulting this one to false and enabling it from the demo and local values would match that, and would stop the `drop: [ALL]` block below from advertising a limit a privileged container does not have.

Greptile flagged this too.

pkg/gpu/mockctl/kmsg.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	return true, nil
+}
+
+// driverPCI reduces an NVML bus ID to the domain:bus:device form the driver
+// prints, dropping the function suffix.
+func driverPCI(busID string) string {
+	if before, _, ok := strings.Cut(busID, "."); ok {
+		return before
```
ArangoGutierrez: driverPCI passes the profile's spelling through, so the shipped profiles emit `PCI:0000:1A:00` while a real NVRM printk and this repo's own BDF surfaces (`strings.ToLower` in internal/pcisysfs/render.go and internal/agent/state.go) are lowercase. Now that EffectiveDevicePCIBusID falls back to the base mock's lowercase address, one binary emits both spellings; lowercasing here lets an agent match the Xid line against the simulated /sys/bus/pci entry.

ArangoGutierrez (commented):
Thanks for this, an Xid that only reached NVML was invisible to the agents this repo exists to exercise, and the kernel-log path closes that.

The driver-shaped record, the dropped PCI function, and the `kernel: ` prefix that keeps `NVRM: ` inside the journal MESSAGE field are all pinned by tests, which is what will keep this matching real agents as the code moves.

- An absent kernel log returns `(false, nil)` and the switch below has no branch for it, so `--kmsg /wrong/path` announces nothing and reports nothing, while docs/nvml-mock-ctl.md says a missing log reports the skip on stderr. This warning also runs before EmitXid checks for an empty path, so a pod with `MOCK_NVML_KMSG=""` and a config that failed to load prints `not announced on ` with a blank path, which is the warn-when-disabled case kernelLogDefault exists to avoid. (cmd/nvml-mock-ctl/main.go:402)
- nodeAgent.kernelLog has no entry in values.schema.json while logging, shutdownTimeout and resources all do, and nodeAgent is additionalProperties: true, so `kernelLog: {enable: false}` validates and still renders privileged: true. Declaring a boolean enabled with additionalProperties: false, the way nodeAgent.logging is declared, catches the typo at template time. (deployments/nvml-mock/helm/nvml-mock/values.yaml:98)

cmd/nvml-mock-ctl/main.go
```diff
@@ -375,9 +381,74 @@ func mutate(cmd, configOverridePath, gpu, mode, links string, afterCalls int, xi
...
+		indexes = make([]int, 0, cfg.NumDevices)
+		for i := range cfg.NumDevices {
+			indexes = append(indexes, i)
+		}
+	}
+
+	busIDs := make([]string, 0, len(indexes))
+
+	for _, index := range indexes {
+		if busID := cfg.EffectiveDevicePCIBusID(index); busID != "" {
```
ArangoGutierrez: targetBusIDs walks cfg.NumDevices from the profile's device list, but GPU_COUNT is what caps the devices the engine actually serves and the CLI never reads it, so `gpu.count: 2` against an 8-device profile makes `fail --gpu all --xid 79` write 8 kernel lines, 6 of them naming GPUs no consumer can see. Bounding the enumeration by GPU_COUNT, already in the node agent's environment, keeps the kernel log to devices that exist.

docs/demo/nv-sentinel/run.sh
```diff
@@ -98,6 +98,28 @@ mapfile -t WORKERS < <(kind get nodes --name "${CLUSTER_NAME}" | grep -v control
...
 info "GPU workers: ${WORKERS[*]}"
 
+# --- Node "syslog" for the syslog health monitor ------------------------------
+# The monitor opens the journal DIRECTORY under a /var/log hostPath and admits
+# only kernel-transport entries, neither of which a stock Kind node offers: it
+# keeps a volatile journal in /run/log/journal and ships ReadKMsg=no, so the
+# Xid `nvml-mock-ctl fail --xid` writes to /dev/kmsg is dropped. Both restarts
+# are needed — journald cannot reload, and it does not create /var/log/journal
+# itself; journal-flush is what creates it and moves the journal there.
+for node in $(kind get nodes --name "${CLUSTER_NAME}"); do
```
ArangoGutierrez: The kernel ring buffer is not namespaced, so all Kind node containers on one host share it: `dmesg` in two nodes of a running cluster returns byte-identical output. Setting ReadKMsg=yes on every node then copies an Xid injected on one worker into every node's journal, so a per-node syslog consumer reports the fault on healthy nodes too; either scope the drop-in to the node under test or say so where the docs treat the kernel line as node-local.

pkg/gpu/mockctl/kmsg.go
giuliocalzo: Agreed, fixed in 66f5b8b0: `driverPCI` now lower-cases as well as dropping the function, so the record reads `kernel: NVRM: Xid (PCI:0000:1a:00): 79` whichever way the address was resolved, and it matches both a real NVRM printk and the `/sys/bus/pci` entries the mock stages.

cmd/nvml-mock-ctl/main.go
giuliocalzo: Confirmed and fixed in 66f5b8b0. The node-agent container's `MOCK_NVML_CONFIG` is `/etc/nvml-mock/config.yaml`, the pristine profile ConfigMap — the staged config the agent stamps `system.num_devices` into lives at a path only the other container uses, so `loadConfig` never saw the runtime count. A test reproduces it: `gpu.count`-equivalent `GPU_COUNT=1` against the two-GPU fixture emitted two kernel lines.

`loadConfig` now applies GPU_COUNT last, mirroring `resolveDeviceCount` in `internal/agent/source/file.go`, and caps rather than extends — a count above the profile's is the agent's warning to give. This bounds the `--gpu N` check as well as the `all` enumeration, so `--gpu 1` at `GPU_COUNT=1` is now the usage error it should always have been.

docs/demo/nv-sentinel/run.sh
giuliocalzo: You are right, and it is worse than cosmetic — I reproduced it on a three-node cluster:

```
$ docker exec mokka-worker sh -c 'echo "kernel: MOKKA-KMSG-SCOPE-PROBE" > /dev/kmsg'
mokka-worker: 1
mokka-worker2: 1
mokka-control-plane: 1
```

With ingestion on every node, one GPU's Xid would have NVSentinel quarantine the whole cluster. 66f5b8b0 scopes it: the demo configures only `${WORKERS[0]}`, and `local/kind/default.kind.yaml` mounts the drop-in on `worker-0` alone instead of all three nodes. `docs/nvml-mock-ctl.md` now says a kernel line is not node-local on Kind and why, so nobody re-enables it fleet-wide, and `local/README.md` and the drop-in's own header say the same where a reader meets them.

deployments/nvml-mock/helm/nvml-mock/values.yaml
giuliocalzo: I would rather keep the default at `true`, but the upgrade path you describe is the strongest argument against it, so tell me if you still want it flipped and I will.

The reasoning for on-by-default: simulating a GPU fault that a kernel-log agent can see is the feature, and a chart that ships it off means every NVSentinel-style evaluation starts by discovering the Xid went nowhere. Unlike `allocationWatcher`, `nri` or `controlPlane`, this toggle does not add a component — it is the only way `--xid` can behave like a driver, since `/dev/kmsg` sits behind the device cgroup and nothing short of privileged gets past it.

On the `drop: [ALL]` block: it is inert under `privileged: true`, so I can drop it from that branch to stop advertising a limit that does not hold.

The other three comments are fixed in 66f5b8b0, including the kernel ring buffer one — it turned out to be a real cluster-wide false-positive.

cmd/nvml-mock-ctl/faults.go
roma-glushko, line 1: @giuliocalzo why do you think that CTL CLI which is a pure client should responsible for kernel log modification? It's either nvml mock or node agent responsibility. I don't think CTL is supposed to do anything with that. Curious to hear your argument.

giuliocalzo: You are right, and I have moved it: the announcement now belongs to the node agent (`internal/agent/kernellog`), and the CLI is back to writing the override document and nothing else.

The agent watches that same document and announces each Xid that appears on it, which turned out to be the better boundary for two reasons beyond layering. It decouples the kernel line from one way of injecting a fault — the allocation watcher, a hand-edited override or a future control plane raise it just the same, where before only `nvml-mock-ctl fail --xid` did. And it fixes two of @ArangoGutierrez's findings by construction rather than by patch: the agent announces from the device state it staged, which is already capped by `gpu.count`, and names each device by the address that state carries — the same one the simulated `/sys/bus/pci` tree uses. The CLI could only infer both from the profile ConfigMap.

The engine was the other candidate, and it is the only place with real timing fidelity, but it runs unprivileged inside every NVML consumer: N processes would each emit, a node with no consumer would emit nothing, and there is no engine-to-agent channel today. Worth noting what neither owner can do, since it bears on the `--after-calls` thread: the trip of a deferred failure lives in whichever consumer makes the Nth guarded call, as an in-memory atomic that never reaches disk, so the announcement follows the injection rather than the fault, whoever makes it.

One thing the move gives up: `fail --xid` used to report `ok: Xid 79 announced on /dev/kmsg` and to say why when it could not. That feedback is now in the agent's log (`kubectl -n mokka logs "$POD" | grep kernellog`), which the troubleshooting section points at.

The chart is unchanged in shape — same `/dev/kmsg` mount, same privileged container, same `MOCK_NVML_KMSG` — because the CLI ran inside the node-agent container to begin with, so this costs no privilege that was not already on the table.

cmd/nvml-mock-ctl/main.go
giuliocalzo: Update on this one: the cap is gone again in 16b2f725, because the thing it protected moved.

The announcement is the node agent's now (`internal/agent/kernellog`, see the thread below), and the agent announces from the device list it staged, which the source already caps by `GPU_COUNT` — so `--gpu all` can no longer name a GPU the engine does not serve, without the CLI re-deriving the node's device count from the profile ConfigMap.

What remains is narrower than what you reported and predates this branch: `--gpu <index>` still bounds against the profile's device list, so an index the profile lists but `gpu.count` excludes passes the check and writes an override no consumer reads. Harmless for the kernel log, wrong for the bounds check. Happy to fix it separately if you want a ticket — it did not seem right to keep a `GPU_COUNT` read in the CLI purely for that once the announcement no longer depended on it.

internal/agent/kernellog/kernellog.go
```diff
@@ -0,0 +1,225 @@
...
+		code := doc.FailureXid(d.index)
+		if code == 0 {
+			// Recovered, or never faulted. Forgetting it here is what makes a
+			// repeat of the same Xid a new event rather than a duplicate.
+			delete(s.announced, d.index)
+			continue
+		}
+		if s.announced[d.index] == code {
+			continue
+		}
```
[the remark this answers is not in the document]
giuliocalzo: Real, and now pinned as behaviour in 1427d606 — but it cannot be closed by a write generation, because the two states are the same document.

`fail --mode healthy` removes the device's failure block, so a clear followed by a re-arm of the same code leaves `overrides.yaml` byte-for-byte what it was when the Xid was announced. A generation counter only moves the problem: at document level it also moves when an unrelated writer touches the file (the allocation watcher rewrites it every few seconds), so a standing fault would re-announce on every one of those; per-device it would have to live inside the device bucket, and `MergeDeviceConfig` decodes that bucket with `DisallowUnknownFields`, so a marker there fails the merge and takes the whole override with it. Making it work means adding a field to the engine's device-config schema whose only purpose is this dedupe, visible in every `nvml-mock-ctl status`. Nor does the poll rate matter, or inotify: two writes between two reads collapse whatever wakes the reader, and this repo watches config by polling with a content hash rather than fsnotify by design (`internal/agent/source/file.go`).

What the window actually costs is a duplicate of a line the log already holds. Recovery prints nothing — a real driver does not announce a GPU getting better — so a consumer scanning the log sees "Xid 79 on this device, never retracted" whether we printed once or twice. The standing state stays correct: the NVML side reports the fault, and the next *distinct* code, or a clear that a read does observe, is announced normally. Only something counting events notices, and one that does is already reading a log where the same fault can arrive twice for one injection (Kind nodes share a kernel ring buffer).

So: `TestSimulator_CollapsesAClearAndRearmBetweenPolls` pins it, the dedupe carries the reasoning, and `docs/tools/node-agent.md` tells the reader before they meet it.

giuliocalzo:
@ArangoGutierrez picking up the two points from your review body, both of which were still open:

**`nodeAgent.kernelLog` missing from `values.schema.json`** — fixed in 1427d606, exactly as you described. `nodeAgent` is `additionalProperties: true`, so `kernelLog: {enable: false}` validated and rendered `privileged: true`: the value reads as off while the chart grants the privilege, which is the worst direction for a typo to fail in. The block is now declared with `additionalProperties: false` like `nodeAgent.logging`, and a chart test asserts the misspelling fails the install rather than silently widening access.

**The `--kmsg` stderr warning that reported nothing on a wrong path** — gone with the code it was in. Following @roma-glushko's point, the announcement moved out of the CLI into the node agent (`internal/agent/kernellog`), so there is no `--kmsg` flag and no CLI-side warning left to disagree with the docs; the agent logs each announcement, and warns once per distinct failure instead of once per tick. That move also resolved your `GPU_COUNT` finding by construction — the agent announces from the device list it staged, which the source already caps — see the thread on that comment.

Still open on your side is the default: `kernelLog.enabled` is `true` in this branch, with the upgrade-path argument you raised unanswered except by preference. Say the word and I will flip it to `false` and enable it from `local/nvml-mock.values.yaml` and the demo.

deployments/nvml-mock/helm/nvml-mock/values.yaml
giuliocalzo: Flipped in ce1a81a7 — your upgrade-path argument is the one that settles it. A value that widens an existing install's privileges when nothing but the chart version changed is not something a default should do, least of all in a chart third parties pull from ghcr.

`nodeAgent.kernelLog.enabled` is `false`, and it is turned on where the privilege is understood and the feature is the point: `local/nvml-mock.values.yaml`, whose Kind clusters already configure journald to ingest kernel messages, and the nv-sentinel demo, whose syslog monitor is half of what that demo exists to show. So the paths that need it keep working, and a plain `helm install`/`upgrade` no longer grants anything it did not before.

You were right about the capability block too: `drop: [ALL]` is inert under `privileged: true`, so the privileged branch now states the privilege alone rather than advertising a bound the container does not have. The unprivileged branch keeps `drop: [ALL]` / `add: [MKNOD]` unchanged, which is now the default rendering.

Tests follow the flip: the default case asserts unprivileged with no kernel-log mount, the enabled case asserts the privilege and that no capability set is claimed with it, and the three default-values snapshots are regenerated. While in the schema I closed the hole from your review body — `kernelLog` is declared with `additionalProperties: false`, so `enable: false` fails the install instead of reading as off while rendering privileged.

ArangoGutierrez (changes requested):
Nice piece of work, and the part I care most about is right: it is off by default, the value
that turns it on is the one that grants the privilege, the schema is closed so a typo cannot
quietly render a privileged pod, and a helm upgrade of an existing install cannot flip it. I
checked the off switch by running it rather than reading it, since a string flag that ignored
an empty env var would have written the node kernel log with the feature nominally off:
urfave/cli v3 does apply an empty env value to a string flag, so `MOCK_NVML_KMSG=""` really
does disable it. The `CharDevice` hostPath and the deliberate absence of `O_CREATE` close the
same hole from both ends, and not claiming a capability bound under privileged is a detail
most charts get wrong. The Kind shared-ring-buffer note in the kind config, the journald
drop-in and the docs is exactly the kind of thing that saves the next person a bad afternoon.

Two things I would like fixed before this goes in.

First, `FailureXid` shadows where the engine merges. `DeviceConfigOverride` deep-merges the
`all` bucket and the per-device bucket, so a per-device `failure` block does not replace the
shared one field by field. That means this sequence:

    nvml-mock-ctl fail --gpu all --mode lost --xid 79
    nvml-mock-ctl fail --gpu 0   --mode lost

leaves device 0 raising Xid 79 through NVML while `FailureXid(0)` returns 0, so every device
on the node gets a kernel line except the one the operator singled out. I reproduced it
against the engine's own merge. The comment and the unit test both assert the shadowing model,
so all three need to move together: deep-merge the two buckets before reading
`failure.xid.code`, and add the two-call case above.

Second, the PCI address goes into `/dev/kmsg` without passing the boundary this repo already
has for it. `validBDF` in `internal/agent/state.go` exists precisely because `bus_id` is
hand-authored, and `activeBDFs` applies it, but `kernellog.Stage` only checks for empty. With
a bus_id like `"0000:1a:00\nkernel: NVRM: Xid (PCI:0000:ff:00): 48"` a raw newline and
arbitrary text reach the write, so somebody with edit on the profile ConfigMap and no node
access can put content into every node's kernel ring buffer, which is the thing health
remediators read. To be fair about the size of it, `driverPCI` lower-cases, so the injected
text cannot reproduce the case-sensitive `NVRM: Xid` your own pattern test pins, and a forged
Xid for another GPU is blocked today by that accident rather than by a check. The sink is
node-wide and unnamespaced, which is what makes it worth a real check: export `validBDF` and
call it in `Stage` next to the empty-address case, dropping the device with the warning that
is already there.

Smaller things, take or leave:

- A failed `/dev/kmsg` write still records the code as announced, so a transient error drops
  that Xid until the fault is cleared and re-injected. Moving the dedupe onto the warning
  rather than the announcement would let the next tick retry.
- `Stage` rebuilds the device list on every state change but keeps `announced`, so a profile
  edit that renumbers devices can leave stale entries that suppress a later line. `Discard`
  already clears the map.
- `kmsg_test.go` carries the long Apache header while its three siblings use the SPDX form.
- The values comment explains that a PodSecurity-enforcing cluster cannot use this, but
  nothing tells the reader the namespace wants
  `pod-security.kubernetes.io/enforce: privileged`. One line in the docs would do it.

The tests hold up. I deleted the dedupe guard, the recovery forget, the lower-casing and the
`kernel: ` prefix in turn, and each one took tests red, so these are real guards and not
decoration. Pinning the clear-and-rearm collapse as a test instead of a comment is the right
way to record a known limit. Helm suite is 173 passing with 13 snapshots matching, so the
snapshot additions are regenerated rather than hand-edited.

deployments/nvml-mock/helm/nvml-mock/values.schema.json
```diff
@@ -165,6 +165,14 @@
           "pattern": "^[0-9]+(ns|us|ms|s|m|h)$",
           "description": "Go duration string passed to --shutdown-timeout."
         },
+        "kernelLog": {
+          "type": "object",
+          "additionalProperties": false,
+          "description": "Announce injected Xids on the node's kernel log. Closed to unknown keys because a typo here would leave the privileged container it grants in place while reading as if it had been turned off.",
+          "properties": {
+            "enabled": { "type": "boolean" }
+          }
```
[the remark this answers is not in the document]
giuliocalzo: Reproduced, and it is the chart's behaviour rather than this block's — every value block does it, including four that predate this branch:

```
--set nodeAgent.kernelLog=null   nil pointer at daemonset.yaml:122
--set topology=null              nil pointer at topology-configmap.yaml:9
--set allocationWatcher=null     nil pointer at daemonset.yaml:231
--set nri=null                   nil pointer at nri-daemonset.yaml:1
--set nodeAgent.logging=null     nil pointer at daemonset.yaml:53
```

That is the direct-lookup idiom the chart uses throughout, and nulling a block Helm then deletes is the one input it does not survive. Requiring the block in the schema does convert it into `at '/nodeAgent': missing property 'kernelLog'`, and I checked that a values file which never mentions the block still renders, since the chart defaults coalesce before validation — but applying that to `kernelLog` alone would leave the newest toggle the only guarded one, and applying it to all five is a chart-wide change that does not belong in a PR about the kernel log. Leaving both for a separate pass.

Worth separating from the schema gap you flagged one comment earlier, which I did fix: `kernelLog: {enable: false}` validated and rendered `privileged: true` — an install that silently grants more than was asked for, so the typo had to fail. This one fails in the opposite direction. `helm template` stops before anything is installed and the message names the exact path, so nothing renders wrong and no privilege is granted by accident; the operator sees a broken template on the spot and fixes the value. A loud failure on an input nobody writes by accident is not worth diverging from the pattern the rest of the chart follows.

giuliocalzo:
@ArangoGutierrez both blockers fixed in aa841041, along with all four of the smaller ones.

**`FailureXid` shadowed where the engine merges.** Confirmed and fixed. Your two-call sequence reproduced immediately — `FailureXid(0)` returned 0 while the engine's merge returned 79 — so the singled-out device was the one device on the node that got no kernel line. Rather than restate the merge rule correctly beside the engine, `FailureXid` now asks the engine: it builds an `engine.ConfigOverrideDoc` from the two buckets and reads `failure.xid.code` out of `DeviceConfigOverride(index)`, so the document can only ever answer with what the running mock raises. `bucketXid` and its shadowing contract are gone, the comment describes the deep merge, and the new test holds `FailureXid` against `MergeDeviceConfig` on the same document instead of against a second copy of the assumption — the way you reproduced it. A simulator-level test covers the symptom too: with `--gpu all --xid 79` then `--gpu 0 --mode lost`, both devices get a line.

**The address reached `/dev/kmsg` unchecked.** Fixed as you suggested: `validBDF` is exported and `Stage` calls it, which also subsumes the empty-address case, since an empty string is not an address either. One warning, now carrying the offending value so the log names it. Your forged `bus_id` is the test fixture verbatim, asserting the forged device announces nothing while its neighbour still does. You were right that the case-sensitivity was doing the work by accident, and that is not a thing to leave load-bearing when the sink is node-wide and unnamespaced; the exported doc comment says why the check exists at this boundary.

The four smaller ones, all taken:

- A failed write no longer records the code as announced, so the next tick retries; the dedupe moved onto the warning, keyed by device and code, so a standing failure still warns once rather than every second. Tested by pointing the path at a directory, injecting, then repairing it.
- `Stage` now drops what was announced for any index whose address changed or vanished, so a renumbered device announces its own fault. `Discard` still clears everything.
- `kmsg_test.go` carries the SPDX header like its siblings.
- `docs/nvml-mock-ctl.md` names the label: `pod-security.kubernetes.io/enforce=privileged`, and says that under `baseline` or `restricted` the DaemonSet is rejected outright rather than degraded.

On your mutation check — I ran the same discipline over the new guards, since a test that passes either way is worse than none. Deleting the `ValidBDF` call, the retry, and the renumber-forget each took their test red; the merge fix was red before it was written. Full suites: `go test -race` green across `internal/agent/...`, `pkg/gpu/mockctl` and `cmd/...`, `golangci-lint` clean, helm still 173 with 13 snapshots.

pkg/gpu/mockctl/config_override.go
```diff
@@ -124,6 +124,41 @@ func (d *Doc) Fail(t Target, mode string, afterCalls int, xidCode uint64) error
...
+
+	xid, ok := failure["xid"].(map[string]any)
+	if !ok {
+		return 0
+	}
+
+	// A document that has been through YAML carries the code as float64, one
+	// built in memory as an int or uint64.
+	switch code := xid["code"].(type) {
+	case float64:
```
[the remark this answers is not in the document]
giuliocalzo: Good catch, and it was real. Fixed in c464a5c8.

Reproduced first: with `code: 79.5` in the override document, `FailureXid` returned 79 while `MergeDeviceConfig` on the same bucket errored — so the agent would have put `NVRM: Xid (PCI:...): 79` on the node's kernel log for a device the mock left healthy. `code: -5` was worse: it wrapped to 2^64-5.

`FailureXid` now asks the engine for the answer outright rather than resolving the buckets through the engine and then reading the code itself:

```go
merged, err := engine.MergeDeviceConfig(&engine.DeviceConfig{}, overrides.DeviceConfigOverride(index))
if err != nil || merged.Failure == nil || merged.Failure.Xid == nil {
    return 0
}

return merged.Failure.Xid.Code
```

A code the mock cannot decode takes the override down there exactly as it does in the running mock, so it is announced nowhere. This also removes the `float64`/`int`/`uint64` type switch, whose `//nolint:gosec` justification ("a negative code cannot reach here: Fail takes a uint64") was the bug in one line: `Fail` is not the only writer of that document, which is the whole reason the agent watches it.

`TestDoc_FailureXidRefusesWhatTheEngineRefuses` pins it over a negative, fractional, non-numeric and past-uint64 code, asserting the engine refuses each one before asserting we announce nothing — otherwise the test would pass on a day the engine started accepting them.

ArangoGutierrez (approved):
/lgtm

PR #798, opened 2026-09-05
feat(node-agent): simulate NVIDIA kernel-module presence

### LINKED ISSUE #797: [Feature]: Simulate NVIDIA kernel-module presence in the node agent
## What

The GPU-driver surface table in MEP-0003 has one gap at [`enhancements/meps/0003-node-agent/README.md:280`](https://github.com/NVIDIA/k8s-test-infra/blob/874ab0319732981bc79cbc9b01c22c8fd00656e6/enhancements/meps/0003-node-agent/README.md#L280): kernel module presence is not simulated, so an unmodified consumer that checks module state sees nothing.

The GPU Operator validator shows the cost. Its `isNvidiaModuleLoaded()` function reads `/sys/module/nvidia/refcnt` with `os.Stat` (NVIDIA/gpu-operator, `cmd/nvidia-validator/main.go:922-933`). The file does not exist, so the validator continues to a `modprobe` command that cannot work in kind. The gpu-operator scenario avoids that step today with `DISABLE_DEV_CHAR_SYMLINK_CREATION=true`.

## What to render

`gpudriver` writes these paths in the overlay root. Each relative path is the kernel path, and each file keeps the kernel format:

- `proc/modules`
- `sys/module/<name>/{refcnt,coresize,initstate}`
- `sys/module/<name>/holders/<holder>`, as a symlink
- `sys/module/nvidia/version`

## Host modules

The kernel puts neither path in a namespace, so the agent reads the host set from its own container and needs no extra mount. A host module must keep its own entry, which prevents a second entry on a node where the real `nvidia` module is loaded ([#591](https://github.com/NVIDIA/k8s-test-infra/issues/591)). Three constraints follow:

- A mount replaces the whole `/sys/module` tree in a served container, so the mirror must carry every host module. A tree with only the NVIDIA modules makes `lsmod` print `-2` in the Size and Used-by columns for each host module. For the same reason the mount belongs on the CDI spec and not on the NRI adjustment, which reaches every container the values do not exclude: two GPU Operator operands bind `/sys/module/firmware_class/parameters/path`, and a read-only tree without that path stops them.
- Both surfaces must derive host presence from one signal, the `/proc/modules` text. Otherwise `/proc/modules` can advertise a module whose sysfs entry cannot answer, which is the state the validator fails on. A built-in module has a sysfs directory but no `/proc/modules` line, so the two sources do not map one to one.
- The render must prune, because the host set shrinks when a module unloads.

## Why one mechanism is not enough

runc permits a short allowlist inside `/proc` and `/proc/modules` is not on it (opencontainers/runc, `libcontainer/rootfs_linux.go:879-890`):

```
"/proc/modules" cannot be mounted because it is inside /proc
```

So `/proc/modules` can only be redirected, through the preload shim.

A preload shim alone is not sufficient either, because the validator is a Go program and Go issues `openat` directly. That is the same limit [#673](https://github.com/NVIDIA/k8s-test-infra/issues/673) reported for the PCI tree, which [#742](https://github.com/NVIDIA/k8s-test-infra/pull/742) fixed with a CDI mount at the kernel path. So `/sys/module` can only be mounted. A CDI container also receives no `containerEdits.env` today ([#747](https://github.com/NVIDIA/k8s-test-infra/issues/747)), so an environment variable can steer only an NRI container.

A CDI-served container therefore also needs a generated `lsmod` that reads the mounted tree, because the shim never applies to it. That script belongs at `/usr/local/bin/lsmod`, which a normal PATH puts first. Not `/usr/bin/lsmod`: with the `kmod` package installed that path is a symlink, so the mount replaces `/usr/bin/kmod` and `modprobe`, `insmod` and `rmmod` all run the script instead.

## Scope

This work touches the files [#742](https://github.com/NVIDIA/k8s-test-infra/pull/742) changed, so it follows the same patterns: one relative-path constant per surface, and a gate that matches the render condition. The startup gate of the GPU Operator driver container is out of scope, because Mokka sets `driver.enabled: false` and replaces that container.

Two known limits. A Go program that reads `/proc/modules` directly still sees the host file, and runc makes that permanent; a search of GPU Operator, DCGM and Mokka finds no consumer of that shape. The mirror is also a copy at one point in time, refreshed by a `State` reconcile, so a module load or unload on the node does not update it ([#689](https://github.com/NVIDIA/k8s-test-infra/issues/689) records the same property for `/sys/devices`).

faganihajizada:
Closes #797.

`gpudriver` writes `/proc/modules` and the module directories under `/sys/module`, and it mirrors the module tree of the node alongside them. `lsmod` then lists `nvidia` and `nvidia_uvm`, and `/sys/module/nvidia/refcnt` exists, which is the file the GPU Operator validator reads in `isNvidiaModuleLoaded()`.

Both surfaces derive host presence from one signal, the `/proc/modules` text, so they cannot disagree about a module. A module with a host line keeps its mirrored directory. A module without one gets the simulated attributes, because `/proc/modules` advertises it.

Three channels deliver the surface, because no single channel reaches both consumer kinds. `libpcisysfs.so` redirects both paths for libc callers. Both CDI specs bind-mount `sys/module` at its kernel path for Go callers, which issue `openat` and never enter the shim. runc refuses mounts inside `/proc`, so CDI-served containers also get a generated `lsmod` that reads the mounted tree.

The second commit corrects the MEP row. It listed "DCGM startup checks" as a consumer of module state, and DCGM does not read module state.


tests/e2e/go/assertions/kmod.go
```diff
@@ -0,0 +1,122 @@
...
+	return sizes
+}
+
+// KernelModules checks the simulated module surface.
+func KernelModules(ctx context.Context, k *kube.Client, pod kube.PodRef) {
+	ginkgo.GinkgoHelper()
+
+	ginkgo.By("checking whether the node loads a real nvidia module")
+	res, err := k.ExecSh(ctx, pod,
+		`if MOCK_PCI_ROOT= test -d /sys/module/nvidia; then echo host; else echo none; fi`)
```
ArangoGutierrez, line 42: `test` is a dash builtin, so the `MOCK_PCI_ROOT=` prefix never reaches the environ that libpcisysfs reads, and the stat still resolves into the mock tree, where `kmod.Render` always creates `sys/module/nvidia`. That makes `hostDriver` true on every node, so the four `if !hostDriver` blocks below it (the nvidia and nvidia_uvm lsmod rows, `refcnt` equal to 1, the version agreement, and the `nvidia_uvm,` holder in /proc/modules) never run. The `env -u LD_PRELOAD -u MOCK_PCI_ROOT` form used in scenario_nri_test.go execs, so it takes effect; `MOCK_PCI_ROOT= sh -c 'test -d ...'` works too.

ArangoGutierrez (commented):
Thanks for this, the module surface is the piece that finally lets lsmod and the validator's refcnt read work on a node with no real driver.

The unit coverage under internal/kmod is thorough, and the mirror-degradation, holder-reconcile and proc-line-gap cases each have a test that would go red on a regression. The e2e side has a problem, noted inline.

docs/helm-chart.md
```diff
@@ -1490,6 +1503,16 @@ while the validator is in `CrashLoopBackOff`, check `CLEANUP_ALL`: it makes the
...
 exits before recreating anything. Nothing sets it by default.
 
+**Driver validator fails after the kernel module check**: keep
+`DISABLE_DEV_CHAR_SYMLINK_CREATION=true` in the validator environment. The module
+check itself passes, because the mock serves `/sys/module/nvidia/refcnt`. The
+validator then calls `CreateLinks()`, which fails with `failed to create devices
+info: no NVIDIA devices found`. That call reads a hardcoded `/proc/devices` with
+no override. runc refuses to bind-mount that path, the reader is Go so the preload
+shim cannot see it, and the name it looks for appears only when a real kernel
+module registers a character-device major. Removal of the flag needs an upstream
```
roma-glushko, line 1513: > Removal of the flag needs an upstream change, not work in this repository.

This paragraph should be rephrased/removed not to be based on the conversation context.

docs/helm-chart.md
```diff
@@ -1490,6 +1503,16 @@ while the validator is in `CrashLoopBackOff`, check `CLEANUP_ALL`: it makes the
 validator `RemoveAll` its output dir, which fails `EBUSY` on the bind mount and
 exits before recreating anything. Nothing sets it by default.
 
+**Driver validator fails after the kernel module check**: keep
```
roma-glushko, line 1506: Does this information have any long-term value? It sounds to be based on the conversation context. It's fine to have a summary of the problem that pushed us to do the rendering (somewhere near the code it's related to).

internal/agent/cdi/spec.go
```diff
@@ -196,10 +198,35 @@ func pciSysfsMounts(state *agent.State) []cdiMount {
...
+// /usr/bin it would replace the kmod symlink and take modprobe with it.
+func kernelModuleMounts() []cdiMount {
+	return []cdiMount{
+		{
+			HostPath:      overlayHostRoot + "/" + kmod.SysModuleRelPath,
+			ContainerPath: "/" + kmod.SysModuleRelPath,
+			Options:       []string{"ro", "nosuid", "nodev", "bind"},
+		},
+		{
+			HostPath:      overlayHostRoot + "/" + kmod.LsmodRelPath,
```
roma-glushko, line 218: Paths should be joined with https://pkg.go.dev/path/filepath#Join in order to avoid issues with concatenation.

internal/agent/gpudriver/gpudriver.go
```diff
@@ -56,14 +57,16 @@ func (s *Simulator) Stage(ctx context.Context, h *host.Host, state *agent.State)
 	g.Go(func() error { return writeProcFS(gctx, h, state) })
 	g.Go(func() error { return writeEngineConfig(gctx, h, state) })
 	g.Go(func() error { return writeMachineType(gctx, h, state) })
+	g.Go(func() error { return writeKernelModules(gctx, h, state, kernelProcModules) })
```
roma-glushko, line 60: Should kernelProcModules() be just called in writeKernelModules() without passing it via interface?

internal/kmod/modules.go
```diff
@@ -0,0 +1,132 @@
...
+)
+
+// ProcModulesRelPath and SysModuleRelPath are overlay paths.
+const (
+	ProcModulesRelPath = "proc/modules"
+	SysModuleRelPath   = "sys/module"
+)
+
+const (
+	nvidiaCoreSize    = 62312448
```
roma-glushko, line 25: nvidiaCoreSize and nvidiaUVMCoreSize are confusing names. I thought it implies the number of SM cores or something like that which would be a device-specific characteristic.

`nvidiaModuleFileSizeBytes` and `nvidiaUVFileSizeBytes` may be better names.

internal/kmod/lsmod.go
```diff
@@ -0,0 +1,33 @@
...
+package kmod
+
+// LsmodRelPath is the staged script path.
+const LsmodRelPath = "driver/usr/local/bin/lsmod"
+
+// LsmodContainerPath is the script mount path.
+const LsmodContainerPath = "/usr/local/bin/lsmod"
+
+// LsmodScript reads the mounted module tree.
+const LsmodScript = `#!/bin/sh
```
roma-glushko, line 13: I love that this is compact and elegant. 
But would not it be better to have one more `shims/*` library for the sake of consistency and discoverability? We do something similar with the IMEX binary there.

deployments/nvml-mock/helm/nvml-mock/templates/daemonset.yaml
roma-glushko, line 1: Is working with proc directory requiring any elevated capabilities or privilege? We are least privilege by default, so may not have everything needed.

internal/agent/gpudriver/stage.go
```diff
@@ -203,6 +205,47 @@ func writeProcFS(ctx context.Context, h *host.Host, state *agent.State) error {
...
 }
 
+const kernelSysModuleDir = "module"
+
+const kernelProcModules = "/proc/modules"
+
+func writeKernelModules(ctx context.Context, h *host.Host, state *agent.State, procModulesPath string) error {
+	mods := kmod.Modules(state.Software.DriverVersion)
+
+	src, err := os.ReadFile(procModulesPath)
```
roma-glushko, line 215: We should probably have a function that parses procModules at this point, so we pass it further in a read to use state (and not double serializing in both Render() and ProcModules()).

internal/kmod/render.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+
+	return fsutil.PruneDir(filepath.Join(dir, "holders"), func(name string) bool { return declared[name] })
+}
+
+// mirrorTree copies src onto dst and reports whether it succeeded, along with
+// the source paths it could not read. A read failure skips the entry, which the
+// prune then removes, so an unreadable host module degrades the mirror rather
+// than failing the whole stage.
+func mirrorTree(src, dst string) (bool, []string, error) {
```
roma-glushko, line 206: mirrorTree()/mirrorEntry() are other candidates for the `internal/fsutils` promotion?

internal/kmod/lsmod.go
roma-glushko, line 13: We should probably clearly outline here in which case we need to have a custom lsmod because it's rather not self-obvious.

internal/kmod/render.go
```diff
@@ -0,0 +1,253 @@
...
+	// Modules are the simulated modules to serve alongside the host's own.
+	Modules []Module
+
+	// SourceRoot is the host /sys/module directory to mirror. An empty or
+	// absent SourceRoot serves the simulated modules alone.
+	SourceRoot string
+
+	// Output is the overlay root. The renderer writes under
+	// <Output>/sys/module and Render returns an error when it is empty.
+	Output string
```
roma-glushko, line 27: A bit confusing name. `OverlayRootDir` or `DestDir` may be a better choice.

shims/libpcisysfs/src/shim.c
roma-glushko, line 1: Does libpcisysfs still make sense after this addition? I think we should call it something like `libmockfs` now.

CHANGELOG.md
```diff
@@ -7,6 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+- node-agent: a surface that simulates the presence of the NVIDIA kernel modules,
```
roma-glushko, line 11: Too much details. We need a concise, user-facing explanation of the change and its effect/result.

internal/kmod/modules.go
```diff
@@ -0,0 +1,132 @@
+// Copyright 2026 NVIDIA CORPORATION
+// SPDX-License-Identifier: Apache-2.0
+
+// Package kmod stages mock kernel module state.
```
roma-glushko, line 4: Worth documenting the actual surface that we try to simulate here like module list output:

```
name       size      refcnt  deps          state  address              [taint]                                                                                                                                                          
  nvidia_uvm 3411968   0       -             Live   0x0000000000000000                                                                                                                                                                    
  nvidia     62312448  1       nvidia_uvm,   Live   0x0000000000000000
```

So it would make the rest of the simulation logic more straightforward for newcomers.

internal/kmod/modules.go
```diff
@@ -0,0 +1,132 @@
...
+
+const (
+	procModulesState   = "Live"
+	procModulesAddress = "0x0000000000000000"
+)
+
+// Module describes a kernel module.
+type Module struct {
+	Name     string
+	CoreSize int
```
roma-glushko, line 39: The same naming ambiguity.

roma-glushko (dismissed):

ArangoGutierrez:
I ran this to be sure I was not just pattern-matching on the shell, and the
result is worse than I first described: the suppression does not take effect at
all, so the branch is not merely fragile, it is inverted.

I built this PR's shim.c as-is and ran the exact form from line 42 under dash,
in a container where the real /sys/module/nvidia does not exist:

    positive control  test -d /sys/module/nvidia                  -> host
    negative control  (MOCK_PCI_ROOT empty at process start)      -> none
    line 42 form      MOCK_PCI_ROOT= test -d /sys/module/nvidia   -> host
    MOCK_PCI_ROOT= sh -c 'test -d /sys/module/nvidia'             -> none
    env -u MOCK_PCI_ROOT test -d /sys/module/nvidia               -> none

Because the real path does not exist, "host" can only mean the shim rewrote into
the mock tree. test is a dash builtin, so no exec happens, and the assignment
never reaches the environment the shim reads through getenv. The shim also caches
the root under pthread_once, so the first path call in that shell fixes the
answer for the whole process.

The consequence is that hostDriver is true on every node, and the four
if !hostDriver blocks below (the nvidia and nvidia_uvm lsmod rows, refcnt == 1,
version agreement, and the nvidia_uvm holder in /proc/modules) never run. That is
why 20 e2e jobs pass without exercising them.

Either of the exec'ing forms above fixes it. env -u MOCK_PCI_ROOT reads clearest,
and it matches what scenario_nri_test.go already does.

One request before this lands: after the fix, please confirm the assertion can
actually fail. Flipping one expected value and seeing the job go red is enough.
The rest of the PR looks good to me, the internal/kmod unit coverage in
particular.

tests/e2e/go/assertions/kmod.go
faganihajizada: You were right, and it was worse than fragile. The probe answered `host` on every node. I reproduced your table with this PR's `shim.c` under dash, then checked the live pod on kind: the old form printed `host`, and `env -u LD_PRELOAD -u MOCK_PCI_ROOT test -d /sys/module/nvidia` printed `none`. All four node reads now use a `nodeView` prefix, so none depends on whether the command is a builtin.

docs/helm-chart.md
faganihajizada: Not in the chart doc, so I removed the entry. The three GPU Operator values files in the repo already set the flag, and `local/gpu-operator/gpu-operator.values.yaml` explains it beside the setting. I corrected that note. Without the flag the validator finds `/sys/module/nvidia/refcnt`, skips `modprobe`, then fails with `failed to create devices info: no NVIDIA devices found`. With the flag set it never reads `refcnt`.

docs/helm-chart.md
faganihajizada: Removed with the rest of the entry. See the thread on line 1506.

internal/agent/cdi/spec.go
faganihajizada: Done. The three host-path joins in `spec.go` now use `filepath.Join`. I did not change fixed suffixes such as `overlayHostRoot + "/driver/dev"`, because a constant suffix has no separator problem. I can convert them too if you prefer one style.

internal/agent/gpudriver/gpudriver.go
faganihajizada: Yes. `writeKernelModules` now reads the path itself. The path is a package var, `kernelProcModules`, like `shimGlob` in `pcibus`, so tests can still point it at a fixture. It reads the agent's own `/proc/modules`, not `h.ProcPath`. The kernel keeps one global module list, and the DaemonSet mounts no `/host/proc`.

internal/kmod/modules.go
faganihajizada: Renamed to `nvidiaModuleSizeBytes` and `nvidiaUVMModuleSizeBytes`, and the field to `SizeBytes`. Dropping "Core" removes the SM-cores reading. I left out "File" because the value is the in-memory size, not the `.ko` size. `show_coresize` returns the size of the loaded text and core data (`kernel/module/main.c`).

internal/kmod/modules.go
faganihajizada: Renamed with the constants. See the thread on line 25.

internal/kmod/lsmod.go
faganihajizada: Moved out of the Go string into `internal/kmod/lsmod.sh`, embedded with `go:embed`, so it is a real shell file. I kept it beside the code. `go:embed` cannot read outside the package directory, so `shims/` would need a new internal-to-shims import, or an image that ships the file for the agent to copy. The agent already writes the script at stage time.

faganihajizada: Added to the `LsmodScript` doc. runc allows bind mounts inside `/proc` only for a short allowlist, and `/proc/modules` is not on it. So a container that gets the module surface only from the CDI mount reads the node's `/proc/modules`. The real `lsmod` takes its module list from that file, so it omits `nvidia`, although `/sys/module/nvidia` is mounted. The script reads the mounted tree instead.

deployments/nvml-mock/helm/nvml-mock/templates/daemonset.yaml
faganihajizada: No extra privilege. The agent reads `/proc/modules` in its own container. The kernel keeps one global module list, so that file shows the node's modules. The file is mode 0444, and on kind the agent reads it with `drop: [ALL]` plus `MKNOD`. It needs no `/host/proc` mount, and the comment on `kernelProcModules` says why.

internal/agent/gpudriver/stage.go
faganihajizada: Done. `kmod.ParseProcModules` parses the text once into a `HostModules` value, and `Render` and `ProcModules` both take that value. It keeps the raw text too, because `ProcModules` passes host lines through unchanged, including lines the parser skips.

internal/kmod/render.go
faganihajizada: Not yet. Only `Render` calls `mirrorTree`, and nothing else mirrors a tree. `ib.copyTree` copies one directory level and stops at the first error, and `stageDMI` copies named attributes. `mirrorTree` skips unreadable entries and returns them, so the caller decides how to log them. This PR moved `pruneDir` to `fsutil` once `kmod` needed it too, and I would do the same here for a second caller.

internal/kmod/render.go
faganihajizada: Renamed to `OverlayRoot`, in both renderers. `pcisysfs.Options.Output` had the same name and meaning, and I did not want the two to drift. I dropped the `Dir` suffix, because a root is a directory. `cdi/spec.go` uses the same word (`overlayHostRoot`).

shims/libmockfs/src/shim.c
faganihajizada: Renamed to `libmockfs`: directory, `.so` and soname, Make targets, Dockerfile, the chart's `LD_PRELOAD`, the NRI preload list, and the docs. Old CHANGELOG entries keep the old name, because they record past releases. I kept `MOCK_PCI_ROOT`. The NRI plugin sets it in every container it injects, so renaming it changes what workloads see. That belongs in its own PR.

CHANGELOG.md
faganihajizada: Shortened to what users see. The delivery details are in `docs/helm-chart.md` and the MEP.

internal/kmod/modules.go
faganihajizada: Added to the package doc: the `/proc/modules` line layout and the `/sys/module/<name>/` attributes. It also says `lsmod` lists modules from the first and reads their columns from the second.

internal/kmod/render.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	return skipped, fsutil.PruneDir(root, func(name string) bool { return declared[name] })
+}
+
+func renderModules(root string, mods []Module, declared, loaded map[string]bool) error {
+	for _, m := range mods {
+		switch {
+		case declared[m.Name] && loaded[m.Name]:
+			continue
```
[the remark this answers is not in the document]
faganihajizada: Fixed. When the mirror cannot read `coresize`, `refcnt` or `initstate` for a host module, `fillHostGaps` writes it from the host's `/proc/modules` line. `TestRender_FillsAHostGapFromTheProcModulesLine` covers the missing-`refcnt` case.

internal/kmod/render.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+		if err := fsutil.Write(path, []byte(content), 0o644); err != nil {
+			return err
+		}
+	}
+
+	if _, err := os.Stat(filepath.Join(dir, "holders")); os.IsNotExist(err) {
+		return linkHolders(dir, host.holders)
+	}
```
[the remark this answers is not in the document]
faganihajizada: Fixed. `fillHostGaps` now reconciles `holders/` with the dependency field of the `/proc/modules` line, even when the mirror kept the directory. `TestRender_ReconcilesAHostHoldersDirectoryThatLostALink` covers it.

faganihajizada:
Done. I ran the `kmod` spec on kind with the a100 profile. `hostDriver` is now false, and the four blocks pass: `nvidia 62312448 1 nvidia_uvm`, `refcnt` 1, both versions `550.163.01`, and `nvidia_uvm,` in `/proc/modules`. I then changed the expected `refcnt` to "2", and the spec failed with `Expected <string>: 1 to equal <string>: 2`. Reverted, it passes again. CI has not run it yet. A report entry now flags a node with a real driver, so a skipped check shows in the log.

internal/kmod/modules.go
```diff
@@ -0,0 +1,158 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+// NVIDIA and NVIDIAUVM are simulated module names.
+const (
+	NVIDIA    = "nvidia"
+	NVIDIAUVM = "nvidia_uvm"
```
roma-glushko, line 25: Would be nice we also simulated `nvidia_modeset` and `nvidia_peermem`. That would close the loop on the driver list that need to keep in mind.

roma-glushko (dismissed):
Looks good for my end ✅ 

Thank you for addressing my comments!

giuliocalzo (dismissed):

internal/kmod/modules.go
faganihajizada: yes 💯

roma-glushko (approved):

PR #799, opened 2026-09-05
docs(e2e): correct the harness lifecycle and drop the variables nothing reads

faganihajizada:
The README described a lifecycle the suite no longer owns. It said `make e2e` builds the image, creates the Kind cluster, loads the image and installs the chart. Tilt does all four, and the suite only attaches and asserts.

Four documented variables are read by no Go file: `E2E_SKIP_BUILD`, `E2E_KEEP_CLUSTER`, `E2E_BUILDX_GHA_CACHE` and `E2E_GOLANG_VERSION`. They are removed from the table. The workflow also set `E2E_SKIP_BUILD`, which nothing read, so that line is removed too.

Three real variables were undocumented and are now in the table: `E2E_CLUSTER_NAME`, `E2E_KUBE_CONTEXT` and `E2E_PROFILES_DIR`.

The README named a `kind-nvml-mock-e2e` context in three places. The default is `kind-mokka`.

`E2E_IMAGE` now states the contract: the ref must already be on every node, because the reshaping scenarios pin the DaemonSet to it and the suite does not build or load it.

Found while implementing #797, but independent of it. This branch starts from `main` and shares no file with [#798](https://github.com/NVIDIA/k8s-test-infra/pull/798).

### DIFF
--- a/.github/workflows/nvml-mock-e2e-go.yaml
+++ b/.github/workflows/nvml-mock-e2e-go.yaml
@@ -19,8 +19,8 @@
 #
 # Image distribution: a single `build-nvmlmock-image` job builds the nvml-mock
 # image once, exports it to a tarball and uploads it as a run-scoped artifact.
-# Every leg downloads that artifact, stages it with `make image-load`, and the
-# harness kind-loads it (E2E_SKIP_BUILD=true) rather than rebuilding per leg.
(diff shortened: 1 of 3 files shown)


ArangoGutierrez (approved):
/lgtm

PR #801, opened 2026-09-07
refactor(nvml-mock-ctl): parse the command line with urfave/cli v3

giuliocalzo:
## What This PR Does

Replaces `nvml-mock-ctl`'s hand-rolled argument parsing with `urfave/cli` v3 — the library `cmd/node-agent` and `cmd/nri-plugin` already use — and splits the 621-line `main.go` into six focused files.

Before, one `flag.FlagSet` held every flag of every subcommand, so `--mode`, `--links`, `--type` and `--socket` were offered on all of them; a loop re-parsed the argument list to interleave global flags with positionals; the usage text was a string maintained by hand beside the code it described; and two switch statements carried `//nolint:cyclop` waivers.

Now each command declares only its own flags and positional arguments, help is generated from the commands themselves, and the read-modify-write of the override document (resolve target, take the flock, apply, validate the whole document, write atomically) lives in one place:

| file | holds |
| --- | --- |
| `main.go` | root command, exit-code mapping, usage-error reporting |
| `target.go` | `--gpu` resolution, config paths, the shared mutate/patch plumbing |
| `args.go` | positional argument declaration and range checks |
| `metrics.go` | `temp`, `power`, `fan`, `util`, `clocks`, `throttle`, `pstate` |
| `faults.go` | `fail`, `nvlink-error`, `sram-ecc`, `fabric-health` |
| `state.go` | `set`, `status`, `reset` |
| `watch.go` | `watch-allocations` |

Both `cyclop` waivers are gone. `--gpu` is now a `Required` flag on the mutating commands rather than a runtime check, and `nvml-mock-ctl sram-ecc --help` finally describes `sram-ecc` instead of every command advertising every flag.

## Why

`nvml-mock-ctl` was the only binary in the tree still parsing its own command line, and the shared `FlagSet` made its help actively misleading: every command listed every other command's flags. The hand-written usage string had to be kept in step with the code and with `docs/nvml-mock-ctl.md` by hand.

### Compatibility

Deliberately unchanged: command names and the `temperature`/`utilization` aliases, positional arguments, the `--file`/`--config` globals **and** their acceptance on either side of the subcommand (root flags are persistent in v3), the `MOCK_NVML_*` environment fallbacks including the "set but empty means unset" rule, the exit codes (2 for a bad invocation or invalid value, 1 for an override file that could not be locked, read or written), and every stderr message prefix (`invalid:`, `lock:`, `load:`, `write:`).

Verified with a differential harness that runs the same invocation through the old and new binaries and compares exit code, stdout and the resulting `overrides.yaml`: **65 of 68 invocations byte-identical**. The three differences are intentional:

- **`--gpu` must now follow the command.** It used to be global, so `nvml-mock-ctl --gpu 0 temp 85` worked; it is now `nvml-mock-ctl temp --gpu 0 85`. Every doc, script and e2e caller in the tree already uses the command-first form, so nothing in-tree changes. Called out in the CHANGELOG.
- Invoking an alias echoes the canonical name: `ok: temp applied to 2` rather than `ok: temperature applied to 2` (twice, for the two aliases).

The existing `main_test.go` passes unmodified and remains the regression harness. Four tests were added for the surface this refactor newly owns: global flags on either side of the command, a missing `--gpu` being a usage error, an unknown command exiting 2, and `--help` listing every command.

## Checklist

- [x] Tests pass (`go test ./cmd/... ./pkg/gpu/mockctl/`; `pkg/gpu/allocwatch` fails identically on `main` in my sandbox, where its unix-socket dial is blocked)
- [x] Linter passes (`golangci-lint run ./cmd/...` — 0 issues; full `make lint-fix` can't run in my sandbox, where its `gen` prerequisite cannot reach the module sumdb)
- [x] Documentation updated (`docs/nvml-mock-ctl.md` command reference regenerated from the new `--help`, and it now documents `watch-allocations`, which the old block omitted)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -216,6 +216,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   public NVML APIs.
 
 ### Changed
+- `nvml-mock-ctl` parses its command line with `urfave/cli` v3, the library the
+  node agent and NRI plugin already use, instead of one hand-rolled
+  `flag.FlagSet` shared by every subcommand. Each command now declares only its
+  own flags and positional arguments, so `nvml-mock-ctl <command> --help`
+  documents that command rather than offering `--mode`, `--links` and `--type`
+  on all of them, and `--help` is generated from the commands themselves rather
+  than a usage string maintained by hand beside them. Command names, aliases,
+  positional arguments, the `--file`/`--config` global flags and their
+  environment fallbacks, the exit codes (2 for a bad invocation or an invalid
+  value, 1 for an override file that could not be locked, read or written) and
+  every stderr message are unchanged. One invocation that used to work no longer
+  does: `--gpu` was global, so it could precede the command
+  (`nvml-mock-ctl --gpu 0 temp 85`), and it now has to follow it
+  (`nvml-mock-ctl temp --gpu 0 85`) — the form every doc, script and e2e caller
+  already uses.
 - Dependencies no longer ship in `vendor/`. Go resolves them through a module
   proxy — NVIDIA's DGXC Artifactory in CI and for the published `nvml-mock`
   image, the public proxy locally — so builds and `make gen` need network access
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -226,11 +226,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   positional arguments, the `--file`/`--config` global flags and their
   environment fallbacks, the exit codes (2 for a bad invocation or an invalid
   value, 1 for an override file that could not be locked, read or written) and
-  every stderr message are unchanged. One invocation that used to work no longer
-  does: `--gpu` was global, so it could precede the command
+  every stderr message are unchanged. Two invocations that used to work no
+  longer do: `--gpu` was global, so it could precede the command
   (`nvml-mock-ctl --gpu 0 temp 85`), and it now has to follow it
   (`nvml-mock-ctl temp --gpu 0 85`) — the form every doc, script and e2e caller
-  already uses.
+  already uses; and `status --gpu ""` is now a usage error rather than a report
+  of every override, so a script whose index variable came out empty is told
+  instead of being handed the whole node's state as if it had asked for it.
 - Dependencies no longer ship in `vendor/`. Go resolves them through a module
   proxy — NVIDIA's DGXC Artifactory in CI and for the published `nvml-mock`
   image, the public proxy locally — so builds and `make gen` need network access
--- a/cmd/nvml-mock-ctl/args.go
+++ b/cmd/nvml-mock-ctl/args.go
@@ -1,0 +1,84 @@
+// Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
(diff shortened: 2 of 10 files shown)


cmd/nvml-mock-ctl/args.go
```diff
@@ -0,0 +1,84 @@
...
+// types them; the arity and range checks that keep a typo from being published
+// as a plausible reading live here.
+//
+// A required single value is declared as a one-value slice rather than the
+// singular Arg types, which accept an absent value and leave the zero value
+// behind. With Min 1 the action only runs once the value has parsed, so the
+// readers below can index the result.
+
+func requiredIntArg(name string) cli.Argument {
+	return &cli.IntArgs{Name: name, UsageText: name, Min: 1, Max: 1}
```
ArangoGutierrez, line 34: requiredIntArg leaves IntegerConfig.Base at zero, so cli.IntArgs parses with strconv.ParseInt base inference where the old singleIntArg used strconv.Atoi. `temp --gpu 0 010` now writes 8 instead of 10 and still exits 0, `085` is rejected as invalid syntax, and `0x0a` is newly accepted. Setting `Config: cli.IntegerConfig{Base: 10}` keeps celsius, percent, mhz, pstate and count decimal.

cmd/nvml-mock-ctl/main_test.go
```diff
@@ -269,6 +270,63 @@ func TestCLI_StatusFilterByGPU(t *testing.T) {
...
+		}
+		var out, errb bytes.Buffer
+		code := run(full, &out, &errb)
+		require.Equalf(t, 0, code, "args %v exited %d: %s", args, code, errb.String())
+		require.Contains(t, readConfigOverride(t, configOverride), "temperature_gpu_c: 85")
+	}
+}
+
+// An explicitly empty target must not be read as "every device". Only reset
+// applies to everything, and only when --gpu is absent altogether, so a
```
ArangoGutierrez, line 295: The second clause does not hold: reset uses optionalGPUFlag, which carries no validator, so `reset --gpu ""` still reaches the all-devices fallback and clears every bucket at exit 0. Either drop that half of the sentence, or put the same validator on optionalGPUFlag, where it only fires when the flag is present and so leaves the bare `reset` default alone.

CHANGELOG.md
```diff
@@ -216,6 +216,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+  node agent and NRI plugin already use, instead of one hand-rolled
+  `flag.FlagSet` shared by every subcommand. Each command now declares only its
+  own flags and positional arguments, so `nvml-mock-ctl <command> --help`
+  documents that command rather than offering `--mode`, `--links` and `--type`
+  on all of them, and `--help` is generated from the commands themselves rather
+  than a usage string maintained by hand beside them. Command names, aliases,
+  positional arguments, the `--file`/`--config` global flags and their
+  environment fallbacks, the exit codes (2 for a bad invocation or an invalid
+  value, 1 for an override file that could not be locked, read or written) and
+  every stderr message are unchanged. Two invocations that used to work no
```
ArangoGutierrez, line 229: The parity claim does not hold for stderr. `--gpu is required` is now `Required flag "gpu" not set`, `power value "NaN" must be finite` is now `watts value NaN must be finite`, and `status --gpu expects an integer index, got "abc"` is now the IntFlag parse error. Narrowing the sentence to the exit codes, which are preserved, keeps it true for anyone grepping stderr.

ArangoGutierrez (approved):
Thanks for this, having each command declare its own flags and positional arguments makes `nvml-mock-ctl sram-ecc --help` describe sram-ecc instead of every command advertising every flag, and the empty-target validator in the follow-up closes a real hole.

Splitting the 621-line main.go into per-command files and centralising the read-modify-write in mutate() is a clear improvement, and the follow-up commit's validator plus its regression test handle the empty-target case well. What is left is a few spots where the new parser's behavior drifts from the old one without the CHANGELOG saying so.

- MOCK_NVML_PODRESOURCES_SOCKET is not templated by the chart: daemonset.yaml passes `--socket={{ .Values.allocationWatcher.podResourcesSocket }}` and its env block sets only MOCK_NVML_CONFIG and MOCK_NVML_OVERRIDES. The flag path is the stronger justification anyway, since an empty values entry renders a literal `--socket=`. (cmd/nvml-mock-ctl/watch.go:78)

PR #808, opened 2026-09-08
fix(mocknvml): keep device indices consistent after visibility filtering

### LINKED ISSUE #807: [Bug]: Mock NVML GetIndex does not round-trip after device visibility filtering
### Component / version

Mock NVML at `874ab0319732981bc79cbc9b01c22c8fd00656e6`, built from source.

### What happened

With two configured mock GPUs and only `/dev/nvidia1` present in the container:

```text
nvmlDeviceGetCount_v2()          -> 1
nvmlDeviceGetHandleByIndex_v2(0) -> SUCCESS, handle h
nvmlDeviceGetIndex(h)            -> 1
nvmlDeviceGetHandleByIndex_v2(1) -> NVML_ERROR_INVALID_ARGUMENT
```

### Expected behaviour

For a non-MIG device and an unchanged device set, the index returned by `GetIndex` should resolve back to the same handle through `GetHandleByIndex_v2`.

`GetHandleByIndex` maps through `visibleDevices`, while `GetIndex` returns the original `d.index`. Is preserving the original index intentional? How should these two APIs stay consistent? This report does not assume that a container's sole accessible GPU must always have NVML index 0.

### Steps to reproduce

Use the manifest below on a node with the mock library already installed. Adjust `nodeSelector` and `hostPath.path` and create the `hami-tests` namespace if needed. No HAMi, device plugin, real GPU, or privileged container is required.

The ConfigMap supplies an empty `/dev/nvidia1` file because the mock checks device presence using `os.Stat`.

<details>
<summary>nvml-index-repro.yaml</summary>

```yaml
# Requires a node with the mock NVML library already installed.
# Adjust nodeSelector and hostPath for your cluster.
apiVersion: v1
kind: ConfigMap
metadata:
  name: nvml-index-repro
  namespace: hami-tests
data:
  nvidia1: ""
---
apiVersion: v1
kind: Pod
metadata:
  name: nvml-index-repro
  namespace: hami-tests
spec:
  restartPolicy: Never
  nodeSelector:
    kubernetes.io/hostname: hami-test-worker
  containers:
    - name: probe
      image: python:3.12-slim-bookworm
      env:
        - name: MOCK_NVML_NUM_DEVICES
          value: "2"
      command: [python3, -u, -c]
      args:
        - |
          import ctypes as C
          import sys

          lib = C.CDLL("/opt/mock/libnvidia-ml.so.1")
          def function(name, *args):
              fn = getattr(lib, name)
              fn.argtypes = list(args)
              fn.restype = C.c_int
              return fn

          uintp, handlep = C.POINTER(C.c_uint), C.POINTER(C.c_void_p)
          init = function("nvmlInit_v2")
          shutdown = function("nvmlShutdown")
          count = function("nvmlDeviceGetCount_v2", uintp)
          get = function("nvmlDeviceGetHandleByIndex_v2", C.c_uint, handlep)
          index = function("nvmlDeviceGetIndex", C.c_void_p, uintp)
          minor = function("nvmlDeviceGetMinorNumber", C.c_void_p, uintp)
          def check(ret):
              if ret != 0:
                  raise RuntimeError(f"NVML call failed: {ret}")

          check(init())
          try:
              n = C.c_uint()
              check(count(C.byref(n)))
              print(f"count={n.value}")
              if n.value != 1:
                  raise RuntimeError("Expected one visible device; check the mock configuration")
              h, back = C.c_void_p(), C.c_void_p()
              idx, physical_minor = C.c_uint(), C.c_uint()
              check(get(0, C.byref(h)))
              check(index(h, C.byref(idx)))
              check(minor(h, C.byref(physical_minor)))
              ret = get(idx.value, C.byref(back))
              same = ret == 0 and h.value == back.value
              print(f"enumerated=0 returned_index={idx.value} minor={physical_minor.value}")
              print(f"roundtrip_ret={ret} same_handle={int(same)}")
              passed = same and physical_minor.value == 1
              print("PASS" if passed else "FAIL: returned index does not resolve to the same device")
              sys.exit(0 if passed else 1)
          finally:
              check(shutdown())
      volumeMounts:
        - name: mock-library
          mountPath: /opt/mock/libnvidia-ml.so.1
          readOnly: true
        - name: visible-device
          mountPath: /dev/nvidia1
          subPath: nvidia1
          readOnly: true
  volumes:
    - name: mock-library
      hostPath:
        path: /var/lib/nvml-mock/driver/usr/lib64/libnvidia-ml.so.1
        type: File
    - name: visible-device
      configMap:
        name: nvml-index-repro
```

</details>

```sh
kubectl --context kind-hami-test apply -f nvml-index-repro.yaml
kubectl --context kind-hami-test -n hami-tests logs -f nvml-index-repro
# After collecting the output:
kubectl --context kind-hami-test delete -f nvml-index-repro.yaml
```

Replace the context name for your cluster. If logs report `ContainerCreating`, retry after startup.

### Environment

Linux/amd64, Ubuntu 24.04.4 LTS, kernel 6.8.0-138-generic; Kind 0.31.0, Kubernetes 1.35.0, containerd 2.2.0. CPU-only host, no physical NVIDIA GPU/driver. Probe image: `python:3.12-slim-bookworm`.

### Logs / artefacts

Original library, Pod exits with code 1:

```text
count=1
enumerated=0 returned_index=1 minor=1
roundtrip_ret=2 same_handle=0
FAIL: returned index does not resolve to the same device
```

I tested a candidate fix that maintains a separate visible NVML index while retaining physical device identity. With the same manifest and only the mounted library changed, the Pod exits with code 0:

```text
count=1
enumerated=0 returned_index=0 minor=1
roundtrip_ret=0 same_handle=1
PASS
```

The candidate also has Go regression coverage for unfiltered enumeration and visible sets `[0]`, `[1]`, `[1,3]`, and `[2,3]`. I have not compared it against real-driver container visibility behaviour.

**I can work on the fix and submit a PR once we agree on the intended indexing behaviour.**

### Pre-submission checks

- [x] I searched existing issues and Discussions and did not find a duplicate. Related #306 introduced visibility filtering.
- [x] This reports a reproducible API inconsistency, rather than a usage question.

jimyag:
## What This PR Does

When only physical GPU 1 is visible, `GetHandleByIndex(0)` returns its handle but `GetIndex` returns 1, which cannot be resolved back. Keep a separate visible NVML index so these calls round-trip, while preserving physical device identity.

- Keep GPU reset keyed by physical index, so resetting visible GPU 0 does not clear another GPU's overrides.
- Filter hidden devices from topology results and return `ERROR_INVALID_ARGUMENT` when querying a hidden handle's index or topology.
- Exclude hidden GPUs from Xid delivery and device-loss checks without consuming their pending events.
- Copy visibility mappings so later caller mutations cannot desynchronize enumeration and cached indices.
- Add regressions for filter changes, mapping ownership, UUID/PCI identity, topology visibility, event isolation, and per-device reset, including rejection of hidden reset handles. Make the topology test independent of host device files.
- Explain the reset visibility/health guard and distinguish visible and physical indices in reset logs.

## Why

Fixes #807.

This follows the mock's existing filtered enumeration model. It does not claim to reproduce every real-driver container visibility behaviour.

Validation on Linux/amd64: `make test` (race detection and coverage), `make lint-fix`, and targeted engine/bridge race tests passed. The standalone shared-library probe also verified the index round trip. Local macOS engine race tests passed.

## Checklist

- [x] Documentation updated (if applicable) — changelog updated; no setup changes

### DIFF
--- a/pkg/gpu/mocknvml/engine/device.go
+++ b/pkg/gpu/mocknvml/engine/device.go
@@ -428,14 +428,15 @@ func (d *ConfigurableDevice) GetConfig() *DeviceConfig {
 	return d.cfg()
 }
 
-// GetIndex returns the device index
+// GetIndex returns the zero-based ordinal among visible devices.
+// A handle for a hidden device returns ERROR_INVALID_ARGUMENT.
 func (d *ConfigurableDevice) GetIndex() (int, nvml.Return) {
 	if ret := d.handleLookupReturn(); ret != nvml.SUCCESS {
 		return 0, ret
 	}
 	index := int(d.nvmlIndex.Load())
 	if index < 0 {
-		return 0, nvml.ERROR_NO_PERMISSION
+		return 0, nvml.ERROR_INVALID_ARGUMENT
 	}
 	debugLog("[NVML] nvmlDeviceGetIndex -> %d\n", index)
 	return index, nvml.SUCCESS
--- a/pkg/gpu/mocknvml/engine/device.go
+++ b/pkg/gpu/mocknvml/engine/device.go
@@ -46,6 +46,7 @@ type ConfigurableDevice struct {
 	*dgxa100.Device
 	fabric      *NodeFabric
 	index       int
+	nvmlIndex   atomic.Int64
 	minorNumber int
 
 	// baseConfig is the pristine, merged (defaults+per-device) YAML config.
@@ -99,6 +100,8 @@ func NewConfigurableDevice(index int, baseDevice *mockserver.Device, config *Dev
 		minorNumber: minorNumber,
 		baseConfig:  config,
 	}
+
+	dev.nvmlIndex.Store(int64(index))
 
 	// Override base device properties from config
 	applyDeviceBaseOverrides(dev, config)
@@ -430,8 +433,18 @@ func (d *ConfigurableDevice) GetIndex() (int, nvml.Return) {
 	if ret := d.handleLookupReturn(); ret != nvml.SUCCESS {
 		return 0, ret
 	}
-	debugLog("[NVML] nvmlDeviceGetIndex -> %d\n", d.index)
-	return d.index, nvml.SUCCESS
+	index := int(d.nvmlIndex.Load())
+	if index < 0 {
+		return 0, nvml.ERROR_NO_PERMISSION
+	}
+	debugLog("[NVML] nvmlDeviceGetIndex -> %d\n", index)
+	return index, nvml.SUCCESS
+}
+
+// PhysicalIndex identifies the device in the shared override configuration.
+// It must not change when a consumer sees only a subset of the devices.
+func (d *ConfigurableDevice) PhysicalIndex() int {
+	return d.index
 }
 
 // GetUUID returns the device UUID. Overrides the embedded dgxa100.Device
@@ -2500,4 +2513,20 @@ func (s *MockServer) isDeviceVisible(deviceIndex int) bool {
 		return true
 	}
 	return slices.Contains(s.visibleDevices, deviceIndex)
+}
+
+// setVisibleDevices keeps enumeration indices consistent without changing physical
+// device identity (minor number, PCI address, or topology index).
+func (s *MockServer) setVisibleDevices(visible []int) {
+	s.visibleDevices = visible
+	for physical, d := range s.configurableDevices {
+		if d == nil {
+			continue
+		}
+		index := physical
+		if visible != nil {
+			index = slices.Index(visible, physical)
+		}
+		d.nvmlIndex.Store(int64(index))
+	}
 }
--- a/pkg/gpu/mocknvml/engine/device.go
+++ b/pkg/gpu/mocknvml/engine/device.go
@@ -2398,6 +2398,7 @@ type MockServer struct {
 	// /dev/nvidia* nodes (via CDI injection), this slice contains only the
 	// devices whose device nodes exist, mimicking real NVML's cgroup-based
 	// device filtering. If nil, all devices are visible (no filtering).
+	// Assign only through setVisibleDevices to keep cached NVML indices in sync.
 	visibleDevices []int
 }
 
@@ -2519,14 +2520,14 @@ func (s *MockServer) isDeviceVisible(deviceIndex int) bool {
 // setVisibleDevices keeps enumeration indices consistent without changing physical
 // device identity (minor number, PCI address, or topology index).
 func (s *MockServer) setVisibleDevices(visible []int) {
-	s.visibleDevices = visible
+	s.visibleDevices = slices.Clone(visible)
 	for physical, d := range s.configurableDevices {
 		if d == nil {
 			continue
 		}
 		index := physical
-		if visible != nil {
-			index = slices.Index(visible, physical)
+		if s.visibleDevices != nil {
+			index = slices.Index(s.visibleDevices, physical)
 		}
 		d.nvmlIndex.Store(int64(index))
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,7 +36,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the same minor number — including where one of them defaulted to it — or when
   a minor falls outside the range a GPU node can carry.
 - mocknvml: keep device indices consistent after visibility filtering, preserve
-  physical GPU targets for reset, and exclude hidden devices from topology results (#807).
+  physical GPU targets for reset, and exclude hidden devices from topology results
+  and event waits (#807).
 - mocknvml: `nvidia-smi --gpu-reset` (`-r`) now resets a GPU instead of
   segfaulting. The mock's export-table dispatcher ended every per-device call by
   writing a zero count through `arg1`, which the reset slots do not carry, so the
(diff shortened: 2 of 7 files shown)


pkg/gpu/mocknvml/engine/device.go
```diff
@@ -430,8 +433,18 @@ func (d *ConfigurableDevice) GetIndex() (int, nvml.Return) {
 	if ret := d.handleLookupReturn(); ret != nvml.SUCCESS {
 		return 0, ret
 	}
-	debugLog("[NVML] nvmlDeviceGetIndex -> %d\n", d.index)
-	return d.index, nvml.SUCCESS
+	index := int(d.nvmlIndex.Load())
+	if index < 0 {
+		return 0, nvml.ERROR_NO_PERMISSION
```
ArangoGutierrez: nvml.h documents nvmlDeviceGetIndex as returning only SUCCESS, ERROR_UNINITIALIZED, ERROR_INVALID_ARGUMENT, ERROR_GPU_IS_LOST and ERROR_UNKNOWN, so NO_PERMISSION is a code no go-nvml or DCGM caller has a branch for, and PendingXidEvent makes it reachable outside tests. ERROR_INVALID_ARGUMENT is what the header assigns to an invalid device. The comment above still reads "GetIndex returns the device index", which now describes neither the ordinal nor this error.

ArangoGutierrez (dismissed):
Thanks for this, and for the repro in #807. Splitting the visible ordinal from the physical identity is the right shape of fix, and the round-trip test covers the cases that matter.

The visibility invariant this establishes is worth carrying a little further, and one of the new tests depends on the host it runs on.

- The same filter is missing from PendingXidEvent and AnyDeviceLost, which still walk every entry of configurableDevices. PendingXidEvent is now the only path that registers a handle for a hidden GPU, so a container that sees one device can still receive an Xid event carrying another GPU's handle, and one lost GPU on the node fails nvmlEventSetWait for every container on it. (pkg/gpu/mocknvml/engine/engine.go:437)
- newFabricEngine runs Init, so detectVisibleDevices stats the real /dev/nvidia* nodes. On a host where only /dev/nvidia1 is present, index 0 already resolves to physical device 1, hidden and h become the same device, and the ERROR_NO_PERMISSION assertion fails. Calling SetVisibleDevicesForTesting(nil) before taking the hidden handle pins it to the intended state. (pkg/gpu/mocknvml/engine/engine_test.go:733)
- With the value discarded, this call is now the only thing stopping a reset on a device the container cannot see, which makes it easy to mistake for dead code and drop later. The two debugLog lines below also print the physical index, which no longer matches the GPU number nvidia-smi showed the user under filtering. (pkg/gpu/mocknvml/bridge/gpu_reset.go:57)

ArangoGutierrez (dismissed):
Thanks for this, the diagnosis is right and the fix is placed well.

The part I care most about is the reset. Overrides are keyed by physical index
(`doc.DeviceConfigOverride(d.index)` in engine/device.go), so before this change
a container that saw only physical GPU 1 and asked to reset its GPU 0 would
clear physical GPU 0's overrides, which is another container's GPU. Routing the
reset through `PhysicalIndex()` and rejecting hidden handles first is the right
shape.

I checked that the invariant holds rather than just the one call path.
`setVisibleDevices` is now the only writer of `visibleDevices`, and every device
sweep in the engine is gated on `isDeviceVisible` with one exception,
`ProcessByPID` at engine.go:401. That one may be deliberate, since real
`nvmlSystemGetProcessName` reads /proc and is not GPU scoped, but it is the last
ungated sweep so a sentence either way would help the next reader.

I also mutation-checked the new tests, since a guard that does not go red is
worse than no guard. All five die: reverting the reindex in `setVisibleDevices`,
swapping `physicalIndex` back to `visibleIndex` in the reset, dropping either
topology gate, dropping the event gates, and dropping the `index < 0` check in
`GetIndex`. The `{3, 1}` and `{}` rows in the round-trip table are doing real
work, they pin order independence and the zero-visible case.

Three small things, none blocking:

`nvmlIndex` is now a cache derived from `visibleDevices`, and nothing at compile
time keeps them together. Both call sites assigned `visibleDevices` directly
before this PR, so a future one would desynchronize the two silently. A one-line
comment on the field saying to assign only through `setVisibleDevices` would
pin it. Related, `setVisibleDevices` keeps the caller's slice rather than
copying it, so a caller mutating it in place has the same effect.

The hidden-handle rejection in `mockInternalResetGPU` is not covered. The new
test only exercises a visible handle, so the guard the comment describes is
untested. Holding a handle for physical 0, setting visible to `{1}`, and
asserting the call returns 0 with the override file unchanged would cover it.

If the calling handle is itself hidden, `TopologyNearestGpus` still returns
SUCCESS with the visible peers instead of an error. Not reachable in practice,
since you cannot get a handle for a hidden device, so this is consistency only.

Locally: `go vet`, `golangci-lint run ./pkg/gpu/mocknvml/...` (0 issues), and
`go test -race ./pkg/gpu/mocknvml/...` are all green on the branch.

jimyag:
Addressed in 5ed01683 and rebased onto main.

Added regressions for caller-side slice mutation, hidden topology handles, and hidden GPU reset preserving the override file. The first two failed before the fixes and pass now.

Also clarified why ProcessByPID ignores GPU visibility.

Targeted race tests on macOS and full `make test` and `make lint-fix` on Linux/amd64 pass.

Thanks for the thorough review!

giuliocalzo (dismissed):
/lgtm

giuliocalzo (approved):
sorted the conflicts and rebased

PR #825, opened 2026-09-10
ci(control-plane): build and publish the control-plane image

### LINKED ISSUE #741: MEP0001: Build, Validate, Publish Control Plane Image in CI/CD

ArangoGutierrez:
## What This PR Does

Adds `.github/workflows/control-plane-publish.yaml`, which builds and pushes the
Mokka control plane to `ghcr.io/nvidia/mokka-control-plane` on the same triggers
as the nvml-mock image: pushes to `main`, `v*` tags, and `workflow_dispatch`.

`nvml-mock-publish.yaml` is untouched.

## Why

`deployments/control-plane/Dockerfile` and `cmd/control-plane` have been in tree
since MEP0001 landed, but no workflow ever built or pushed the image. Searching
all thirteen workflows on `main` for `control-plane` returns nothing, so the
control plane could only be run from a local build. With v0.4.0 shipping MEP0001
that leaves the headline feature unconsumable from a release artifact.

## Why a separate workflow rather than a matrix

`nvml-mock-publish.yaml` puts `IMAGE_NAME` in workflow-level `env:` and is named
for its image, so a matrix would mean restructuring the release-critical path of
an already-shipping image in order to add a second one, and would couple both
images' pass/fail signal and their `workflow_dispatch` re-runs. A second file
keeps the shipping path untouched.

## Notes for review

The new file mirrors the existing job step for step, including the multi-arch
build, the DGXC Go proxy handling (both the `GOPROXY` build arg and the `goauth`
secret, which are required together), keyless cosign signing, the SBOM
attestation, and the credential cleanup. All nine pinned action SHAs are
identical to the ones already in `nvml-mock-publish.yaml`.

Two deliberate points:

- There is no `paths:` filter, for the reason the existing file documents:
  GitHub applies paths filters to tag pushes, a tag push has no changed-files
  diff, and the v0.2.0 tag published no image because of exactly that.
- `cache-from`/`cache-to` carry `scope=control-plane`. The BuildKit GHA cache is
  keyed by scope name repository-wide, so leaving it unset would share the
  default bucket with nvml-mock Publish and the two images would evict each
  other. This follows `scope=nvml-mock` and `scope=kind-node` in
  `nvml-mock-e2e-go.yaml`. It is worth giving nvml-mock Publish an explicit scope
  too, but that is a separate change.

The workflow has never executed. Merging to `main` exercises it before any tag
is pushed, which is the point at which OIDC, keyless signing and first-time GHCR
package creation get proven.

## Checklist

Verified locally: `actionlint` clean on the new file, the YAML parses, the image
builds from `deployments/control-plane/Dockerfile` (arm64, distroless, nonroot),
and a second build supplying the `GOPROXY` build arg and the `goauth` secret also
succeeds.

### DIFF
--- a/.github/workflows/control-plane-publish.yaml
+++ b/.github/workflows/control-plane-publish.yaml
@@ -1,0 +1,136 @@
+# Copyright 2026 NVIDIA CORPORATION
+# SPDX-License-Identifier: Apache-2.0
+name: control-plane Publish
+
+permissions: {}
+
+# No paths filter here: GitHub applies paths filters to tag pushes too, and
+# a tag push has no changed-files diff, so a paths filter silently suppresses
+# every tag-triggered run (the v0.2.0 tag published no image until it was
+# retagged by hand). workflow_dispatch allows re-running a publish at any ref
+# without re-pushing tags.
+on:
+  push:
+    branches:
+      - main
+    tags:
+      - 'v*'
+  workflow_dispatch:
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: nvidia/mokka-control-plane
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      id-token: write  # needed for keyless cosign signing
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@1f40c72289eff860ee54a304f1438e3cff362e0a # v4
+
+      - name: Set up Docker Buildx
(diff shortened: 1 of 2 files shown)


.github/workflows/control-plane-publish.yaml
```diff
@@ -0,0 +1,136 @@
...
+
+permissions: {}
+
+# No paths filter here: GitHub applies paths filters to tag pushes too, and
+# a tag push has no changed-files diff, so a paths filter silently suppresses
+# every tag-triggered run (the v0.2.0 tag published no image until it was
+# retagged by hand). workflow_dispatch allows re-running a publish at any ref
+# without re-pushing tags.
+on:
+  push:
```
roma-glushko, line 13: Would be nice to build them on PR changes, so we see failures early.

roma-glushko (approved):

PR #828, opened 2026-09-10
test(gpu): cover minor-number mappings without root

### LINKED ISSUE #827: test(agent,engine): two minor-number paths from #823 have no coverage that runs
Follow-up from #823, which is correct. Two of the three behavioural
surfaces it changes have no test that actually executes, so a regression in
either would go unnoticed.

## 1. stageCharDevs has coverage that never runs

`TestStageCharDevs_UsesConfiguredMinorNumber` (and the two older
`TestStageCharDevs_*` tests) begin with `skipUnlessRootLinux`, which skips
unless `runtime.GOOS == "linux" && os.Getuid() == 0`.

CI runs `make test` on `ubuntu-latest` with no container and no sudo, so the
runner is uid 1001 and all three skip. From the last green Golang job on main:

    --- SKIP: TestStageCharDevs_CreatesDeviceNodes
        gpudriver_test.go:194: requires root on Linux (mknod)
    --- SKIP: TestStageCharDevs_PrunesShrunkDeviceSet

Reverting `d.MinorNumber` to `d.Index` in `internal/agent/gpudriver/stage.go`
leaves the whole gpudriver package green, which confirms nothing catches it.

Suggestion: split the node-name and minor computation out of `stageCharDevs`
into a pure function and assert on that, so the mapping is covered without mknod.
The mknod-dependent part can stay root-gated.

## 2. The visibility filter's wiring is untested

`TestDetectVisibleDevices_ReportsIndicesNotMinorNumbers` calls
`detectVisibleDevicesAt(dir+"/nvidia%d", []int{1, 0, 3, 2})` with the slice
built by hand. That covers the helper, but not the loop in
`detectVisibleDevices` that derives the slice from
`config.GetDeviceMinorNumber(i)`.

Changing that loop to append `i` instead leaves the engine package green.

Suggestion: one test that goes through a `Config` with a non-identity
minor mapping and asserts the reported indices.

## Why this matters

#823 exists because index and minor were conflated. These two paths are exactly
where that conflation would come back, and they are the two the suite cannot
currently see.

sylvesterkaczmarek:
## What This PR Does

Extract the character-device names and major/minor pairs into a pure function used by staging. Move configuration-to-minor resolution inside the visibility helper so ordinary tests exercise the same path as production.

## Why

Closes #827.

The existing staging checks require root, and the visibility tests supplied the minor-number list themselves. Both index/minor regressions could therefore pass CI. The new coverage includes reordered minors and explicit minor zero; the privileged device-creation tests remain unchanged.

## Validation

- `make test`: 1,032 tests passed across 43 packages; 16 existing environment-dependent tests skipped on macOS arm64.
- `make lint-fix`: passed, including vet and govulncheck. The pinned golangci-lint v2.12.2 was rebuilt with Go 1.27.1 to match the local toolchain.
- Replacing the staging minor numbers with indices fails the new staging assertion.
- Replacing configured visibility minors with indices fails the visibility assertion. Both mutations were reverted.

## Checklist

- [x] Commits are signed off and cryptographically signed.
- [x] Tests pass.
- [x] Linter passes.
- [x] Source files retain their license headers.
- [x] Documentation and changelog reviewed; no user-facing behavior changes.

### DIFF
--- a/internal/agent/gpudriver/gpudriver_test.go
+++ b/internal/agent/gpudriver/gpudriver_test.go
@@ -190,6 +190,22 @@ func TestStageNVMLShim_CopiesLibAndCreatesLinks(t *testing.T) {
 	}
 }
 
+func TestCharDevsForDevices_UsesConfiguredMinorNumbers(t *testing.T) {
+	t.Parallel()
+
+	devices := []agent.DeviceSpec{
+		{Index: 0, MinorNumber: 3},
+		{Index: 1, MinorNumber: 0},
+	}
+	require.Equal(t, []charDev{
+		{"nvidia3", 195, 3},
+		{"nvidia0", 195, 0},
+		{"nvidiactl", 195, 255},
+		{"nvidia-uvm", 510, 0},
+		{"nvidia-uvm-tools", 510, 1},
+	}, charDevsForDevices(devices))
+}
+
 func TestStageCharDevs_CreatesDeviceNodes(t *testing.T) {
 	skipUnlessRootLinux(t)
 
--- a/internal/agent/gpudriver/stage.go
+++ b/internal/agent/gpudriver/stage.go
@@ -22,6 +22,25 @@ import (
 	"github.com/NVIDIA/k8s-test-infra/pkg/gpu/mocknvml/engine"
 )
 
+type charDev struct {
+	name         string
+	major, minor uint32
+}
+
+func charDevsForDevices(devices []agent.DeviceSpec) []charDev {
+	devs := make([]charDev, 0, len(devices)+3)
+	for _, d := range devices {
+		devs = append(devs, charDev{fmt.Sprintf("nvidia%d", d.MinorNumber), 195, uint32(d.MinorNumber)})
+	}
+	devs = append(devs,
+		charDev{"nvidiactl", 195, 255},
+		charDev{"nvidia-uvm", 510, 0},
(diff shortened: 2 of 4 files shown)


ArangoGutierrez:
Hey @sylvesterkaczmarek !!!   
Thanks for your contribution; we will review it ASAP!

ArangoGutierrez:
/ok to test [6dbca77](https://github.com/NVIDIA/k8s-test-infra/pull/828/commits/6dbca773c6386fb95c943603d801fe5f1dde1e7d)

ArangoGutierrez (commented):
Thanks for this, the minor-number handling from #823 finally has coverage that runs in CI.

The diff reads like a pure code move, but it changes what the suite catches. On 28cef0d5, naming the chardev from d.Index instead of d.MinorNumber leaves `go test ./internal/agent/gpudriver/...` green, and making detectVisibleDevices ignore GetDeviceMinorNumber leaves `go test ./pkg/gpu/mocknvml/engine/...` green. Both go red here. TestStageCharDevs_UsesConfiguredMinorNumber nominally covers the first one, but it sits behind skipUnlessRootLinux and golang.yaml runs on ubuntu-latest as a non-root user, so it skips on every CI run. Pulling charDevsForDevices out and handing detectVisibleDevicesAt the *Config is what moves both invariants onto the path CI actually executes.

giuliocalzo (approved):
/lgtm

Issue #614, opened 2026-08-07
[MEP0001] Mokka Control Plane: Implementation


roma-glushko:
This is implementation of [MEP0001](https://github.com/NVIDIA/k8s-test-infra/pull/536) that includes:
- creation of a new component called Mokka Control Plane
- adding Redis as runtime dependency to the Mokka helm chart. Deploying a local instance of redis
- Implementation of the sGPU Controller that manages SGPUInventory, SGPUProfile, SGPURuntimePolicy objects
- Creation of a new CRD chart
- Functional implementation of sGPU node <-> K8s Node assignment
- Functional implementation of sGPU rack and node runtime state management via Redis
- Rethinking of the current structure of the mock CLIs to turn them into a proper node agent (a new MEP is required)
- Control Plane to Node Agent State propagation

Related issues:
- https://github.com/NVIDIA/k8s-test-infra/issues/598

Issue #254, opened 2026-03-13
tracking: NVML API coverage gaps


ArangoGutierrez:
## Overview

Remaining NVML API gaps compared to nvml-mock. As of PR #240, we're at ~95% parity (~60/64 functions). These issues track the remaining ~4 function gaps plus infrastructure improvements.

## Sub-issues

- [ ] #241 — Full MIG lifecycle (4 remaining functions)
- [ ] #242 — PCI sysfs simulation for NFD discovery
- [ ] #243 — Symbol version script (.map) for dlopen compat
- [ ] #244 — Dynamic metrics (temperature, power, utilization)

## Context

See `.agents/reports/2026-02-22-cross-project-comparison.md` for the full gap analysis.

PR #829, opened 2026-09-11
feat(nvml-mock): implement the full MIG lifecycle

### LINKED ISSUE #241: feat(nvml-mock): implement full MIG lifecycle (4 remaining functions)
## Context

The mock NVML currently handles MIG as "disabled" — `GetMaxMigDeviceCount` returns 0 and `GetMigDeviceHandleByIndex` returns `NOT_FOUND`. This is sufficient for the GPU Operator with `mig.strategy=none`, but doesn't support MIG-enabled testing.

nvml-mock implements 7/7 MIG functions; we have 3/7.

## Scope

Implement the remaining 4 MIG functions:
- `CreateGpuInstance` / `DestroyGpuInstance`
- `CreateComputeInstance` / `DestroyComputeInstance`

These should store state in-memory (not persisted across restarts) and be driven by YAML profile configuration for MIG-capable GPUs (A100, H100).

## Acceptance Criteria

- [ ] MIG instances can be created/destroyed via NVML API
- [ ] `GetMigDeviceHandleByIndex` returns valid handles after creation
- [ ] Engine unit tests cover create/destroy lifecycle
- [ ] A100 and H100 profiles include MIG partition definitions
- [ ] device-plugin with `migStrategy=single` works against mock

## References

- Cross-project comparison: `.agents/reports/2026-02-22-cross-project-comparison.md`
- Enhancement plan Section 3.2 (MIG): `.agents/plans/2026-02-22-gpu-mock-enhancement-plan.md`

giuliocalzo:
## What This PR Does

Makes MIG work in `nvml-mock` the way it works against a real driver.

`nvidia-smi -mig 1`, `nvidia-smi mig -cgi/-cci/-dgi/-dci`, `mig -lgi/-lci` and `nvidia-smi -L` all run unmodified against the mock, and a partition made by one command is visible to every other consumer on the node. A board boots partitioned with `gpu.mig.enabled`, and `gpu.mig.gpuInstances` names the partitioning to apply. No profile ships a layout of its own: how a board is carved is a deployment choice, so a MIG install always states it, and a capable board given none is refused rather than coming up enabled with nothing partitioned. With MIG on, the device plugin serves MIG resources under `migStrategy=single` and the node agent publishes CDI entries per partition, so a MIG pod is scheduled and admitted on a CPU-only node. Off by default, because a partitioned board stops publishing `nvidia.com/gpu`.

## Why

Closes https://github.com/NVIDIA/k8s-test-infra/issues/241

`nvml-mock` exists so GPU-dependent software can be exercised on CPU nodes. MIG was the largest remaining hole: every lifecycle call returned `NVML_ERROR_NOT_SUPPORTED`, so nothing that partitions a GPU could be run against the mock at all.

That is a large class of software to leave untestable. MIG is how expensive boards get shared in a multi-tenant cluster, so a lot of the ecosystem is built around it — the GPU Operator's MIG manager, GFD, the device plugin's MIG strategies, schedulers and autoscalers that reason about slices, and the admission and capacity policies layered on top. Partitioning also changes what a node advertises, from `nvidia.com/gpu` to per-profile resources, so this is not a detail those consumers can be indifferent to; it is the shape of the node they plan against.

Without it, testing any of that needs real A100/H100/B200 hardware held for the purpose — exactly the cost `nvml-mock` exists to remove — and it keeps the whole partitioned configuration out of CI, so its bugs surface only on the hardware that is hardest to get time on.

### Scope boundary

A repartition made at runtime changes the **NVML view** only. The device plugin cannot allocate what it produces until the pod restarts, so the layout a node boots with is the layout that is allocatable. That is deliberate for v1 and documented in `docs/nvml-mock-ctl.md`.

Repartitioning is driven through `nvidia-smi` alone. `nvml-mock-ctl` gains no MIG command: it would be a second, non-standard spelling of what the real interface already covers, and a second writer of the same state. `nvml-mock-ctl status` still reports the layout and `reset` still clears it.

## Checklist

## Verification

- Unit suites across the engine, bridge, `mockctl`, `cmd/nvml-mock-ctl`, `internal/agent` and the e2e helper packages; `golangci-lint` reports 0 issues.
- Helm unit tests: 183 passed, including new cases for the MIG values.
- The containerized ABI suite (`tests/mocknvml`), which dlopens the built `libnvidia-ml.so` and calls it as a C caller would. It includes a round-trip leg that mutates in one process and reads the result back in another, declaring its own types so it cannot pass by agreeing with the writer's.
- The MIG e2e, which now supplies its own layout at install rather than relying on a profile's. Verified locally as a Ginkgo dry-run over `h100`, `b200` and `gb300`; the CI run on this commit is what exercises it against a live cluster, on `h100` and `b200`, one board per architecture.


ArangoGutierrez (commented):
Thanks for this, recording each MIG mutation back into the override document is what makes a partition outlive the nvidia-smi that created it, and that is the right shape for a per-process engine.

This takes MIG from stubbed to a working lifecycle over the C ABI (mode toggle, GPU and compute instance create and destroy), adds a declared-layout form applied by difference so a running consumer follows a repartition, and pins an a100/h100 e2e leg with the device plugin in migStrategy=single. Two things want another pass before merge: the chart-side gpu.mig gate, and handle retirement when MIG mode is switched back off.

- Against the two MIG profiles the chart ships this line cannot run as written. a100 and h100 both declare a full seven-instance layout, so `-mig 1` records mode only, the reconcile materialises all seven from the profile, and the create then comes back ERROR_INSUFFICIENT_RESOURCES with remaining capacity at 0. A `mig -dgi` step ahead of it, or dropping the line, makes the sequence runnable. (docs/nvml-mock-ctl.md:374)

deployments/nvml-mock/helm/nvml-mock/templates/_helpers.tpl
```diff
@@ -168,6 +169,24 @@ preserves comments and key order from the profile file.
...
 {{- end -}}
+{{- if $migEnabled -}}
+{{- /*
+Turn on the partitioning the profile already declares, rather than describing
+a layout here. A profile that declares none would boot MIG-enabled and
+unpartitioned, which publishes no GPU resource at all under
+migStrategy=single, so that is refused rather than rendered.
+*/ -}}
+{{- $mig := get $defaults "mig" | default (dict) -}}
+{{- if not (get $mig "gpu_instances") -}}
```
ArangoGutierrez, line 192: This check runs before .Values.gpu.mig.gpuInstances is spliced into $mig, so an explicit layout can never satisfy it. On gb300 (the chart default), gb200 and b200, all MIG-capable at max_gpu_instances: 7 but shipping no default layout, `--set gpu.mig.enabled=true` together with a complete gpuInstances value aborts the render blaming the profile, while a100 and h100 render the same flags fine. Moving the check below the override, or widening it to `or (get $mig "gpu_instances") .Values.gpu.mig.gpuInstances`, restores the path values.yaml documents gpuInstances for.

deployments/nvml-mock/helm/nvml-mock/tests/configmap_test.yaml
```diff
@@ -535,3 +535,78 @@ tests:
...
+      gpu:
+        profile: a100
+        mig:
+          enabled: true
+    asserts:
+      # The flag is Helm's, not the engine's schema; leaking it would make the
+      # mock warn about an unknown key.
+      - notMatchRegex:
+          path: data["config.yaml"]
+          pattern: "^[[:space:]]*enabled: true"
```
ArangoGutierrez: helm-unittest compiles this with Go's regexp and no (?m), so `^` anchors to the start of the whole config.yaml string and the pattern can never match an interior line. Setting `enabled` on $mig in the template leaks `enabled: true` straight into the rendered mig block and all 43 tests stay green, so the case this guards is the one it cannot catch. `(?m)` on its own overmatches the infiniband block's own `enabled: true`; `(?m)^ {4}enabled: true` goes red on the leak and stays green on the current template.

pkg/gpu/mocknvml/engine/mig.go
```diff
@@ -0,0 +1,1258 @@
...
+	}
+	if mode != nvml.DEVICE_MIG_ENABLE && mode != nvml.DEVICE_MIG_DISABLE {
+		return nvml.ERROR_INVALID_ARGUMENT, nvml.ERROR_INVALID_ARGUMENT
+	}
+
+	st.mu.Lock()
+	defer st.mu.Unlock()
+
+	if mode != st.mode {
+		st.destroyAllLocked(d)
```
ArangoGutierrez: SetMigMode reaches destroyAllLocked without collecting st.devices first, so onRepartition never learns which partitions went away and retireMigDeviceHandles never runs. After nvmlDeviceSetMigMode(dev, 0) a MIG device handle the engine already handed out still resolves, still reports IsMigDeviceHandle, and still answers GetUUID with the destroyed partition's UUID; the same handle after GpuInstanceDestroy correctly stops resolving. A later reconcile cannot repair it either, since destroyAllLocked has already emptied st.devices. Collecting st.devices before the teardown and passing them to onRepartition, the way rebuildMIG does, closes it.

pkg/gpu/mocknvml/engine/mig.go
```diff
@@ -0,0 +1,1258 @@
...
+		// to a shape we control, keyed on the parent text so distinct parents
+		// stay distinct.
+		h := fnv.New64a()
+		_, _ = h.Write([]byte(base))
+		sum := h.Sum64()
+		return fmt.Sprintf("MIG-%08x-%04x-%04x-%02x%02x-%012x",
+			uint32(sum>>32), uint16(sum>>16), uint16(sum), byte(giID), byte(ciID), sum&0xffffffffffff)
+	}
+
+	groups[3] = fmt.Sprintf("%02x%02x", byte(giID), byte(ciID))
```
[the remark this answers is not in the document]
giuliocalzo: Confirmed and fixed in 158822a6, though by rejecting the config rather than changing the derivation.

Two devices differing only in the spliced group do produce identical MIG UUIDs, and a consumer resolving one through `nvmlDeviceGetHandleByUUID` gets whichever was found first. The shipped profiles are unaffected — they vary the last group, which the derivation preserves — so this is reachable only through an operator-written profile, but the fourth group is a natural place to hand-write an index.

Preserving full parent identity in the derivation would change every MIG UUID and churn the fixtures, snapshots and e2e goldens, for a collision the config can be checked for directly. `validateDeviceUUIDs` now rejects the profile at load, matching how `config.go` already handles two devices colliding on a derived minor number. The collision is detected by asking `migDeviceUUID` itself for one fixed pair of instance ids rather than restating which group it writes: the id contribution is identical for both parents, so agreeing there is agreeing for every partition either board can carry. The blunter case it generalizes — two devices given one UUID — is reported separately, since that collides as full GPUs before MIG enters into it.

All 14 shipped profiles are now run through `validateYAMLConfig` in the test suite, since these checks reject a config on derived properties and tightening one could otherwise turn a shipped profile into a pod that will not start.

pkg/gpu/mocknvml/bridge/mig_persist.go
```diff
@@ -0,0 +1,301 @@
...
+// engine's handle for it. A nil placement asks the engine to choose one.
+func migCreateGpuInstance(
+	deviceHandle unsafe.Pointer, profileID int, placement *nvml.GpuInstancePlacement,
+) (unsafe.Pointer, nvml.Return) {
+	const what = "GPU instance create"
+	dev := engine.GetEngine().LookupConfigurableDevice(deviceHandle)
+	if dev == nil {
+		return nil, nvml.ERROR_INVALID_ARGUMENT
+	}
+	path, index, ret := migPersistBaseline(dev, what)
```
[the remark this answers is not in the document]
giuliocalzo: Confirmed and fixed in 27f9eb48, by holding one lock across the whole span as suggested.

The deadlock hazard that would normally rule that out is not there: the engine reads the override document without taking the lock at all, relying on atomic rename, so nothing re-enters it during the mutation. The only obstacle was that the `mockctl` MIG writers each acquired the lock themselves.

`MIGTx` is now that single hold, with a method per recording function. The one-shot functions remain, expressed over the same deltas, for the mutations that stand alone. Worth noting this also restores the discipline the CLI already keeps, where one hold spans load, apply and write — the bridge was the only path that split it.

Two side effects of committing once instead of writing per step: the seed and the delta land together, closing the window where a seed was recorded and the delta meant to edit it then failed, leaving a device claiming an explicit layout it was never given; and a mode change gets a transaction too, without a seed, since it owns the whole layout rather than editing it.

Covered by `TestMIGTx_SpansTheChoiceOfAnIDAndItsRecord`, which has eight writers each draw the next free id from the document and record it. With the lock removed from `BeginMIG` it fails as described — all eight choose id 0 and one instance lands.

deployments/nvml-mock/helm/nvml-mock/templates/_helpers.tpl
giuliocalzo: Confirmed and fixed in 68076fb1. Reproduced exactly as described: `--set gpu.profile=gb300 --set gpu.mig.enabled=true` with a complete `gpuInstances` aborted blaming the profile. The override is applied before the check now.

I did not take the widened condition, because it would render MIG onto a t4. `t4` and `l40s` set `max_gpu_instances: 0` — no layout supplied here makes those boards capable — while `gb300`, `gb200` and `b200` are capable and merely ship no default layout. Those are different mistakes, so there are now two refusals: one for a board that cannot partition at all, one for a capable board left with no layout from either source.

New cases cover gb300 rendering with `gpuInstances`, gb300 still refused without one, and t4 still refused *with* one. `values.yaml` is updated too — its comment described the old behaviour.

deployments/nvml-mock/helm/nvml-mock/tests/configmap_test.yaml
giuliocalzo: Confirmed and fixed in 68076fb1, with the exact pattern suggested.

Verified the assertion was inert before changing it: setting `enabled` on `$mig` in the template leaked `enabled: true` into the rendered mig block and all 175 tests stayed green. Checked the indentation inside the `data["config.yaml"]` string rather than the manifest — mig keys sit at four spaces, infiniband's own `enabled: true` at two — so `(?m)^ {4}enabled: true` separates them. Re-injecting the leak now turns the assertion red.

pkg/gpu/mocknvml/engine/mig.go
giuliocalzo: Confirmed and fixed in 708a900c. A test asserting the retirement reported zero devices where seven were destroyed, so the two teardown paths did disagree exactly as described.

Rather than collecting `st.devices` at the call site the way `rebuildMIG` does, `destroyAllLocked` now returns what it destroyed. The bug was precisely a caller not collecting first, and once the table is empty there is no second chance — so the teardown hands the devices back instead of trusting each caller to remember. `reconcileMIG`'s own hand-rolled collection goes with it.

The hook is invoked after `st.mu` is released, following `reconcileMIG`'s ordering.

internal/agent/migcaps/migcaps.go
```diff
@@ -0,0 +1,114 @@
...
+
+func removeSurface(h *host.Host) error {
+	gpuDirs, err := filepath.Glob(filepath.Join(h.Root, capabilitiesDir, "gpu*"))
+	if err != nil {
+		return fmt.Errorf("scan %s: %w", capabilitiesDir, err)
+	}
+
+	paths := make([]string, 0, 3+len(gpuDirs))
+	paths = append(paths,
+		filepath.Join(h.Root, capDevDir),
```
roma-glushko: h has RootPath() sugar method to use here and everywhere.

deployments/nvml-mock/helm/nvml-mock/profiles/a100.yaml
```diff
@@ -289,11 +289,17 @@ device_defaults:
...
+  # The layout below is what nvidia-mig-parted's all-1g.5gb config produces,
+  # declared but inert: it only takes effect while mode_current is "enabled".
+  # MIG ships off because a partitioned board stops publishing nvidia.com/gpu
+  # under migStrategy=single. Turn it on with --set gpu.mig.enabled=true.
   mig:
     mode_current: "disabled"
     mode_pending: "disabled"
     max_gpu_instances: 7            # A100 supports up to 7 MIG instances
-    # GPU instances would be defined here if MIG enabled
+    gpu_instances:
```
roma-glushko: What this configuration is meant to represent? The active MIG profile is a purely runtime value, while profile is a device specification.

The specification then should represent all supported profiles and their characteristics like described in https://docs.nvidia.com/datacenter/tesla/mig-user-guide/latest/supported-mig-profiles.html#a100-mig-profiles

Then you should be able to set a specific profile in runtime either via ctl or CRDs/control plane.

deployments/nvml-mock/helm/nvml-mock/templates/_helpers.tpl
```diff
@@ -168,6 +169,33 @@ preserves comments and key order from the profile file.
...
+first would refuse the very case gpuInstances documents itself for.
+
+Two refusals, because they are different mistakes. A board whose profile sets
+max_gpu_instances to 0 cannot partition at all, and no layout supplied here
+changes that. A capable board with no layout from either source would boot
+MIG-enabled and unpartitioned, which publishes no GPU resource whatsoever
+under migStrategy=single.
+*/ -}}
+{{- $mig := get $defaults "mig" | default (dict) -}}
+{{- if .Values.gpu.mig.gpuInstances -}}
```
roma-glushko, line 186: why do we have to enable it via chart values? Why ctl/CRDs would not be enough?

roma-glushko (commented):
What does this PR mean for GPU Operator migManager container which is disabled in our tests right now (to the best of my knowledge)? 

Will it be possible to enable it and expect to work?

tests/mocknvml/main.go
```diff
@@ -234,10 +237,11 @@ func checkProcessUtilization() {
 	if ret != nvml.SUCCESS {
 		log.Fatalf("checkProcessUtilization: DeviceGetHandleByIndex(1): %v", nvml.ErrorString(ret))
 	}
-	if s1, ret := d1.GetProcessUtilization(0); ret != nvml.SUCCESS || len(s1) != 0 {
-		log.Fatalf("checkProcessUtilization: device 1 (processes: []) -> ret=%v, %d samples; want SUCCESS, 0", nvml.ErrorString(ret), len(s1))
+	if s1, ret := d1.GetProcessUtilization(0); ret != nvml.ERROR_NOT_FOUND || len(s1) != 0 {
+		log.Fatalf("checkProcessUtilization: device 1 (processes: []) -> ret=%v, %d samples; want NOT_FOUND, 0",
```
roma-glushko: We should use testify/require in the tests btw.

internal/agent/migcaps/migcaps.go
giuliocalzo: Good catch — `RootPath`'s own doc comment asks callers to prefer it over `filepath.Join(h.Root, ...)` ("so a call site names the namespace it writes to"), and this package was ignoring that at every site.

Fixed in fdade776: all 7 non-test call sites in `migcaps.go` and `stage.go`, plus the 12 in `migcaps_test.go`.

I scoped it to the code this PR adds rather than the whole tree — there are ~35 other direct `h.Root` uses, mostly pre-existing and in tests, and sweeping them here would bury the MIG change. Happy to do that as a follow-up if you want it.

deployments/nvml-mock/helm/nvml-mock/profiles/a100.yaml
giuliocalzo: Agreed on the principle, and I think that separation is already there — the supported-profile catalog is not in this YAML.

`resolveMIGProfiles` derives it in code from go-nvml's per-board tables, matched on device name and memory size, and the comment there is explicit about why: "go-nvml owns the tables themselves; duplicating them here would mean two sources of truth for slice counts and placements that must agree."

So `nvidia-smi mig -lgip` publishes the full board catalog with its real characteristics regardless of what this file declares. From the a100 leg of the e2e:

```
Name         ID   Instances   Memory
1g.5gb       19   0/7         4.75 GB
1g.5gb+me    20   0/1         4.75 GB
2g.10gb      14   0/3         9.62 GB
...
```

including the `+me` variants, the per-profile engine counts, and the hardware profile ids (19/20/14) rather than NVML enum ordinals.

What `gpu_instances` sets here is only the **initial** runtime layout — which partitions exist at boot — and it is inert until `mode_current` is enabled. It exists because a node must be able to come up already partitioned before any consumer starts; see my reply on `_helpers.tpl` for why that can't be deferred to runtime. Repartitioning at runtime goes through `nvidia-smi` (or ctl) and is reconciled by difference, so a running consumer follows it.

If the objection is to the naming rather than the split — `gpu_instances` reading like a device spec when it's initial state — I agree that's confusing and I'm happy to rename it.

deployments/nvml-mock/helm/nvml-mock/templates/_helpers.tpl
giuliocalzo: Because this sets **install-time** state, which ctl/CRDs can't reach: under `migStrategy=single` a partitioned board stops publishing `nvidia.com/gpu`, so the node has to come up already partitioned before the device plugin starts. Enabling MIG afterwards means the plugin has already advertised the wrong resource and has to be restarted to pick up the change.

Runtime changes do go through `nvidia-smi`/ctl and are reconciled by difference — this gate is only about initial state.

It also fails fast on two mistakes that are otherwise silent, which is what the two `fail` branches below are for: a board that can't partition at all (`max_gpu_instances: 0`), and a MIG-capable board with no layout from either the profile or `gpuInstances`, which would boot MIG-enabled and unpartitioned and publish no GPU resource whatsoever.

If the control plane grows a MIG CRD, I'd expect this gate to become the bootstrap path for it rather than a competing one.

tests/mocknvml/main.go
giuliocalzo: Agreed as a convention — `AGENTS.md` does ask for testify/require — but it can't apply to this file as it stands: `tests/mocknvml` is `package main` in its own module, with no `*_test.go` files and no testify dependency. It's a standalone binary that runs inside a container against the mock `libnvidia-ml.so` and asserts with `log.Fatalf`, because the whole point is to exercise the real C ABI through the dynamic linker rather than the Go engine directly.

`require` needs a `*testing.T`, so adopting it means converting the entire harness (113 checks) to `go test`. That's worth doing and I'd support it, but it predates this PR and would swamp the MIG change under an unrelated refactor.

Want me to file it as a follow-up issue?

giuliocalzo:
> What does this PR mean for GPU Operator migManager container which is disabled in our tests right now? Will it be possible to enable it and expect to work?

This PR doesn't change migManager's status — it stays `enabled: false` in both `tests/e2e/gpu-operator-values.yaml` and `local/gpu-operator/gpu-operator.values.yaml`, and no e2e leg covers it.

It does make enabling it plausible for the first time. migManager drives `nvidia-mig-parted`, and the reconcile path here is deliberately shaped after `nvidia-mig-parted apply`: apply a declared layout by difference, so a running consumer follows a repartition rather than needing a restart. The C ABI surface migManager needs — mode toggle, GPU/compute instance create and destroy, possible placements, remaining capacity — is what this PR implements, and it's now exercised over the real ABI (`nvidia-smi mig -cgi/-cci/-dgi/-dci`, by profile name and by hardware profile id) in an a100/h100 e2e leg.

But I haven't tried it, so I won't claim it works. migManager also does node labeling, cordon/drain and device-plugin restarts, and any of those could need work beyond NVML. I'd rather enable it in a follow-up with its own e2e leg than assert it here and have it be wrong.

roma-glushko (changes requested):
@giuliocalzo This is a substantial update. In order to review it efficiently I would like to understand:
- the full lifecycle of mig partioning in simple words. 
- did we consider any other approaches to simulate mig?
- how ecosystem uses MIG to see whether we need any other work after this one to get general compatibility. Just like with chroot nvidia-smi usage, it's not always straightforward.
- Any followup work needed here? If so where is the epic so when we finish it we can largely call it a day?

pkg/gpu/mocknvml/engine/mig.go
```diff
@@ -0,0 +1,1301 @@
...
+		retired = st.destroyAllLocked(d)
+	}
+	st.mode = mode
+	st.pending = mode
+	st.mu.Unlock()
+
+	// Outside the lock, as in reconcileMIG: the hook is the engine's, and what
+	// it does with the devices is not this lock's business.
+	if len(retired) > 0 && d.onRepartition != nil {
+		d.onRepartition(retired)
```
ArangoGutierrez, line 460: The retirement here covers MIG device handles but not the instance handles. `onRepartition` lands in `retireMigDeviceHandles` (mig_engine.go:565), which only does `e.handles.Retire`, while `GpuInstanceDestroy` additionally retires `e.computeInstances` and `e.gpuInstances`.

So on a 7x1g A100, after `nvmlDeviceSetMigMode(dev, 0)`, the MIG device handle correctly answers `ERROR_NOT_FOUND`, but a GPU instance handle and a compute instance handle both still resolve `SUCCESS`. The same shape is in `rebuildMIG` (mig_reconcile.go) and in `reconcileLiveInstance`.

Collecting the instances alongside `st.devices` and retiring all three registries would close it.

pkg/gpu/mocknvml/engine/mig_profiles.go
```diff
@@ -0,0 +1,298 @@
...
+		nvml.GPU_INSTANCE_PROFILE_2_SLICE:      14,
+		nvml.GPU_INSTANCE_PROFILE_3_SLICE:      9,
+		nvml.GPU_INSTANCE_PROFILE_4_SLICE:      5,
+		nvml.GPU_INSTANCE_PROFILE_7_SLICE:      0,
+	}
+
+	// fourSliceProfileIDs covers the A30, whose slices are twice an A100's, so
+	// each of its profiles takes the ID of the A100 profile holding the same
+	// fraction of its board.
+	fourSliceProfileIDs = migProfileIDs{
```
ArangoGutierrez: This table omits the two +me variants the A30's go-nvml table carries, `1_SLICE_REV1` (enum 7) and `2_SLICE_REV1` (enum 8). `reported()` falls through to the enum for anything unmapped, so those profiles enumerate with Id 7 and 8, but `enumOf(7)` and `enumOf(8)` return false, and `CreateGpuInstance` along with the placement and capacity calls then refuse a profile the board just advertised. `sevenSliceProfileIDs` handles the equivalent case with 20 and 15, so it is only this table.

The reason it is not caught is next door: the a30 case in mig_profiles_test.go pins three of the five profiles that board publishes, so 7 and 8 fall through unpinned.

No a30.yaml ships today, so this reaches a user-authored A30 profile rather than a shipped one, since `resolveMIGProfiles` matches on device name.

pkg/gpu/mocknvml/engine/mig_profiles_test.go
```diff
@@ -0,0 +1,234 @@
...
+		name        string
+		deviceName  string
+		memoryBytes uint64
+	}{
+		{"a100 40gb", "NVIDIA A100-SXM4-40GB", a100_40GiB},
+		{"a100 80gb", "NVIDIA A100-SXM4-80GB", 2 * a100_40GiB},
+		{"a30", "NVIDIA A30", 25769803776},
+		{"h100", "NVIDIA H100 80GB HBM3", h100_80GiB},
+		{"h200", "NVIDIA H200 141GB HBM3e", 151397302272},
+		{"b200", "NVIDIA B200 180GB HBM3e", 193273528320},
```
ArangoGutierrez: The b200 row of this test cannot fail. b200 resolves to a nil ID table, so `reported(enum) == enum` and the uniqueness claim reduces to "map keys are distinct", true by construction rather than by anything about the board. Injecting a collision into `sevenSliceProfileIDs` (2_SLICE 14 -> 19) turns a100_40gb, a100_80gb, h100 and h200 red while b200 stays green.

The a30 row is not affected by this: it is the genuinely partly-mapped board the test's own comment describes, and it earns its place. Dropping the b200 row, or marking it a placeholder, would make the case say what it means.

deployments/nvml-mock/helm/nvml-mock/values.schema.json
```diff
@@ -110,6 +110,38 @@
...
+                "type": "object",
+                "additionalProperties": true,
+                "properties": {
+                  "profile": {
+                    "type": "string",
+                    "description": "Canonical MIG profile name, e.g. \"1g.5gb\". Mutually exclusive with profile_id."
+                  },
+                  "profile_id": {
+                    "type": "integer",
+                    "description": "Raw NVML GPU instance profile ID, for a partition with no canonical name.",
```
ArangoGutierrez: This says NVML GPU instance profile ID, but configuration.md added in this same PR says the opposite: `profile_id` is the ID column from `nvidia-smi mig -lgip`, and explicitly not NVML's enum, which numbers the same profiles in the opposite order. The engine agrees with the docs, via `enumOf`.

Worth matching the wording here, because the two numberings overlap in a way that fails quietly: `profile_id` 0 is `7_SLICE` on an A100, so copying NVML's enum 0 for `1_SLICE` carves the whole board into one instance instead of a seventh of it.

deployments/nvml-mock/helm/nvml-mock/templates/_helpers.tpl
```diff
@@ -168,6 +169,33 @@ preserves comments and key order from the profile file.
...
+migStrategy=single.
+*/ -}}
+{{- $mig := get $defaults "mig" | default (dict) -}}
+{{- if .Values.gpu.mig.gpuInstances -}}
+{{- $_ := set $mig "gpu_instances" .Values.gpu.mig.gpuInstances -}}
+{{- end -}}
+{{- if not (get $mig "max_gpu_instances") -}}
+{{- fail (printf "gpu.mig.enabled is set but profile %q is not a MIG-capable board: it declares no mig.max_gpu_instances" .Values.gpu.profile) -}}
+{{- end -}}
+{{- if not (get $mig "gpu_instances") -}}
```
ArangoGutierrez: This checks the list is non-empty rather than that it describes any partitions, so a zero count slips past:

```
helm template mig-zero deployments/nvml-mock/helm/nvml-mock \
  --set gpu.profile=a100 --set gpu.mig.enabled=true \
  --set 'gpu.mig.gpuInstances[0].profile=1g.5gb' \
  --set 'gpu.mig.gpuInstances[0].count=0'
```

returns rc=0 and renders `mode_current: enabled` with `gpu_instances: [{count: 0, profile: 1g.5gb}]`, which is the state this message exists to prevent. values.schema.json allows it too, with `count` minimum 0. Summing the counts rather than measuring the list would cover it.

ArangoGutierrez (commented):
All four points from the last round are fixed. I checked them against the tree rather than against the replies, since the branch was rewritten in between and the commits named in those replies no longer exist: the guard now sits below the gpuInstances splice, the notMatchRegex patterns carry `(?m)`, SetMigMode retires through onRepartition, and dropping the shipped layouts makes the create sequence in nvml-mock-ctl.md runnable. Naming the layout at install instead of shipping one is the right call, and it answers the profile-versus-runtime-state question too.

The new guards hold up under mutation. I broke seven of them one at a time, the profile-ID table, the SetMigMode retirement, both MIGTx transaction guards, b200's max_gpu_instances, the no-shipped-layout guard and the chart-layout resolution guard, and every one turned a test red. The concurrency guard killed its mutant on three runs out of three. That is not the usual outcome for a change this size.

Five things inline. None of them argue with the shape of the design.

Two smaller notes. The two new exported Engine methods and `migStateOf` have no Go-level test; only the Docker harness and the e2e reach them. And `testMIGProfileIDs` is gated on `deviceCount > 4` without emitting a skip record, so if a fixture ever shrinks, the leg would drop out of the tally rather than be reported as skipped. The gate is open today, with 8 devices.

Two notes that are not code. Issue #241's acceptance criteria now disagree with the PR in two places: it asks for MIG partition definitions in the A100 and H100 profiles, which fe0cb394 deliberately removed, and it says state is in-memory and not persisted, which mig_persist.go changes. Both read as improvements rather than mistakes, so it is probably the issue that wants updating. And on Roma's question about the epic: MIG is #241, under #254, the NVML API coverage tracking issue.

Separately, docs/helm-chart.md's chart value reference carries no `gpu.mig.*` rows, and mig.md points readers there for every chart value. That predates this PR, but this PR is what makes those values the primary MIG surface.

pkg/gpu/mocknvml/engine/mig.go
giuliocalzo: Fixed in 48de721a.

The hook could not carry it: `onRepartition` was typed `func([]*ConfigurableDevice)`, so the instances a teardown destroyed had nowhere to go, and `destroyAllLocked` wiped `Device.GpuInstances` without ever naming them. It now reports a `migRetired` — devices, GPU instances, compute instances — and `retireRepartitioned` clears all three registries, compute instances before their parents, the order `GpuInstanceDestroy` uses.

Carried through all three paths you named: `SetMigMode`, `rebuildMIG`, and `reconcileLiveInstance` plus `reconcileComputeInstances` beneath it.

One case deriving the instances from `st.devices` would have missed, so it is worth stating: a GPU instance holding no compute instance has no MIG device pointing back at it — `nvidia-smi mig -cgi` without `-C` — so it has to be collected from the instance tree instead. `TestSetMigMode_RetiresABareInstanceHandle` covers that; `TestSetMigMode_RetiresInstanceHandles` covers the case you reported. Both fail on the old code with SUCCESS where NVML returns ERROR_INVALID_ARGUMENT.

pkg/gpu/mocknvml/engine/mig_profiles.go
giuliocalzo: Fixed in 77f1442a.

Confirmed against the go-nvml table: the A30 publishes five profiles, enums 0, 1, 3, 7 and 8, and `fourSliceProfileIDs` mapped three. So 7 and 8 enumerated through `reported()`'s fallback and were refused by `enumOf`, exactly as you describe.

Their real IDs are 21 and 6, from A30 `-lgip` listings. Worth noting they do not follow from the rule the rest of that table used: an A100 lists its `+me` profiles as 20 and 15, so no narrowing of the 7-slice numbering produces 21 and 6. The comment asserting that rule was wrong for those two rows and is corrected.

On the test next door — pinning IDs one at a time cannot catch this, because a test written from a partial table forgets the same profiles the table does. `TestResolveMIGProfiles_EveryPublishedProfileResolvesBack` enumerates the board instead and asserts the invariant that actually matters, that a board accepts every ID it advertises. It fails on the old table with `profile enum 7 is advertised under ID 7, which must be accepted back`.

deployments/nvml-mock/helm/nvml-mock/templates/_helpers.tpl
giuliocalzo: Fixed in e7dee2ce. Your repro returns rc=1 now with the `declares no partitions` message.

It sums the counts, as you suggested. Two details it settles: an entry with no `count` is one instance, which is how the engine reads it, so a layout of bare profile names still renders; and a zero alongside real partitions still renders, since that asks for none of that profile rather than for nothing.

I left `count` at minimum 0 in the schema rather than raising it to 1 — a zero entry is harmless once the total is what decides, and the guard now gives a better message than a schema violation would. Its description says so.

Covered by two cases in `configmap_test.yaml`: the zero-count layout is refused, and the entry omitting `count` is accepted.

deployments/nvml-mock/helm/nvml-mock/values.schema.json
giuliocalzo: Fixed in c98a187e. The docs and the engine were right; the schema was the odd one out.

It now says the ID column of `nvidia-smi mig -lgip`, not NVML's enum, and keeps your example of why it matters — ID 0 is the whole board on an A100 where the enum 0 is a seventh of it — because that is the part that fails silently rather than erroring.

pkg/gpu/mocknvml/engine/mig_profiles_test.go
giuliocalzo: Fixed in 77f1442a — dropped the b200 row.

Reproduced your falsification first: with 2_SLICE remapped 14 -> 19, a100_40gb, a100_80gb, h100 and h200 go red and b200 stays green. A board with no ID table cannot collide, so the row asserted nothing.

What is worth pinning about those boards is that they report the enum at all, which `TestResolveMIGProfiles_UnverifiedBoardsReportTheEnum` already does, so nothing is lost. The test comment now says why only mapped boards are listed.

The a30 row stays, though it no longer has the character your comment describes: it is fully mapped as of this push, so it now guards `fourSliceProfileIDs` against a collision the same way the others guard the 7-slice table.

pkg/gpu/mocknvml/engine/mig_engine_test.go
```diff
@@ -0,0 +1,178 @@
...
+func TestComputeInstanceHandleLifecycle(t *testing.T) {
+	t.Parallel()
+
+	e := newFullyPartitionedEngine(t, 1)
+
+	deviceHandle, ret := e.DeviceGetHandleByIndex(0)
+	require.Equal(t, nvml.SUCCESS, ret)
+	instances, ret := e.DeviceGetGpuInstances(deviceHandle, nvml.GPU_INSTANCE_PROFILE_1_SLICE)
+	require.Equal(t, nvml.SUCCESS, ret)
+	giHandle := instances[0]
```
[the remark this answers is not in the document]
giuliocalzo: This collection has a specified order, so the rule does not apply here, but it was relied on without being stated — so I have pinned it in 48de721a.

`DeviceGetGpuInstances` reads through `liveGpuInstances`, which ends `sort.Slice(gis, func(i, j int) bool { return gis[i].Info.Id < gis[j].Info.Id })` and is documented "snapshots the GPU instances of a device, ID-ordered". `GpuInstanceGetComputeInstances` reads through `liveComputeInstances`, which sorts the same way. The sort is not for the tests: the reconciler needs a deterministic teardown order, which is why `reconcileExplicitInstances` notes that its snapshot is ID-ordered.

So `instances[0]` is the lowest-ID instance on every run, not an arbitrary one. `TestGpuInstanceHandleLifecycle` now asserts the listing is increasing in instance ID before indexing it, which makes the guarantee explicit and fails if the sort is ever dropped — reversing the comparator turns it red.

tests/e2e/go/scenario_mig_test.go
```diff
@@ -0,0 +1,1033 @@
...
+				Expect(migComputeInstanceCountOnNode(ctx, h, node)).To(Equal(baseCIs+2),
+					"-C should have given each new partition a compute instance")
+				Expect(migDeviceCountOfGPU(ctx, h, node, 0)).To(Equal(2),
+					"nvidia-smi -L should show a MIG device for each of GPU 0's two partitions")
+
+				By("nvidia-smi mig -dci -gi <id> -ci 0 && -dgi -gi <id>: delete one partition, by id")
+				// The first instance's id, read from the listing above: NVML
+				// chooses these, so a spec that assumed 0 and 1 would be
+				// asserting against the allocator rather than the delete.
+				doomed, spared := created[0].InstanceID, created[1].InstanceID
```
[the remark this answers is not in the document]
giuliocalzo: This one already selects by ID rather than by position, and the assertion is order-independent, so there is nothing to change.

`created[0].InstanceID` reads an ID *out of* the listing and every mutation afterwards names that ID: `-dci -gi <doomed> -ci 0`, then `-dgi -gi <doomed>`. The comment above it makes that the point — NVML chooses the IDs, so a spec assuming 0 and 1 would be testing the allocator rather than the delete.

On the survivor assertion: the two partitions are created from the same profile and are interchangeable, so the pair is symmetric. If the listing came back the other way round, `doomed` and `spared` swap together and the assertion still holds — exactly one instance is deleted and the survivor is the one `-dgi` did not name. There is no listing order that deletes one instance and expects the other to be gone.

pkg/gpu/mocknvml/engine/mig_explicit.go
```diff
@@ -0,0 +1,255 @@
...
+func (d *ConfigurableDevice) MIGLayoutRecords() []MIGGPUInstanceRecord {
+	st, ret := d.migEnabled()
+	if ret != nvml.SUCCESS {
+		return nil
+	}
+
+	instances := st.liveGpuInstances(d)
+	records := make([]MIGGPUInstanceRecord, 0, len(instances))
+	for _, gi := range instances {
+		profileID := int(gi.Info.ProfileId)
```
ArangoGutierrez: `MIGLayoutRecords` records `gi.Info.ProfileId`, which is NVML's profile enum, but every reader of `instances[].profile_id` resolves it through `migProfileIDs.enumOf` as the ID the board publishes, and `migCreateGpuInstance` converts with `DeviceGpuInstanceReportedProfileID` before recording a create. On an A100 a 1g instance carries enum 0, and 0 is the published ID of `7_SLICE`, so the baseline `migBeginDelta` seeds on the first runtime mutation replays a 7x1g board as one 7g instance, the other six failing ERROR_INSUFFICIENT_RESOURCES. Recording `st.profileIDs.reported(int(gi.Info.ProfileId))` here matches the create path and the round trip holds. `TestMIGLayoutRecords_MirrorsTheLiveBoard` pins the enum, so it moves to the published ID with it, and the round-trip test asserts only the instance count today, which is why this is not caught.

docs/configuration.md
```diff
@@ -122,6 +122,16 @@ device_defaults:
     rx_throughput_kbps: 0
 ```
 
+`lspci` sees these through the rendered sysfs tree and enumerates every GPU, but
+two of its lines are missing next to real hardware. `Kernel driver in use:
+nvidia` needs a `driver` symlink, which the tree does not render and the
+`libpcisysfs` shim could not surface anyway — it intercepts `open`/`stat`, not
```
ArangoGutierrez: The shim is `libmockfs` since the rename the CHANGELOG records; `shims/libpcisysfs` is gone.

Also at tests/e2e/go/assets/device-plugin-mock-mig.yaml:45.

ArangoGutierrez (commented):
Thanks for this, taking MIG from stubs to a create and destroy path that `nvidia-smi mig -cgi/-cci` drives unmodified is a big step for the mock.

Runtime MIG now goes end to end: every mutation is recorded in the override document, so a partition made by one `nvidia-smi` outlives that process, with `gpu.mig.enabled`, a migStrategy=single e2e leg and a guide around it. The profile-ID numbering used to seed that document is the piece to settle before this lands.

- `installMIGChart` carries two stacked doc comments. The first describes the profile's declared partitioning, which no profile ships since the layout moved to being named at install, and the second one is accurate. (tests/e2e/go/scenario_mig_test.go:703)
- No profile declares a layout now, and `installMIGChart` sets `gpu.mig.gpuInstances[0].profile` and `count` itself, so this reads backwards on both halves. h100 and b200 are the right boards to run; it is the stated reason that would mislead whoever adds the next one. (.github/workflows/nvml-mock-e2e-go.yaml:804)

pkg/gpu/mocknvml/engine/mig.go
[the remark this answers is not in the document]
ArangoGutierrez: `validateDeviceUUIDs` rejects that config at load: it keys a map on `migDeviceUUID(dev.UUID, 0, 0)` and fails the pair whose UUIDs differ only where the instance IDs are spliced in, covered by `TestValidateYAMLConfig_RejectsUUIDsThatCollapseOntoOneMIGDevice`. Devices that declare no UUID take the engine default, which varies the last group.

pkg/gpu/mocknvml/bridge/mig_persist.go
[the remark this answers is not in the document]
ArangoGutierrez: `migCreateGpuInstance` opens `migBeginDelta`, which takes the override lock through `mockctl.BeginMIG` and holds it across the probe, the seed, the engine mutation and the record. Baseline selection and ID assignment sit inside one hold, released by the transaction close.

pkg/gpu/mocknvml/engine/mig_explicit.go
giuliocalzo: Fixed in 33cb4760 — it now records `st.profileIDs.reported(int(gi.Info.ProfileId))`, as you suggested.

Your prediction reproduces verbatim. The round-trip test written to assert the profile fails on the old code with:

```
[MIG] device 0: cannot recreate instance 1 (): ERROR_INSUFFICIENT_RESOURCES
... 2 through 6 the same
Error: "[...]" should have 7 item(s), but has 1
```

One boundary I checked while confirming the asymmetry, since it decides how far the fix reaches: compute instances are not affected. `resolveDeclaredComputeInstanceProfile` returns `*cfg.ProfileID` as given, and there is no per-board remapping for compute instance profiles — `migProfileIDs` is GPU-instance only — so `computeInstanceRecords` recording the enum is symmetric with its reader and stays as it is.

`TestMIGLayoutRecords_MirrorsTheLiveBoard` moves to 19 with the code, as you expected. The round-trip test now asserts the profile rather than the instance count, which is what leaves this caught.

The function comment claimed profiles are recorded "by ID, which is the identity the instance carries" — the part that was actually wrong, since the identity it carries is the enum. It now says which of the two numberings it means and why, with your A100 example.

docs/configuration.md
giuliocalzo: Fixed in c2785b5f. Both references updated to `libmockfs.so` — `docs/configuration.md` and the e2e asset you pointed at.

`rg libpcisysfs` outside `CHANGELOG.md` now comes back empty; the CHANGELOG keeps it because that is the entry recording the rename.

deployments/nvml-mock/helm/nvml-mock/profiles/a100.yaml
```diff
@@ -289,11 +289,15 @@ device_defaults:
   # ---------------------------------------------------------------------------
   # MIG configuration
   # ---------------------------------------------------------------------------
+  # MIG ships off because a partitioned board stops publishing nvidia.com/gpu
```
roma-glushko: Either I don't understand how MIG works or this is a wrong statement. 

My understanding is the opposite. MIG profiles are mostly fixed to the GPU type (and driver version?).
Blackwell Ultra has 7 MIG profiles supported and you can't slice it in any other way but that:

https://docs.nvidia.com/datacenter/tesla/mig-user-guide/latest/supported-mig-profiles.html#b200-mig-profiles

So if we allow it to be sliced any a random way, we are not incorporating the MIG logic that we should experience in the real life.

Let me know what's your arguments against this.

deployments/nvml-mock/helm/nvml-mock/profiles/a100.yaml
```diff
@@ -289,11 +289,15 @@ device_defaults:
...
+  # under migStrategy=single. How a board is carved is a deployment choice
+  # rather than a property of the silicon, so no layout is declared here:
+  # --set gpu.mig.enabled=true needs gpu.mig.gpuInstances to supply one. Seven
+  # 1g.5gb instances is what nvidia-mig-parted's all-1g.5gb config produces.
   mig:
     mode_current: "disabled"
```
roma-glushko: I think the current state doesn't make sense to be specified in profile. It's a runtime information. It makes much more sense to explicitly state whether mig is supported by the GPU device.

deployments/nvml-mock/helm/nvml-mock/values.yaml
```diff
@@ -55,6 +55,34 @@ gpu:
...
+  # slice is named for the share of its own board it holds, so confirm a
+  # board's names with `nvidia-smi mig -lgip` rather than NVIDIA's listing
+  # for the hardware it stands in for.
+  #
+  # A board whose profile declares no mig.max_gpu_instances is refused, as is a
+  # capable board left with no layout: MIG on with nothing partitioned
+  # publishes no GPU resource at all under migStrategy=single.
+  #
+  # The instances stay mutable at runtime through NVML, as on real hardware.
+  mig:
```
roma-glushko: We have two ways to set runtime information e.g. CTL CLI + Control Plane CRDs. Do we also want to mix in helm values as a third way?
Because some of those values.yaml overrides don't make to me. I would remove them if possible. So I want to make sure this is useful and we add it consciously.

docs/guides/mig/README.md
```diff
@@ -0,0 +1,320 @@
...
+        name: nvidia-device-plugin-mock
+    spec:
+      nodeSelector:
+        mokka.nvidia.com/type: sgpu
+      tolerations:
+        - operator: Exists
+      containers:
+        - name: nvidia-device-plugin
+          image: nvcr.io/nvidia/k8s-device-plugin:v0.18.2
+          # $0 is the "--" below, so "$@" is exactly the args list.
```
roma-glushko: So MIG requires this custom device plugin daemonset? It doesn't work with the default device plugin helm chart? I want to understand whether this is a requirement in case of Mokka MIG support

docs/guides/mig/README.md
```diff
@@ -0,0 +1,320 @@
...
+`NVML_ERROR_NO_PERMISSION` rather than succeeding in one process only, which is
+what the driver reports when the same call is made without the permissions it
+needs.
+
+`nvml-mock-ctl status` shows the recorded layout and `nvml-mock-ctl reset`
+clears it. There is deliberately no `nvml-mock-ctl mig`: it would be a second,
+non-standard spelling of an interface `nvidia-smi` already covers, and a second
+writer of the same state.
+
+## How slices are named
```
roma-glushko, line 170: Need to link the official docs with the list of supported MIG profiles somewhere in this guide.

docs/guides/mig/README.md
```diff
@@ -0,0 +1,320 @@
...
+
+Take the names a board accepts from `nvidia-smi mig -lgip` rather than from the
+MIG user guide. On Blackwell, name partitions by `profile` rather than
+`profile_id`: NVIDIA publishes no profile IDs for those boards, so the mock
+reports NVML's enum instead and the ids will not match a real board's listing.
+
+## Limits
+
+A runtime repartition changes the **NVML view** only. `/dev/nvidia-caps` and
+the `mig-minors` table are staged once when the `nvml-mock` pod starts, so the
```
roma-glushko: > staged once when the `nvml-mock` pod starts

There is no pods like that anymore, they are called node-agent if I remember correctly.

docs/guides/mig/run.sh
roma-glushko, line 1: Let's not add these run.sh scripts. It's something to remove going forward.

pkg/gpu/mocknvml/engine/config.go
```diff
@@ -263,6 +270,439 @@ func validateYAMLConfig(config *YAMLConfig) error {
...
+//   - Two placements at one offset advertise a single slot twice, so a
+//     consumer enumerating them sees room for two partitions where the board
+//     has room for one.
+func validateMIGPlacements(placements []MIGPlacementSpec, maxGPUInstances int) error {
+	memoryUnits := nextPowerOfTwo(maxGPUInstances)
+	seen := make(map[uint32]struct{}, len(placements))
+	for i, p := range placements {
+		if p.Size == 0 || p.Size&(p.Size-1) != 0 {
+			return fmt.Errorf("placements[%d]: size %d is not a power of two", i, p.Size)
+		}
```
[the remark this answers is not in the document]
giuliocalzo: Confirmed and fixed in e416dea0, with the check you suggested.

The comment on `validateMIGPlacements` already asserted that every start is aligned to its size — it was stated as one of the invariants the geometry rests on, and never checked. Your repro is the case that matters: within one profile every placement is the same width, so distinct starts do not separate them, and `{1, 2}` with `{2, 2}` repeats no offset while sharing unit 2.

One detail on placement. Alignment now runs *after* the bounds test rather than before it. `{start: 6, size: 4}` on a 7-slice board breaches both, and the existing case expects to hear that it runs past the eight memory units the board has, which is the more fundamental complaint; an alignment message there would be true and less useful.

With alignment and distinct starts both enforced there is no overlap left to look for within a profile, since aligned placements of equal size either coincide or do not meet — so no separate intersection check is needed. Across profiles an overlap is expected and correct, which is what the occupancy masks are for.

Verified no shipped table changes: all 14 profiles still load, and the new case in `TestValidateMIGConfig_RejectsDeclaredGeometryThatCannotPartitionTheBoard` fails without the check.

internal/agent/source/file.go
```diff
@@ -141,32 +144,75 @@ func readTopology(path string, log *zap.Logger) ([]byte, error) {
...
+	path := engine.MIGProfilesPathFor(f.configPath)
+	if path == "" {
+		return nil
+	}
+
+	data, err := os.ReadFile(path)
+	if err != nil {
+		f.log.Debug("MIG profile table not readable; treating the board as unpartitioned",
+			zap.String("path", path), zap.Error(err))
+		return nil
```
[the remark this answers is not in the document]
giuliocalzo: Confirmed, and the consequence you describe is real — fixed in e416dea0, though one frame up from where you pointed.

`migTable` swallowing the error is not on its own destructive: `ApplyMIGProfilesOverlay` reports a `ReadFile` failure, which fails `parseProfile`, which fails `compileState`, so `poll` sends an error and returns before it ever sends state. That is what the doc comment leans on.

The hole is the line above it. `os.Stat` treated *every* error as absence, and stat is what decides presence — so a permission fault or an I/O error on the lookup degraded silently in both places at once and the withdrawal went through. It now reports anything that is not `fs.ErrNotExist`, which keeps the deliberate case (no table mounted means the board is not MIG-capable, not a load failure) and makes `migTable`'s comment true rather than half-true.

I left `migTable` as it is instead of duplicating the judgement, and pinned the contract from the outside in `TestFileSource_AnUnreadableMIGTableDoesNotWithdrawThePartitions`: a board polls partitioned, the table is replaced with a symlink loop — present to the namespace, and `ELOOP` rather than not-found — and the next poll must emit an error and no state, because a state carrying no table is precisely what makes `gpudriver` retract one. A symlink loop rather than an unreadable directory so the test does not depend on the uid it runs as.

Worth noting how it was verified, since a test written against this could easily assert nothing: removing the guard outright fails the build on the now-unused import, so both new tests were checked by neutralizing the condition instead. Both go red that way and green with it restored.

docs/guides/mig/run.sh
```diff
@@ -0,0 +1,504 @@
...
+
+# Name of a Running pod to exec into. `kubectl rollout status` returns once the
+# new pods are Ready, but a terminating old pod can still appear in the
+# listing, so filter on the phase rather than taking the first item.
+wait_for_pod() {
+  kubectl_ctx rollout status "daemonset/${RELEASE_NAME}" \
+    --timeout="$(demo::install_timeout)" >/dev/null
+  kubectl_ctx get pods -l "${POD_SELECTOR}" \
+    --field-selector=status.phase=Running \
+    -o jsonpath='{.items[0].metadata.name}'
```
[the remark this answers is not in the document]
giuliocalzo: Moot: `docs/guides/mig/run.sh` no longer exists. It was removed in d8230b4e ("docs(mig): write the MIG guide for a reader, not a runner"), and `docs/guides/mig/` is now `README.md` alone — the guide shows the commands and their output rather than shipping a script to run, so there is no pod selection or cleanup path left to get wrong. Resolving.

docs/guides/mig/run.sh
```diff
@@ -0,0 +1,504 @@
...
+
+if [ "$(smi -L | grep -c '^GPU ' || true)" -ne "${TOTAL_GPUS}" ]; then
+  fail "the board should be back to ${TOTAL_GPUS} whole GPU(s)"
+fi
+ok "the board reports ${TOTAL_GPUS} whole GPU(s) again"
+
+# Clear the runtime override so the node is left reporting the layout the
+# chart installed, rather than the empty one this demo finished on.
+info "Clearing the runtime override"
+kubectl_ctx exec "${POD}" -- nvml-mock-ctl reset | sed 's/^/    /' >&2
```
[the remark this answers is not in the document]
giuliocalzo: Moot: `docs/guides/mig/run.sh` no longer exists. It was removed in d8230b4e ("docs(mig): write the MIG guide for a reader, not a runner"), and `docs/guides/mig/` is now `README.md` alone — the guide shows the commands and their output rather than shipping a script to run, so there is no pod selection or cleanup path left to get wrong. Resolving.

deployments/nvml-mock/helm/nvml-mock/profiles/mig/b200.yaml
```diff
@@ -0,0 +1,306 @@
...
+        decoders: 1
+        jpeg: 1
+      - nvml_profile: "2_SLICE"
+        slices: 2
+        instances: 1
+        multiprocessors: 36
+        shared_copy_engines: 3
+        decoders: 1
+        jpeg: 1
+      - nvml_profile: "1_SLICE_REV1"
```
roma-glushko, line 159: what's the role of 1_SLICE_REV1 vs 1_SLICE? Does it correspond to +me extension?

deployments/nvml-mock/helm/nvml-mock/profiles/mig/gb300.yaml
```diff
@@ -0,0 +1,307 @@
...
+#
+# Every row declares its full geometry. `slices` restates the width the
+# nvml_profile enum already spans so load can reconcile the two rather
+# than trust either. `placements` are in *memory units*, not compute
+# slices: a 7-slice board has eight units, so a 1g row is seven starts of
+# size 1 and the full-board row is one `{start: 0, size: 8}`.
+# `compute_instances` is the row's `mig -lcip` listing, and its engine
+# counts are NVML's Shared* fields, which repeat the GPU instance's.
+supported_profiles:
+  - name: "1g.35gb"
```
roma-glushko, line 32: TIL that B200, GB200, GB300 have different profiles because apparently they have different amount of HBM.

deployments/nvml-mock/helm/nvml-mock/profiles/mig/b200.yaml
```diff
@@ -0,0 +1,306 @@
...
+# counts are NVML's Shared* fields, which repeat the GPU instance's.
+#
+# The names are corroborated by NVIDIA's gpu-operator default mig-parted
+# config (assets/state-mig-manager/0400_configmap.yaml), which keys B200
+# by PCI device 0x290110DE and lists 1g.23gb, 2g.45gb and 3g.90gb.
+supported_profiles:
+  - name: "1g.23gb"
+    nvml_profile: "1_SLICE"
+    slices: 1
+    profile_id: 0
```
roma-glushko, line 34: should it be called `nvml_profile_selector` to be precise? Because there is also `nvmlGpuInstanceProfileInfo.id` which device profile ID(?).

deployments/nvml-mock/helm/nvml-mock/profiles/mig/b200.yaml
```diff
@@ -0,0 +1,306 @@
...
+    instances: 7
+    memory_mb: 23040
```
roma-glushko, line 36: btw we could have a similar approach to loading byte values like standard time.Duration struct. So we could specify memory_mb: 23040 as memory: "23Gb". Would it make sense here or we need to have high precision here?

deployments/nvml-mock/helm/nvml-mock/profiles/a100.yaml
```diff
@@ -289,11 +289,15 @@ device_defaults:
   # ---------------------------------------------------------------------------
   # MIG configuration
   # ---------------------------------------------------------------------------
+  # MIG ships off because a partitioned board stops publishing nvidia.com/gpu
+  # under migStrategy=single. How a board is carved is a deployment choice
+  # rather than a property of the silicon, so no layout is declared here:
+  # --set gpu.mig.enabled=true needs gpu.mig.gpuInstances to supply one. Seven
+  # 1g.5gb instances is what nvidia-mig-parted's all-1g.5gb config produces.
   mig:
```
roma-glushko, line 297: Do we still need this section if we separated mig profiles into their own YAML files?

deployments/nvml-mock/helm/nvml-mock/profiles/gb300.yaml
```diff
@@ -91,12 +91,12 @@ device_defaults:
     pwr_object: "1.0"
 
   # ---------------------------------------------------------------------------
-  # Memory configuration - 288 GiB HBM3e per GPU (1.5x GB200's 192 GiB)
+  # Memory configuration - 278 GiB HBM3e per GPU (1.5x GB200's 186 GiB)
   # ---------------------------------------------------------------------------
   memory:
-    total_bytes: 309237645312       # 288 GiB HBM3e
+    total_bytes: 298500227072       # 278 GiB HBM3e
```
roma-glushko, line 97: Great catch 👍

deployments/nvml-mock/helm/nvml-mock/templates/_helpers.tpl
roma-glushko: @giuliocalzo this install time state is not something that lays down very well on the existing control plane idea. Let's keep in mind there is going to be one single mokka chart installation per cluster. From there the actual configuration of each node depends on Inventory -> Racks -> sGPU Nodes assigned to that K8s node. If you do something like this and set it via helm chart, then that MIG profile gets applied to the whole cluster (no matter that GPU rack type it needs to simulate). That's not right.

roma-glushko (dismissed):

deployments/nvml-mock/helm/nvml-mock/profiles/mig/b200.yaml
giuliocalzo: Yes, it's `+me`, and the repo didn't say so anywhere — fixed in 7b8698c8.

At GPU-instance level `REV1` is the media extension. Same compute slice, same framebuffer, but it carries the board's video engines, which is visible in this table two rows up from your comment: `1g.23gb` declares `decoders: 1`/`jpeg: 1` on 18 SMs and `1g.23gb+me` declares the same engines on the same 18 SMs — on `a100` the contrast is cleaner still, where plain `1g.5gb` declares no decoder at all and `1g.5gb+me` declares a decoder, a JPEG engine and an OFA. It offers `instances: 1` because a board has exactly one such set to give away. `REV2` is a different axis: one compute slice but two memory units, so it's named for the bigger framebuffer (`1g.45gb` here, `1g.10gb` on `a100`) and only four fit.

The wrinkle, and the reason line 155 specifically is confusing: that line is inside `compute_instances`, and the compute-instance enums reuse the `REV1` suffix for the same idea applied to a different object. So it's the media-capable *compute instance* shape, not a `+me` GPU instance, and it reports video engines only when the GPU instance it sits in has them to share.

`docs/mig.md` now has a [What the REV suffixes distinguish](https://github.com/NVIDIA/k8s-test-infra/blob/feat/issue-241-mig-lifecycle/docs/mig.md#what-the-rev-suffixes-distinguish) section, and the five tables point at it instead of repeating it.

deployments/nvml-mock/helm/nvml-mock/profiles/mig/gb300.yaml
giuliocalzo: Exactly right, and it's memory *only*. Normalise the names and `memory_mb` away and the three tables are identical — same widths, same `profile_id`s, same SM and engine counts, same placements:

```
$ diff <(norm b200.yaml) <(norm gb200.yaml) && echo IDENTICAL
IDENTICAL
$ diff <(norm gb200.yaml) <(norm gb300.yaml) && echo IDENTICAL
IDENTICAL
```

All three are 7-slice boards with the same geometry; what differs is 180 vs 186 vs 278 GiB of HBM, and since a MIG profile is named for the framebuffer it holds, the names move with it. (`a100` vs `b200` is a real difference by contrast — 14 SMs vs 18, one copy engine vs two, and different published ids.)

One detail that surprised me while checking: the names are not a clean function of the memory either. `b200`'s `1g` holds 22.5 GiB and `gb200`'s holds 23.25 GiB, and *both* publish as `1g.23gb` — they only diverge at the wide end, `7g.180gb` vs `7g.186gb`. Which is why the rows carry a raw `memory_mb` rather than deriving it from the name; more on that in your question a couple of comments down.

giuliocalzo: I've kept the same structure of the hardware profile to avoid problem

deployments/nvml-mock/helm/nvml-mock/profiles/mig/b200.yaml
giuliocalzo: Keeping the name: the same key binds both enum families (`NVML_GPU_INSTANCE_PROFILE_*` in a `supported_profiles` row, `NVML_COMPUTE_INSTANCE_PROFILE_*` in a `compute_instances` entry), so a name committing to one is wrong in half its uses — but you were right that nothing said which end of the getter each field is, fixed in a7c07782, and a rename is still free until this merges if you want one.

deployments/nvml-mock/helm/nvml-mock/profiles/mig/b200.yaml
giuliocalzo: Keeping `memory_mb` as-is: it maps straight onto NVML's own `nvmlGpuInstanceProfileInfo_t.memorySizeMB`, so the raw number is what the driver reports — and `23GB` would actually be 512 MB over this row's 23040, since the "23" in `1g.23gb` is nvidia-smi's rounded display figure rather than the allocation.

deployments/nvml-mock/helm/nvml-mock/profiles/a100.yaml
giuliocalzo: Yes — this block is the board's *capability* (mode plus `max_gpu_instances`), and only the partition *list* moved out, which is what the second ConfigMap is for: the tables run 285-311 lines each, so inlining would take `a100.yaml` from 473 to 758 lines and ship a table that `t4` and `l40s` have no use for at all.

deployments/nvml-mock/helm/nvml-mock/templates/_helpers.tpl
giuliocalzo: Agreed it shouldn't stay here — though the constraint isn't MIG-specific: `gpu.profile` is already a single cluster-wide value, so a mixed-rack cluster is mis-simulated at the board level before MIG enters it, and `ControlPlaneSource` is still a skeleton (#614) with `FileSource` the only path a node actually has today, so dropping this gate would give no MIG rather than per-rack MIG.

`SGPURackProfile` already models the capability half (`GPUCapabilities.MIG.Supported` / `MaxGPUInstances`) and only lacks the layout, which is where `gpuInstances` belongs once projection is real. I'd rather do that as its own PR than here — this one is already +24k across 126 files — and the same goes for the gpu-operator/mig-manager side. Tracking it in #902.

docs/guides/mig/README.md
giuliocalzo: Added in d6f13972 — [Supported MIG Profiles](https://docs.nvidia.com/datacenter/tesla/mig-user-guide/latest/supported-mig-profiles.html) is now linked from the "How slices are named" section here, and from `docs/mig.md` where a new board's rows get transcribed, both noting it has no GB200 or GB300 table.

roma-glushko (approved):

PR #830, opened 2026-09-11
chore(release): cut v0.4.0-rc1

ArangoGutierrez:
## What This PR Does

Cuts the v0.4.0-rc1 release candidate.

- `nvml-mock` chart `version` 0.3.0 -> 0.4.0-rc1
- `CHANGELOG.md`: `[Unreleased]` promoted to `[0.4.0-rc1] - 2026-09-14`, with a
  fresh empty `[Unreleased]` above it

Two files. `appVersion` stays at `550.163.01` because it names the NVIDIA driver
version the mock presents, not this project's release. `mokka-crds` is untouched:
it is already at 0.4.0 and published, from #826.

## Why an rc, and why a full prep

Following the v0.3.0 precedent, where the candidate carried its own chart version
and its own changelog section, and the final release later absorbed it:

    v0.3.0-rc1   chart 0.3.0-rc1   CHANGELOG ## [0.3.0-rc1] - 2026-07-31
    v0.3.0       chart 0.3.0       CHANGELOG ## [0.3.0] - 2026-08-01

## The CHANGELOG consolidation

`[Unreleased]` had accumulated duplicated subheadings across the cycle: two
`Added`, two `Changed`, two `Fixed` and three `Removed`. Rendered as-is the notes
would repeat each of those sections. They are merged into one of each, in
Keep-a-Changelog order.

Nothing was reworded or dropped. 63 bullets before, 63 after, and every
non-heading content line is identical as a multiset, so the only change is
regrouping.

## Note on the version gate

#840 decoupled the helm unittest suites from the chart version, so this bump
needs no test or snapshot changes. It also means the suite no longer fails when
the chart version is wrong: reverting the bump here leaves all 173 tests and 13
snapshots green, where before #840 it produced 14 test and 11 snapshot failures.

Not a blocker for this PR, but worth a follow-up: a CI check that the chart
version matches the release tag would restore that safety without bringing back
the per-release assertion churn #840 removed.

## Verification

    make helm-tests        173 tests, 13 snapshots, rc=0
    make helm-crds-tests   rc=0

## Checklist


giuliocalzo:
I've opened a [PR](https://github.com/NVIDIA/k8s-test-infra/pull/840) to fix once for all the helm unittest snapshot

giuliocalzo (dismissed):

giuliocalzo (approved):

PR #840, opened 2026-09-14
test(chart): decouple the helm unittest suites from the chart version

giuliocalzo:
## What This PR Does

Pins every `nvml-mock` helm-unittest suite to a fixed chart version:

```yaml
chart:
  version: 0.0.0-dev
  appVersion: 0.0.0-dev
```

and drops the assertions on the two version-bearing labels,
`metadata.labels["helm.sh/chart"]` (in `rbac_test.yaml`,
`configmap_test.yaml`, `daemonset_test.yaml`) and
`metadata.labels["app.kubernetes.io/version"]` (in `daemonset_test.yaml`).
The three affected snapshots are re-recorded; the diff there is only the
two label values.

## Why

The suites rendered against whatever `Chart.yaml` said, so `0.3.0` and
`550.163.01` were baked into three suites and eleven snapshot entries. Any
chart release therefore had to rewrite expectations and re-record snapshots
in files that had nothing to do with the release, burying the real diff.

The dropped assertions were not earning their keep either: both labels come
verbatim from the shared `nvml-mock.labels` helper, so asserting them
re-tested Helm's own templating rather than any chart behaviour. Each
"should include standard labels" case still covers the labels that carry
chart-specific meaning (`app.kubernetes.io/name`, `instance`, `managed-by`).

Verified by temporarily setting `Chart.yaml` to `version: 9.9.9` /
`appVersion: "999.99.99"`: the full suite still passes, where before it
failed 11 tests.

## Checklist

- [x] Tests pass (`make helm-tests`: 9 suites, 173 tests, 13 snapshots)
- [ ] Linter passes (`make lint-fix`) — no Go code changed
- [ ] New code has SPDX license headers — no new files
- [ ] CHANGELOG.md updated (if user-facing change) — test-only change


ArangoGutierrez (approved):
/lgtm

Issue #776, opened 2026-09-01
GPU Operator Scale Testing


roma-glushko:
We are planning to do scale testing of [NVIDIA GPU Operator](https://github.com/nvidia/gpu-operator) using Mokka.

With this effort, the GPU operator team is looking for answering these questions:

- How does the GPU Operator behave when there are large number of nodes? CPU, memory, informer cache sizes, API server request rate, etc. This can be simulated using kwok by having X fake nodes.
- What is the performance when upgrades are orchestrated across a large number of nodes? Wall time, what happens on random failures, etc.

This brings to more things for us to investigate:

- How can we make it possible to run the original GPU operator's `driver`, `toolkit`, `migManager` with Mokka? (raised by @avasilevskii )
- GFD should do the job of setting all K8s labels (gonna be fully complete once MEP0003 is implemented) (raised by @avasilevskii )
- How can we make it possible to test GPU driver upgrades? (raised by @rajathagasthya )

PR #843, opened 2026-09-14
docs(mep): MEP0004: GPU Operator Driver Manager Integration

### LINKED ISSUE #841: Investigate how to integrate GPU Operator driver manager with Mokka

roma-glushko:
## What This PR Does

This MEP serves a specification and overview of what it takes us to integrate with GPU Operator Driver Manager.

The document is meant to be discussed with the whole team.

## Why

The https://github.com/NVIDIA/k8s-test-infra/issues/776 epic has motivated us to look into the driver manager integration.

## Checklist


avasilevskii:
As discussed on Slack: I wouldn't modify Mokka for that specific (relatively narrow) use-case.
Instead, I'd propose to provide stub images for the driver-manager init container and the main driver container. The rest can use Mokka: that way we'd keep the full-fidelity state machine simulation with workloads cordoning and re-scheduling at the minimal cost of replacing two components that are not that relevant for scalability (both can be tested in a single node scenario). The stubs should however simulate the delays introduced by real components (e.g. driver installation) to not skew the timings/expected QPS.

ArangoGutierrez:
> As discussed on Slack: I wouldn't modify Mokka for that specific (relatively narrow) use-case. Instead, I'd propose to provide stub images for the driver-manager init container and the main driver container. The rest can use Mokka: that way we'd keep the full-fidelity state machine simulation with workloads cordoning and re-scheduling at the minimal cost of replacing two components that are not that relevant for scalability (both can be tested in a single node scenario). The stubs should however simulate the delays introduced by real components (e.g. driver installation) to not skew the timings/expected QPS.

Agree with @avasilevskii  we need to find a plug and play solution for this corner cases

PR #846, opened 2026-09-14
docs(release): add a backport marker and a check that finds missed backports

ArangoGutierrez:
## What This PR Does

Records which merged PRs are supposed to reach the release branch, and adds a
check that reports the ones that have not.

- `.github/PULL_REQUEST_TEMPLATE.md` gains a **Backport** section carrying the
  agreed marker line, plus a checklist item for the matching label
- `hack/check-backports.sh` lists marked PRs that are missing from the release
  branch, exiting 1 when any are outstanding

The `backport/release-0.4` label already exists in the repo.

## Why

Agreed in the DGXC channel today: no code freeze. `main` stays a development
branch and is allowed to run ahead, the release line lives on `release-X.Y`, and
fixes for the open release are cherry-picked across.

That leaves one question unanswered — *how do we know what should be
cherry-picked?* Without a record, assembling RC2 means reading every merge commit
by hand and hoping nothing was missed. A missed backport is silent: the fix is on
`main`, looks merged, and simply is not in the release.

## Two markers, either of which counts

- the `backport/release-X.Y` **label** — queryable via the API, survives merge,
  cannot be typo'd
- the line `This PR SHOULD BE BACKPORTED TO release-X.Y BRANCH` in the PR body —
  the convention agreed in chat

Supporting both means the agreed text convention works immediately, while the
label gives something machine-checkable. Roma suggested a label twice in that
thread; this is that, without dropping what was agreed.

## How the check decides

A marked PR has landed when a commit on the release branch carries
`(cherry picked from commit <sha>)`, which `git cherry-pick -x` writes. That is
why the cherry-pick command in the docs uses `-x`: it is the provenance the check
reads.

    hack/check-backports.sh [release-branch] [remote]    # defaults: release-0.4 upstream

## Verification

Both halves were tested for discrimination, not just for running.

**Selection**, against a fixture covering every shape:

```
#101 label only                      selected
#102 marker only                     selected
#103 lowercase marker                selected
#104 both                            selected
#107 null body + label               selected
#105 neither                         NOT selected
#106 release-0.3 marker              NOT selected
#108 unmerged (no mergeCommit)       NOT selected
```

**Cherry-pick detection**, against a real repository with two commits where only
one was picked with `-x`: reports `landed` and `MISSING` respectively.

Running it against `release-0.4` today reports 0 outstanding, which is correct —
nothing has been marked yet. That result on its own proves nothing, which is why
the two tests above exist.

## Checklist

- [ ] Linter passes (`make lint-fix`) — no Go changes
- [ ] CHANGELOG.md updated (if user-facing change) — maintainer process, not user-facing


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -14,3 +28,4 @@
 - [ ] New code has SPDX license headers
 - [ ] Documentation updated (if applicable)
 - [ ] CHANGELOG.md updated (if user-facing change)
+- [ ] If this is a fix for the open release, the `backport/release-X.Y` label is applied
```
roma-glushko, line 31: I think the label + this notice is enough to establish the workflow. The "This PR SHOULD BE BACKPORTED TO release-0.4 BRANCH" part is redundant.

PR #847, opened 2026-09-14
feat(mocknvml): implement the power limit setters and the workload power profile getters and setters

giuliocalzo:
## What This PR Does

Makes `nvidia-smi -pl` and the whole `nvidia-smi power-profiles` subcommand work. Both were backed by generated stubs, so the mock reported power state it had no way to change, and `power-profiles` answered "Workload Power Profiles feature is not supported on this device" on every profile including Blackwell.

### Commands now available

| Command | Behaviour |
| --- | --- |
| `nvidia-smi -pl <watts>` | Caps the device. Moves the power management limit and the enforced limit — including DCGM's `NVML_FI_DEV_POWER_CURRENT_LIMIT` — but not the default limit, which is the board property a consumer resets back to. Refused outside the `min_limit_mw` / `max_limit_mw` constraints the device advertises. |
| `nvidia-smi power-profiles -l` | Lists the profiles the board advertises. |
| `nvidia-smi power-profiles -ld` | Same, plus each profile's priority and the profiles it conflicts with. |
| `nvidia-smi power-profiles -gr` | Reports the requested profiles. |
| `nvidia-smi power-profiles -ge` | Reports the enforced profiles, i.e. what survived arbitration. |
| `nvidia-smi power-profiles -sr <ids>` | Adds to the requested set. |
| `nvidia-smi power-profiles -cr <ids>` | Removes from the requested set. |

Writes go into the runtime override document rather than the writing process, so they behave like the driver-level writes they model: a later `nvidia-smi` reads the new value, and a reset (`nvidia-smi -r`, `nvml-mock-ctl reset`) clears it.

```console
$ nvidia-smi power-profiles -sr 2 -i 0
Successfully set the requested profiles.
$ nvidia-smi power-profiles -gr -i 0
2. Compute
```

### Which profiles support it

`gb200` and `gb300` advertise a profile set through a new `power.workload_power_profiles` config block. Two axes decline, and they decline differently: a board that declares no profiles reports the feature unsupported, while a driver older than 570 does not export the symbols at all.

| Profile | Driver | `power-profiles -l` |
| --- | --- | --- |
| `gb200` | 580.65.06 | lists 8 profiles, exit 0 |
| `gb300` | 570.124.06 | lists 8 profiles, exit 0 |
| `b200` | 560.35.03 | `Function Not Found`, exit 13 |
| `h100` / `a100` | 550.163.01 | `Function Not Found`, exit 13 |

`b200` is the interesting case: Blackwell silicon, but its profile pins a driver that predates the API, so it stays declined. A unit test pins this table so "lists its profiles" cannot quietly become "always lists them".

### Behaviour worth knowing

**Requested and enforced are separate**, because asking for mutually exclusive profiles is allowed. Enforced is what survives arbitration: profiles are admitted in NVML's priority order (lower wins) and one is dropped when it conflicts with a profile already admitted, ties broken on id so the answer does not depend on config ordering.

**The shipped profiles request nothing**, matching every real GB200, GB300 and B200 capture under `testdata/hardware`, which report `N/A` for both masks. `nvidia-smi -q -x` still matches those captures byte for byte.

**An unadvertised id is refused** with `INVALID_ARGUMENT` rather than partially applied, so a caller naming one bad profile alongside several valid ones gets a clean refusal. Config keeps ignoring an unknown id; a live caller is told.

**Two nvidia-smi behaviours, neither the mock's.** Within one invocation only `-ge` reflects a write, because nvidia-smi evaluates `-sr` / `-cr` before `-ge` but after `-gr`. And a comma-separated list is applied in full only to the first GPU it visits, the rest getting just the first profile — so the write specs are scoped with `-i 0`.

## Verification

- `make gen` drops all seven stubs (260 → 253). `make test`, `make test-e2e-framework` and `make helm-tests` pass; `golangci-lint` reports 0 issues.
- Unit tests cover the cap round trip, the untouched default limit, boundary and out-of-range caps, the `_v2` scopes that decline rather than fold into the GPU limit, injected lost devices, the DCGM field-value route, priority arbitration including a three-way conflict, mask bits crossing the 32-bit word boundary (ids 31/33/254), and clearing everything staying distinguishable from never having written.
- Layout tests pin every struct size and version tag against go-nvml, which is what catches a mask decoded at the wrong offset — that still returns `SUCCESS`.
- `make test-mocknvml-bridge` drives the real C ABI through the Docker harness, reading every write back rather than trusting its return code: **113 passed, 0 failed**. `nvidia-smi 580` calls the two *deprecated* requested-profile entry points rather than the current one, found by tracing stub calls, so all three are implemented — leaving the current one alone would have left `-sr` and `-cr` failing.
- New e2e specs in the standalone scenario assert the configured list on every GPU, nothing requested or enforced by default, a refusal on the declining profiles, a set/clear round trip, arbitration between conflicting profiles, and a request set by one `nvidia-smi` still visible to the next. Each write spec clears what it asked for, since a request now outlives the invocation that made it.

Two decisions were settled against real nvidia-smi rather than by unit tests, which would have cemented either choice: the profile metadata array is indexed by profile id rather than densely packed (packing renders the wrong priority and conflicts for every profile past a gap), and the requested-profile masks default to empty.

Addresses two of the three setter families in #849; persistence mode (`nvidia-smi -pm`) is still per-process, so the issue stays open.

## Checklist


roma-glushko (approved):
Looks great 👍

ArangoGutierrez:
Nice work on this one. Recording the cap and the profile request in the override document is what finally makes `nvidia-smi -pl` followed by a separate `nvidia-smi` report what a real driver would.

Two follow-ups on the tests, both post-merge, since each one stays green when the thing it covers is taken out.

`TestSetPowerLimit_FailsWithoutADocumentPath` asserts a bare `require.Error`. Delete the `if path == ""` guard from `mutateDevice` and it still passes: `LockOverride("")` drops a stray `.lock` into `pkg/gpu/mockctl/` and `WriteAtomic("")` then fails on the rename, so an error comes back either way. `errNoOverrideDocument` is already there as a sentinel, so `require.ErrorIs(t, err, errNoOverrideDocument)` would pin the branch the test name promises.

`TestWorkloadPowerProfilesAreBlackwellOnDriver570` never exercises its driver half. `SupportsWorkloadPowerProfiles` is `declared && DriverMajor() >= workloadPowerProfileMinDriver`, and of the shipped profiles only gb200 (580.65.06) and gb300 (570.124.06) declare a block, both above the threshold. Cutting the function down to `return p.workloadProfilesDeclared` leaves the test passing. The comment calls b200 the case that pins this, but b200 declines on `declared` alone, and the loop's `require.False(t, declared)` rules out the one fixture that would flip the clause. A profile declaring a block on a 560 driver, or a per-profile `DriverMajor()` assertion, would close it.

The rest held up well. The layout tests pin the C and go-nvml sides against sizes derived by hand rather than reading one side from the other, and the nvidia-smi assertions carry real negative controls for both the unsupported board and the pre-570 driver.

One note on Greptile's walkthrough, which asked to hold the merge until CDI-injected containers receive the generated `/proc/modules` file. Nothing in this diff touches either. `/proc/modules` does match `tests/e2e/go/assertions/kmod.go` and `internal/agent/gpudriver/stage.go` on the tree, so that is a real absence rather than a search that could not hit, and neither file is in the changed set. The only `modules` token in the diff is an unchanged CHANGELOG context line from #798. Merging was the right call.

PR #864, opened 2026-09-16
docs: add managed EKS installation guide

### LINKED ISSUE #817: Document how to install Mokka in a real managed K8s cluster
We have spent a lot of time running Mokka in kind. That experience has been documented. We had to create a custom kind image and setup CDI there.

What it takes to run Mokka in a managed K8s env like EKS? We should try it out and create an installation page that guides our users through the process.

olegleyz:
## Summary

- Add a self-contained guide for installing Mokka on prepared Amazon EKS CPU workers.
- Provide optional Terraform for a disposable reference cluster using the AWS accelerated AL2023 image on CPU-only instances.
- Install the pinned Mokka and NVIDIA device-plugin releases and verify GPU allocation from a neutral Ubuntu workload.
- Validate the Terraform configuration in CI and track its dependencies quarterly with Dependabot.

## Verified configuration

- EKS 1.35 with two private, on-demand `t3.large` workers across two availability zones
- EKS-optimized accelerated AL2023 x86_64 image `1.35.7-20260911`
- NVIDIA runtime configured in CDI mode without physical GPUs
- Mokka chart and image `0.4.0-rc1` using the `gb300` profile
- NVIDIA device plugin `0.18.2` using its default `envvar` allocation strategy

These versions define the tested reference configuration, not an AWS support statement.

## Validation

- `make docs`
- `make managed-eks-terraform-check`
- Terraform apply followed by a no-change plan
- Healthy Mokka and device-plugin DaemonSets across both CPU workers
- Neutral `ubuntu:24.04` pod requesting one `nvidia.com/gpu` and reporting one simulated GB300

Closes #817


.github/workflows/deploy-pages.yaml
```diff
@@ -63,6 +65,14 @@ jobs:
       - name: Install documentation dependencies
         run: make docs-deps
 
+      - name: Set up Terraform
+        uses: hashicorp/setup-terraform@b9cd54a3c349d3f38e8881555d616ced269862dd # v3
+        with:
+          terraform_version: 1.16.2
```
ArangoGutierrez: Can we make sure dependabot tracks this

docs/guides/install/aws/eks/terraform/main.tf
```diff
@@ -0,0 +1,140 @@
+data "aws_availability_zones" "available" {
```
ArangoGutierrez, line 4: License header

ArangoGutierrez (commented):
Awesome documentation Oleg! just a couple nits

docs/guides/install/aws/eks/terraform/main.tf
olegleyz: Addressed in eb296d17. I added the standard NVIDIA copyright and SPDX license headers to every new hand-authored Terraform and YAML asset. The generated `.terraform.lock.hcl` is covered through `REUSE.toml` rather than editing the generated file. I am leaving this thread open for reviewer verification and resolution.

.github/workflows/deploy-pages.yaml
olegleyz: Addressed in eb296d17. Dependabot now scans the Terraform directory quarterly for provider and module updates, reducing update noise; the existing GitHub Actions ecosystem entry tracks the setup-terraform action. The workflow keeps the CLI input at `~1.16.0`, so setup-terraform automatically resolves new 1.16 patch releases while minor upgrades remain explicit review decisions.

mkdocs.yml
```diff
@@ -139,6 +139,7 @@ nav:
       - Overview: guides/README.md
       - Scenarios:
           - NVIDIA Device Plugin: guides/device-plugin.md
+          - Amazon EKS: guides/managed-eks/README.md
```
roma-glushko: I have been thinking to bubble it up to the main tab -> installation group so we have kind and EKS installation of Mokka in a very visible place.

docs/guides/install/aws/eks/terraform/main.tf
```diff
@@ -0,0 +1,116 @@
...
+    mokka = {
+      ami_type            = "AL2023_x86_64_NVIDIA"
+      ami_release_version = var.eks_ami_release_version
+      instance_types      = var.worker_instance_types
+
+      min_size     = var.worker_count
+      max_size     = var.worker_count
+      desired_size = var.worker_count
+
+      disk_size = 30
```
ArangoGutierrez, line 81: `disk_size` only reaches `aws_eks_node_group` when `use_custom_launch_template` is false, and it stays true here because `enable_bootstrap_user_data` needs the custom template, so v21.25.0 resolves this to null and the workers boot on whatever root volume the accelerated AL2023 AMI declares. Setting it through `block_device_mappings` with `volume_size = 30` and `encrypted = true` applies the size and takes the volume off the account default for EBS encryption.

ArangoGutierrez (approved):
Thanks for this, the EKS guide fills a real gap, and forcing the container runtime into CDI mode from the post-nodeadm bootstrap is exactly the part people get wrong on managed nodes.

This lands a self-contained managed-cluster path: assets pulled straight from the repo with no checkout needed, a pinned accelerated AL2023 node group that prepares the runtime before kubelet starts, and `make managed-eks-terraform-check` in the docs workflow so the Terraform cannot drift unnoticed.

docs/guides/install/aws/eks/README.md
roma-glushko, line 1: `docs/guides/install/aws/eks/` - I think we should have a terraform module like this as an example but just not hidden inside docs. Let's have a `examples/{name_of_the_example/*` structure and put that content of the TF module & other needed files here. That would serve as a template for our users to based upon.

PR #866, opened 2026-09-16
test(e2e): drop the dead validate/spike scripts, port the last two checks to Go

giuliocalzo:
## What This PR Does

Deletes the ten `tests/e2e/validate-*.sh` / `spike-*.sh` scripts, plus two
orphaned manifests and `tests/redeploy.sh`. Before deleting, each script was
audited check-by-check against the Go suite; the two checks that had no Go
equivalent are ported here rather than dropped.

**Ported to Go** (new specs, `ibfabric` label, standalone scenario):

- `assertions.IBNetDiscover` — the `libibnetdisc` directed-route walk must print
  a `Ca` record and reach a non-local GUID. The local GUID set collects
  `node_guid` and `sys_image_guid` as well as `port_guid`, because the mock
  derives the node GUID from the port GUID by clearing the EUI-64 U/L bit
  (`fabric.nodeGUIDFromPortGUID`). A port-only set lets the local CA's own node
  GUID pass as a cross-pod peer and false-passes the assertion — there is a unit
  test pinning exactly that.
- `assertions.SMInfo` — `sminfo` must name a master SM with a non-zero GUID, and
  both pods must name the same one, which is the fabric-global "one SM identity"
  property a per-pod mock would not have.

The parsing for both lives in an untagged file (`assertions/ibfabric.go`) so its
15 unit tests run in the regular `go test ./...` job, following the pattern
`gfd_labels.go` established. The GUID helpers that were inline in
`iblinkinfo.go` moved there and are now shared.

**Also removed:**

- `tests/e2e/gfd-mock.yaml` and `validator-mock.yaml` — orphaned copies of the
  manifests the suite embeds from `go/assets/`. The `gfd` one named GFD v0.17.0,
  an image path `VERSION-MATRIX.md` records as unresolvable; the active copy
  pinned at v0.8.2 is authoritative.
- `tests/redeploy.sh` — superseded by the Tilt workflow.

Comments and docs that cited a now-deleted path were rewritten to describe the
check instead.

## Why

No `Makefile` target, Tiltfile resource or workflow reached any of these
scripts. The `nvml-mock-e2e.yaml` workflow that used to run them is gone; only
`nvml-mock-e2e-go.yaml` remains, and it runs the Go suite. Several assertions
already said in their own doc comments that they ported a specific script.

The audit found every assertion in all ten scripts has a Go home, most of them
stronger than the original — `validate-nvidia-smi.sh` grepped `nvidia-smi -L`
for a product name, while `nvidiasmi.Inventory` decodes `-q -x` and checks
`attached_gpus`, per-GPU `product_name`, element-count consistency and phantom
processes.

Two things genuinely go away, both from `spike-dcgm.sh` and neither an
assertion: standalone DCGM (`nv-hostengine` + `dcgmi` driven directly against
the mock via `LD_LIBRARY_PATH`) is no longer exercised anywhere, so the
`dcgmi health` watches and the `[NVML-STUB]` hit capture have no successor. The
health report was printed and never asserted on, and the stub capture is a
discovery tool rather than a test. `VERSION-MATRIX.md` now records those DCGM
versions as what a manual spike last resolved, not as something that runs.

Because `sminfo` and `ibnetdiscover` now run under the deny-list default label
filter, they execute on every pipeline instead of by hand.

## Checklist

Verified locally: `helm-tests` (173 tests), `test-e2e-framework`,
`hack/check-doc-links.sh`, `golangci-lint run ./...` (0 issues), and the Ginkgo
suite compiles under the `e2e` tag.


ArangoGutierrez (approved):
/lgtm

PR #867, opened 2026-09-16
feat(gpuarch): order GPU architecture generations behind one comparable type

giuliocalzo:
## What This PR Does

Adds `internal/gpuarch`, a dependency-free type that orders NVIDIA GPU architecture generations, and moves the two places that compared architectures by hand onto it.

`Arch` carries NVML's own values (`NVML_DEVICE_ARCH_*`), so the mock engine converts with a plain cast. It offers the comparisons the call sites actually want — `AtLeast`, `Before`, `Is`, `Between` — plus `String` and `Parse`.

Behaviour changes, all covered in `CHANGELOG.md`:

- `device_defaults.architecture` accepts `rubin`, which NVML defines and the mock previously resolved to `NVML_DEVICE_ARCH_UNKNOWN`. Names are also matched case-insensitively with surrounding whitespace trimmed; an exact match was required before.
- An unrecognized architecture is now logged as a warning before degrading to `UNKNOWN`. The value is unchanged — nothing in the config path reported the typo, so an operator saw every architecture-gated feature report unsupported with no diagnostic anywhere.
- The e2e harness rejects a chart profile whose `device_defaults.architecture` is missing or names no recognized generation. This is test-suite strictness only; the chart and the mock still accept such a profile.

## Why

`NVML_DEVICE_ARCH_UNKNOWN` is `0xFFFFFFFF`, so a bare `>=` reports unidentified hardware as *newer* than Blackwell. Both existing gates had to pair the comparison with an explicit `!= DEVICE_ARCH_UNKNOWN` — a pairing that is invisible in review and silently wrong when omitted. Every comparison on `Arch` answers false for `Unknown` in either operand position, so that invariant lives in one place instead of being re-derived at each call site.

The e2e harness had separately grown a `preAmpereArchitectures` map keyed on YAML spellings: the same question answered a second way, with its own list to forget to update. It now derives its expectations from the same type, and `Profile` stores the parsed generation instead of both the generation and the raw string.

The package takes no NVML dependency because go-nvml does not build with `CGO_ENABLED=0` and the e2e harness must compare architectures without taking on cgo. `TestArchConstantsMatchNVML` pins the two numberings together, so a renumbering in go-nvml breaks the build rather than the gates.

One detail worth flagging for review: NVML's values are monotonic but not dense — 11 and 12 are reserved and unpublished, so `Blackwell` (10) is followed by `Rubin` (13). Ordering by the raw value stays correct across that gap, and a test walks the whole set to keep it that way.

## Checklist


roma-glushko (dismissed):
LOVE THIS PR 😍

roma-glushko (approved):

PR #870, opened 2026-09-16
feat(node-agent): simulate the modules the GPU Operator validator reads

### LINKED ISSUE #869: [Feature]: Simulate the kernel modules the GPU Operator validator checks
## What

[#798](https://github.com/NVIDIA/k8s-test-infra/pull/798) simulates two kernel modules: `nvidia` and `nvidia_uvm`. The GPU Operator validator checks five modules by name. Four of them we do not simulate.

| module | how the validator checks it | simulated |
| --- | --- | --- |
| `nvidia` | stats and reads `/sys/module/nvidia/refcnt` | yes |
| `nvidia_peermem` | `lsmod \| grep -E '^nvidia_peermem\s'` | no |
| `gdrdrv` | `lsmod \| grep -E '^gdrdrv\s'` | no |
| `nvidia_fs` | `lsmod \| grep nvidia_fs` | no |
| `mlx5_core` | `lsmod \| grep mlx5_core` | no |

Three of the checks sit behind feature flags: `GDRCOPY_ENABLED`, `GDS_ENABLED` and `GPU_DIRECT_RDMA_ENABLED` select `gdrdrv`, `nvidia_fs` and `nvidia_peermem`. The `mlx5_core` check runs instead when `USE_HOST_MOFED` is `true`, and the `nvidia` check is part of driver validation and always runs.

So the mock works today only while those features are off. Enable GDRCopy, GPUDirect Storage or GPUDirect RDMA against it and the check never passes: the validator errors, or retries forever when run with its wait flag.

`nvidia_modeset` is worth adding in the same pass. No consumer checks it, but a real node loads it, so the mirror looks wrong without it.

This makes the modules visible, not functional. GPUDirect RDMA itself is [#374](https://github.com/NVIDIA/k8s-test-infra/issues/374).

## Proposal

Add the modules below to `Modules()` in `internal/kmod/modules.go`. That function is the only production source of the simulated set, called once from `internal/agent/gpudriver/stage.go:228`. It drives three surfaces: `/proc/modules`, `/sys/module/<name>/` and the generated `lsmod`.

Delivery does not change. The CDI specs mount `sys/module` whole. `libmockfs.so` already redirects both paths.

| module | coresize | present when |
| --- | --- | --- |
| `nvidia_modeset` | 2097152 | always |
| `gdrdrv` | 262144 | always |
| `nvidia_fs` | 245760 | always |
| `nvidia_peermem` | 16384 | `IBEnabled` |
| `mlx5_core` | 3145728 | `IBEnabled` |

Give each one no holders of its own and no `version`, as `nvidia_uvm` has today. Sizes are build-specific, so treat them as representative rather than exact.

One knock-on: on a real node `mlx5_core` is held by `mlx5_ib`, so `lsmod` will print `0` in its Used-by column where a real node prints `1 mlx5_ib`. Nothing checks that, and adding `mlx5_ib` purely to hold it is not worth the entry.

`nvidia_peermem` and `mlx5_core` need a fabric, and real nodes only load them where one is configured, so both follow `IBEnabled` at `internal/agent/state.go:263`. That means `Modules()` takes the state rather than just the driver version. `writeKernelModules` already holds the whole `*agent.State` and narrows it on the way in.

Each module added to `nvidia`'s used-by list raises its refcount, because `Refcnt()` returns `len(Holders)`. Real captures disagree about which of these appear there, so pick a list and do not over-invest.

Adding all five locally fails 12 tests, and only those: 4 in `internal/kmod/modules_test.go`, 6 in `internal/kmod/render_test.go`, 1 in `internal/kmod/lsmod_test.go` and 1 in `internal/agent/gpudriver/gpudriver_test.go`. They all pin `nvidia`'s refcount or name its holder list. The count is the same whether one module is added or five, so do the whole set in one change.

Two surfaces assert the same values but could not be run here, so expect them on top: `shims/libmockfs/shim_test.go:109`, whose shim is glibc-only, and three e2e assertions in `tests/e2e/go/assertions/kmod.go:61` and `:109` and `tests/e2e/go/scenario_nri_test.go:496`.

Two documents name the simulated modules and go stale otherwise: `docs/helm-chart.md` and the surface table in `enhancements/meps/0003-node-agent/README.md`.

## The nvidia_modeset fixture conflict

The test suite uses `nvidia_modeset` as a stand-in for a module that the host loads and we do not simulate. The name appears 15 times in `internal/kmod/render_test.go`. Lines 281-283 state the premise:

```go
require.NoFileExists(t, filepath.Join(served, "holders", "nvidia_modeset"),
    "the previous pass's host holder must not survive onto a simulated module")
```

These tests prove that host values and simulated values stay separate. They work because `nvidia_modeset` is not simulated. Simulating it inverts the premise.

The conflict is hidden today. An earlier refcount failure stops the same test first. Correct that refcount expectation and the assertion fails on its own.

Rename the fixture to something synthetic, such as `host_only_holder`. Any real module name can become a simulation target later and reproduce this exact conflict, which is how `nvidia_modeset` became a problem in the first place.

faganihajizada:
Closes #869.

## What

The GPU Operator validator reads five kernel modules. We simulated one of them, `nvidia`.

| module | how the validator reads it | gated by |
| --- | --- | --- |
| `nvidia` | stats `/sys/module/nvidia/refcnt` | always runs |
| `gdrdrv` | `lsmod \| grep -E '^gdrdrv\s'` | `GDRCOPY_ENABLED` |
| `nvidia_fs` | `lsmod \| grep nvidia_fs` | `GDS_ENABLED` |
| `nvidia_peermem` | `lsmod \| grep -E '^nvidia_peermem\s'` | `GPU_DIRECT_RDMA_ENABLED` |
| `mlx5_core` | `lsmod \| grep mlx5_core` | `USE_HOST_MOFED` |

So enabling GDRCopy, GPUDirect Storage or GPUDirect RDMA stopped at a check for a module the mock never served. The validator errors, or retries forever behind its wait flag.

## Change

`Modules()` now serves `nvidia_modeset`, `gdrdrv` and `nvidia_fs` alongside the existing two. It adds `nvidia_peermem` and `mlx5_core` when InfiniBand is enabled.

Sizes for `nvidia_modeset`, `gdrdrv` and `mlx5_core` were measured on two aarch64 nodes running driver 580.173.02, which reported identical values.

## Test fixture rename

`render_test.go` used `nvidia_modeset` as its stand-in for a module the host loads and we do not simulate. Simulating it inverts that premise. The fixture is now `host_only_holder`, which cannot become a simulation target later.

## Verification

Verified in a kind cluster against this build. The `kmod` and `nri-cdi-kmod` e2e specs pass, and all four validator greps match through the `libmockfs` redirect.


giuliocalzo (dismissed):

giuliocalzo (dismissed):

docs/faq.md
```diff
@@ -8,7 +8,6 @@ than fail in an interesting way:
 
 | Surface | Consumer that reads it |
 |---|---|
-| Kernel module presence — `/proc/modules`, `/sys/module/nvidia/` | `lsmod`, the GPU Operator driver-container gate, DCGM startup checks |
```
roma-glushko, line 11: That's a great 🙌

internal/agent/gpudriver/stage.go
```diff
@@ -225,7 +225,7 @@ const kernelSysModuleDir = "module"
 var kernelProcModules = "/proc/modules"
 
 func writeKernelModules(ctx context.Context, h *host.Host, state *agent.State) error {
-	mods := kmod.Modules(state.Software.DriverVersion)
+	mods := kmod.Modules(state.Software.DriverVersion, state.NodeShape.Network.IBEnabled)
```
roma-glushko, line 228: Would be nice with experiment with loading these network IB modules in the IB simulator instead of GPU driver simulator. That would make it more coherent with IB simulation e.g. you could answer the question like "what IB footprint we simulate right now?" by just looking at the corresponding IB simulator.

internal/kmod/modules.go
```diff
@@ -19,10 +19,17 @@ import (
...
 )
 
-// NVIDIA and NVIDIAUVM are simulated module names.
+// Simulated module names. The GPU Operator validator greps lsmod for
+// nvidia_fs, gdrdrv, nvidia_peermem and mlx5_core, and stats nvidia's refcnt.
+// Nothing checks nvidia_uvm or nvidia_modeset; a real node loads them.
 const (
-	NVIDIA    = "nvidia"
-	NVIDIAUVM = "nvidia_uvm"
+	NVIDIA        = "nvidia"
```
roma-glushko, line 26: Thank you for spending some time researching the whole set of modules to simulate 🙌

roma-glushko (approved):

internal/agent/gpudriver/stage.go
faganihajizada: Good idea in principle, and I agree the IB should be answerable in one place. Do you want to handle it in this PR or separate effort?

cc @olegleyz since you were looking into network mock area

giuliocalzo (approved):

Issue #506, opened 2026-07-25
GPU metric values are static and do not respond to workload


ArangoGutierrez:
## Summary

Mokka's GPU metric values are static. They do not respond to workload, so any check asserting "this metric reflects real work" cannot be satisfied, and any check asserting only that a metric *exists* passes regardless of whether the value is meaningful.

Found during the Mokka + AICR pre-silicon preflight POC (2026-07-25). This is the single largest honesty caveat in the coverage catalog.

## What is static today

| Surface | Current behaviour |
|---|---|
| `processes` | Empty list in every shipped profile |
| `memory.used_bytes` | Constant 0 |
| DCGM `DCGM_FI_PROF_*` | Fixed ratios of a utilisation number that is itself 0 unless the opt-in simulator is enabled |
| NVLink counters | Accrue off wall-clock, not off traffic |

Cross-check before acting: these were reported by a capability survey during the POC and the profile YAML should be re-read directly, since counts and defaults drift.

## Why it matters for AICR coverage

`accelerator-metrics` requires `DCGM_FI_DEV_GPU_UTIL`, `FB_USED`, `GPU_TEMP` and `POWER_USAGE` to be **present**. It passed in the POC run. The exporter-to-NVML path is genuinely exercised, which is why it is bucket A, but the values themselves are checked by nobody. `FB_USED` is a constant zero and the check still passes.

This is why the POC's bucket B came out empty while the findings note still carries a caveat: AICR writes integration assertions rather than value assertions, so nothing is green *purely* because the mock answered, but where values are synthetic a check validates the path and stays silent on the values. That distinction should be a property of the simulator, not a footnote.

## What to do

Scope carefully. Full workload correlation is a large piece of work and probably not warranted. The useful subset:

1. **Make `memory.used_bytes` respond to allocation.** A pod holding a `nvidia.com/gpu` claim should move the number. This is the single most visible "the dashboard is obviously fake" tell.
2. **Populate `processes` from actual container claims.** The device plugin and DRA driver already know which containers hold which devices; surfacing that through NVML would make `nvidia-smi` output plausible and would let isolation checks assert something real.
3. **Drive `DCGM_FI_DEV_GPU_UTIL` from the opt-in simulator by default**, or document loudly that it is 0 without it.

Item 2 is the highest value for AICR specifically, because `secure-accelerator-access` reasons about which container can see which device.

## Explicitly out of scope

[...]

PR #872, opened 2026-09-17
feat(nvml-mock): add vr200 preview profile (Vera-Rubin Superchip)

### LINKED ISSUE #459: [Feature]: Add support for Vera Rubin NVL72
### Problem statement

We are missing the new, cutting-edge Vera Rubin rack on our profile list. 

### Proposed solution

- Add a new VR profile  
- Make sure it works against nvidia-smi & GPU Operator

### Scope / non-goals

Make it possible to mock the new VR-based cluster setups.

### Audience

End users

### Pre-submission checks

giuliocalzo:
## What This PR Does

Adds `vr200`, a preview GPU profile for a four-GPU Vera-Rubin tray: 288 GB of HBM4 per GPU, a 2.3 kW power ceiling, NVLink 6.0 with C2C to a Vera host CPU, a ConnectX-9 XDR fabric, and PCIe Gen6. It is built from an `nvidia-smi -q -x` capture on driver 615.23 / CUDA 13.4, which ships alongside it as a hardware fixture.

```bash
helm install nvml-mock ./deployments/nvml-mock/helm/nvml-mock --set gpu.profile=vr200
```

It ships as a preview: Vera-Rubin is pre-release silicon, so reported values will shift as drivers mature. The product name still reads `NVIDIA Graphics Device` because the board predates the driver's product table, and the `/dev/nvidia*` minor numbers do not follow the device index. What the capture cannot supply is either left absent (`workload_power_profiles`, `gpu_operation_mode`) or modelled from published NVIDIA specifications and labelled as an assumption in the profile — compute capability 10.7 (sm_107), the 18-link NVLink inventory, and the two 800 Gb/s HCAs per GPU.

The profile is covered like every other shipped one: the fake-gpu-operator integration publishes it, and it runs every scenario in the pull-request e2e matrix, DRA included. That last one needed #875, which pinned the driver chart to `dra-driver-nvidia-gpu` 0.5.0 — it builds against `go-nvlib` v0.12.0, whose `GetArchitectureAsString` carries a Rubin case, so the plugin resolves NVML architecture enum 13 instead of refusing the device. On the previously floating chart it failed with `error interpreting device architecture as string: 13`.

## Why

Mokka exists so that work depending on expensive GPU infrastructure can run on CPU nodes, and Rubin is the generation consumers are now preparing for. This profile lets them exercise scheduling, topology and power behaviour against a Rubin-class node well before running on one, with values traceable to a real capture rather than to guesswork.

## Checklist


deployments/nvml-mock/helm/nvml-mock/README.md
```diff
@@ -31,6 +31,13 @@ Set `gpu.profile` to one of the profiles shipped in `profiles/`:
 | `gb300` (default) | GB300 NVL | 288 GiB | Blackwell Ultra |
 | `l40s` | L40S | 48 GiB | Ada Lovelace |
 | `t4` | Tesla T4 | 16 GiB | Turing |
+| `vr200` | Vera Rubin NVL72 | 280 GiB | Rubin |
+
+`vr200` is a Vera-Rubin compute tray authored from a capture of pre-release
+hardware, so it differs from the others in two ways: it reports `NVIDIA Graphics
+Device` as its GPU name, because the board shipped before the driver's product
+table carried one, and it is not in the e2e suite's profile set, so its rendered
```
ArangoGutierrez: vr200 is in the workflow's default `gpu_profiles` now and ci.yaml passes no override, so e2e, e2e-gpu-operator and e2e-nri all run it. Only e2e-dra excludes it, so naming that one job would match what tests/e2e/README.md says.

docs/helm-chart.md
```diff
@@ -718,24 +731,33 @@ helm install nvml-mock oci://ghcr.io/nvidia/k8s-test-infra/chart/nvml-mock \
...
+| | A100 | H100 | B200 | GB200 | GB300 | L40S | T4 | VR200 |
+|---|---|---|---|---|---|---|---|---|
+| **Profile name** | `a100` | `h100` | `b200` | `gb200` | `gb300` | `l40s` | `t4` | `vr200` |
+| **Full name** | A100-SXM4-40GB | H100 80GB HBM3 | B200 | GB200 NVL | GB300 NVL | L40S | Tesla T4 | Vera Rubin NVL72 |
+| **Architecture** | Ampere | Hopper | Blackwell | Blackwell | Blackwell Ultra | Ada Lovelace | Turing | Rubin |
+| **Compute capability** | 8.0 | 9.0 | 10.0 | 10.0 | 10.0 | 8.9 | 7.5 | 10.7 |
+| **CUDA cores** | 6,912 | 16,896 | 18,432 | 18,432 | 21,632 | 18,176 | 2,560 | not captured |
+| **Memory** | 40 GiB HBM2e | 80 GiB HBM3 | 192 GiB HBM3e | 192 GiB HBM3e | 288 GiB HBM3e | 48 GiB GDDR6 | 16 GiB GDDR6 | 280 GiB |
+| **NVLink** | v3, 12 links | v4, 18 links | v5, 18 links | v5, 18 links | v5, 18 links | — | — | v6, 18 links |
+| **NVLink BW** | 600 GB/s | 900 GB/s | 1.8 TB/s | 1.8 TB/s | 1.8 TB/s | — | — | 1.8 TB/s |
```
ArangoGutierrez: Every other cell in this row is the bidirectional aggregate from its profile's own comment (a100 600 GB/s, h100 900 GB/s, Blackwell 1.8 TB/s), and vr200.yaml reads `18 links ~= 1.8 TB/s/GPU (3.6 TB/s bidir)`. At 1.8 TB/s the table shows Rubin matching Blackwell.

deployments/nvml-mock/helm/nvml-mock/values.yaml
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 gpu:
-  profile: gb300          # any *.yaml basename under profiles/ (e.g. a100, h100, b200, gb200, gb300, l40s, t4)
+  profile: gb300          # any *.yaml basename under profiles/ (e.g. a100, h100, b200, gb200, gb300, l40s, t4, vr200)
   # Number of GPUs to expose. Empty (default) derives the count from the
   # selected profile's `devices:` list (t4 and the Grace-Blackwell trays
   # gb200/gb300 -> 4, the 8-GPU baseboards -> 8), so the profile is the single
```
ArangoGutierrez: Two comments here still describe a seven-profile chart: this one lists the 4-GPU profiles as t4 plus gb200/gb300 when vr200 also renders 4, and the dynamicMetrics note at line 20 sends a vr200 reader to the 250W baseline when the profile sets `base_mw: 400000`. The `gpu.count` row in docs/helm-chart.md did get vr200.

ArangoGutierrez (dismissed):
Thanks for this, authoring vr200 from a real driver 615 capture and checking that capture in next to the profile makes the numbers verifiable instead of guessed.

Adds vr200 as the eighth chart profile and wires it into the pull-request e2e matrix, with the e2e-dra exclusion documented against the go-nvlib Rubin gap. Where the capture says nothing the profile says "not captured" rather than borrowing a Blackwell number, which is the right call. What is left is the spots that still describe a seven-profile chart and the per-profile assertion tables the new profile has not joined.

- `helm template --set integrations.fakeGpuOperator.enabled=true` renders eight ConfigMaps on this branch, but this sample output stops at gpu-profile-t4. docs/guides/runai-fgo/README.md also says "these seven ConfigMaps" and "Mokka owns those seven ConfigMap names". The "FGO ships its own profiles under the same seven names" wording stays right, since their builtin set has not grown. (docs/helm-chart.md:127) https://github.com/NVIDIA/k8s-test-infra/blob/9ebaef859fd5c25ead8127baf1f99e667b9df8e6/docs/helm-chart.md#L119-L131
- vr200 joined the architecture, C2C and rack-scale maps in this file but not `TestDerivations`, the table whose comment calls it the guard that fails when a profile drifts from its capture. It is the only profile with `hcas_per_gpu: 2`, so its 8 HCAs across 4 GPUs are pinned by no literal anywhere. `TestLoadConfig_AllProfilesConsistent` in pkg/gpu/mocknvml/engine/config_profiles_test.go has the same gap, and its `memGiB` field cannot express vr200: 300442189824 bytes is 279.81 GiB, not 280. (tests/e2e/go/profile/profile_test.go:62) https://github.com/NVIDIA/k8s-test-infra/blob/9ebaef859fd5c25ead8127baf1f99e667b9df8e6/tests/e2e/go/profile/profile_test.go#L55-L62
- `gpmProfiles` is documented as the Hopper and later profiles serving DCGM_FI_PROF_*, and Rubin qualifies, but vr200 is not in the map while e2e-gpu-operator now runs it. The GPM branch is skipped rather than failed, so the gap stays silent. (tests/e2e/go/scenario_gpu_operator_test.go:129) https://github.com/NVIDIA/k8s-test-infra/blob/9ebaef859fd5c25ead8127baf1f99e667b9df8e6/tests/e2e/go/scenario_gpu_operator_test.go#L128-L129
- customer_boost.go and conf_compute.go both say "all seven captures in testdata/hardware"; qx-vr200.xml makes it eight, and both claims hold for it. (tests/e2e/go/assertions/nvidiasmi/customer_boost.go:21) https://github.com/NVIDIA/k8s-test-infra/blob/9ebaef859fd5c25ead8127baf1f99e667b9df8e6/tests/e2e/go/assertions/nvidiasmi/customer_boost.go#L19-L23

.github/workflows/nvml-mock-e2e-go.yaml
```diff
@@ -314,6 +314,12 @@ jobs:
       fail-fast: false
       matrix:
         gpu-profile: ${{ fromJson(inputs.gpu_profiles) }}
+        # The DRA driver resolves NVML's architecture enum through go-nvlib,
```
roma-glushko: Do we need a task for this?

roma-glushko (dismissed):
Super happy we're finally adding it!

deployments/nvml-mock/helm/nvml-mock/README.md
giuliocalzo: Fixed in 6e0a8590. The paragraph now names the DRA scenario as the one that skips vr200, rather than claiming the profile sits outside the e2e set.

docs/helm-chart.md
giuliocalzo: Fixed in 6e0a8590 — the cell reads 3.6 TB/s, the bidirectional aggregate its own comment gives, which is the convention the rest of the row follows.

deployments/nvml-mock/helm/nvml-mock/values.yaml
giuliocalzo: Fixed in 6e0a8590. The device-count note now lists gb200/gb300/vr200 as the four-GPU superchip trays, and the dynamicMetrics note gives vr200 its ~400W base instead of sending a reader to the 250W baseline.

.github/workflows/nvml-mock-e2e-go.yaml
giuliocalzo: issue opened https://github.com/NVIDIA/k8s-test-infra/issues/874

giuliocalzo: Yes — filed #874. It carries the removal steps (a driver release on go-nvlib v0.12.0, dropping the matrix `exclude`, `make e2e-dra` green on vr200, and retiring the two docs notes), and links the upstream report. The workflow comment and the DRA guide now point at it in 5764d6a6.

pkg/gpu/mocknvml/engine/config_profiles_test.go
```diff
@@ -166,23 +166,29 @@ func TestLoadConfig_GB300Profile(t *testing.T) {
...
-		{"L40S", "l40s.yaml", "ada_lovelace", 8, 9, 48, 8},
-		{"T4", "t4.yaml", "turing", 7, 5, 16, 4},
+		{"A100", "a100.yaml", "ampere", 8, 0, 40 * giB, 8},
+		{"H100", "h100.yaml", "hopper", 9, 0, 80 * giB, 8},
+		{"B200", "b200.yaml", "blackwell", 10, 0, 192 * giB, 8},
+		{"GB200", "gb200.yaml", "blackwell", 10, 0, 192 * giB, 4},
+		{"GB300", "gb300.yaml", "blackwell", 10, 0, 288 * giB, 4},
+		{"L40S", "l40s.yaml", "ada_lovelace", 8, 9, 48 * giB, 8},
+		{"T4", "t4.yaml", "turing", 7, 5, 16 * giB, 4},
+		{"VR200", "vr200.yaml", "rubin", 10, 7, 300442189824, 4},
```
ArangoGutierrez: vr200 lands in this table but not in shippedProfiles in config_profiles_utilization_test.go, so TestProfiles_ReportNonZeroUtilization runs seven subtests, a100 through t4, and never vr200. Deleting vr200's dynamic_metrics block leaves that suite green, which is exactly the #506 regression the literal is hand-maintained to catch. Adding "vr200" there passes as-is, since the profile already ships the 10/45 gpu and 5/25 memory bands the constants assert.

ArangoGutierrez (dismissed):
Thanks for this, the vr200 profile is built from a real driver 615 capture and it says plainly which values could not be read from it instead of inventing them, which makes it easy to trust.

The new profile is placed on the correct side of each existing axis (C2C, platform identity, NV18 fabric, GPM metrics), so those table-driven guards keep discriminating rather than quietly becoming always-true, and switching the memory column to raw bytes is the right call for a board that reports 279.81 GiB. What is left is the hand-maintained profile lists under pkg/gpu/mocknvml/engine, which still enumerate the seven older profiles.

- gracePlatformProfiles still maps only the seven older filenames, so TestProfilePlatformIdentity checks neither side for vr200 even though the map's comment asks a new profile to choose one. The e2e platform spec and TestPlatformIdentityIsRackScaleOnly do cover this ground, so it is redundancy rather than a hole; slot 2 over tray 1 and module ids 1 through 4 satisfy the existing assertions if you add it. (pkg/gpu/mocknvml/engine/config_profiles_platform_test.go:27) https://github.com/NVIDIA/k8s-test-infra/blob/6b539e3544eb1c3bd1a2fb9e285fc612e005b45c/pkg/gpu/mocknvml/engine/config_profiles_platform_test.go#L25-L35

pkg/gpu/mocknvml/engine/config_profiles_test.go
```diff
@@ -166,23 +166,28 @@ func TestLoadConfig_GB300Profile(t *testing.T) {
...
-		{"L40S", "l40s.yaml", "ada_lovelace", 8, 9, 48, 8},
-		{"T4", "t4.yaml", "turing", 7, 5, 16, 4},
+		{"A100", "a100.yaml", "ampere", 8, 0, 40 * giB, 8},
+		{"H100", "h100.yaml", "hopper", 9, 0, 80 * giB, 8},
+		{"B200", "b200.yaml", "blackwell", 10, 0, 192 * giB, 8},
+		{"GB200", "gb200.yaml", "blackwell", 10, 0, 192 * giB, 4},
+		{"GB300", "gb300.yaml", "blackwell", 10, 0, 288 * giB, 4},
+		{"L40S", "l40s.yaml", "ada_lovelace", 8, 9, 48 * giB, 8},
+		{"T4", "t4.yaml", "turing", 7, 5, 16 * giB, 4},
+		{"VR200", "vr200.yaml", "rubin", 10, 7, 288 * giB, 4},
```
ArangoGutierrez, line 190: vr200 lands in this table but not in shippedProfiles in config_profiles_utilization_test.go, so TestProfiles_ReportNonZeroUtilization still runs seven subtests and never reaches vr200. Deleting the profile's dynamic_metrics block leaves that suite green, which is the #506 regression the literal is hand-maintained to catch. Adding "vr200" there passes as-is, since the profile already ships the 10/45 gpu and 5/25 memory bands the constants assert.

pkg/gpu/mocknvml/README.md
```diff
@@ -119,6 +119,7 @@ YAML configs allow full control over GPU properties. See `configs/` for examples
 - `mock-nvml-config-gb300.yaml` - GB300 NVL (4x Blackwell Ultra with 288 GiB HBM3e, 1.4 kW TDP)
 - `mock-nvml-config-l40s.yaml` - L40S (8x L40S, 48 GiB)
 - `mock-nvml-config-t4.yaml` - T4 (4x T4, 16 GiB)
+- `mock-nvml-config-vr200.yaml` - VR200 NVL (4x Rubin with 280 GiB, 2.3 kW power ceiling — one Vera-Rubin compute tray)
```
ArangoGutierrez: This entry says 280 GiB, but mock-nvml-config-vr200.yaml sets total_bytes to 309237645312, which is 288 GiB, and the chart README, docs/configuration.md and the comparison table all carry 288. It is the last place the pre-nameplate figure survives.

ArangoGutierrez (dismissed):
Thanks for this, shipping the vr200 profile alongside the nvidia-smi capture it was authored from means every value in it can be traced back to what the hardware actually reported.

vr200 threads all the way through: the engine config, the chart profile, the default e2e matrix, gpmProfiles and the docs tables. Moving total_bytes to the nameplate 288 GiB also puts the memory column back on a whole-GiB multiplier, so that table reads the same way for every profile.

- The CHANGELOG entry for this profile says two details follow the hardware rather than convention, and docs/helm-chart.md calls vr200 the only profile whose PCI domains are non-zero and whose minor numbers do not follow the device index. The minor numbers here run 0, 3, 1, 2 against gb300's 0, 1, 2, 3, so "one way" leaves out the half a chart reader would reach for vr200 to test. (deployments/nvml-mock/helm/nvml-mock/README.md:36)

pkg/gpu/mocknvml/engine/config_profiles_test.go
giuliocalzo: Good catch — added `vr200` to `shippedProfiles` in 8058e1c9. It passes unchanged, and `TestProfiles_ReportNonZeroUtilization` now runs eight subtests instead of seven, so dropping the profile's `dynamic_metrics` block fails here the way #506 intended.

pkg/gpu/mocknvml/README.md
giuliocalzo: Fixed in 8058e1c9 — that line now reads 288 GiB HBM4, matching `total_bytes`. It was the last 280 left in the tree.

tests/e2e/go/profile/profile_test.go
```diff
@@ -270,7 +273,7 @@ func TestWorkloadProfilePairsBackTheSetterAssertions(t *testing.T) {
 // become "always reports one". b200 is the interesting case — Blackwell, but a
 // board in no rack. See issue #642.
 func TestPlatformIdentityIsRackScaleOnly(t *testing.T) {
-	rackProfiles := map[string]bool{"gb200": true, "gb300": true}
+	rackProfiles := map[string]bool{"gb200": true, "gb300": true, "vr200": true}
```
ArangoGutierrez, line 276: vr200 joins `rackProfiles` here, but `TestPlatformIdentityModuleIDsAreDistinct` still loops over gb200 and gb300 only and `gracePlatformProfiles` in the engine tests has no vr200 entry, so setting all four vr200 `module_id`s to 1 keeps both packages green, where the same edit on gb300 fails both. Adding vr200 to that loop and `"vr200.yaml": true` to the map passes as-is, since its ids already run 1 to 4.

ArangoGutierrez (approved):
Thanks for this. Building vr200 from a real driver 615 capture, and shipping qx-vr200.xml next to it, means the new Rubin profile can be checked against what the boards actually report.

This adds vr200 as the eighth chart profile, a four-GPU Vera-Rubin tray, and wires it into the e2e matrix, the FGO ConfigMaps, the profile oracles and the docs. It is also the only profile with PCI domains other than 0000 and with minor numbers that don't follow the device index, so it doubles as a check on consumers that assume either.

- This paragraph says vr200 differs from the others in one way, its GPU name, while the CHANGELOG entry counts the minor numbers as a second difference and docs/helm-chart.md also calls it the only profile with non-zero PCI domains. Dropping the count, or linking the profile comparison in docs/helm-chart.md, keeps the three from disagreeing. (deployments/nvml-mock/helm/nvml-mock/README.md:37)
- The values.yaml comment now gives vr200 its ~400W base, but the same list in the `gpu.dynamicMetrics.power.*` row and in the Dynamic Metrics prose at line 876 still ends at gb300 ~800W. (docs/helm-chart.md:636) https://github.com/NVIDIA/k8s-test-infra/blob/8058e1c9fe1ed278357b820094fe2b22170804c1/docs/helm-chart.md#L636
- This calls VR200 the only capture carrying elements the schema has no field for, but `edpp_multiplier`, one of the examples listed, is also in qx-l40s.xml and qx-t4.xml. (tests/e2e/go/assertions/nvidiasmi/testdata/hardware/README.md:33)
- At 90C this is now the lowest shutdown threshold of any profile, but local/observability/scenarios/inject-thermal.sh still accepts `HOT_TEMP_C` up to 92 on the grounds that a100/h100 set the floor, so 91 or 92 on vr200 gets clamped and only fails once the poll budget runs out. Lowering that guard and the note in local/observability/README.md to 90 keeps the early failure. (deployments/nvml-mock/helm/nvml-mock/profiles/vr200.yaml:177)

PR #873, opened 2026-09-17
feat(chart): stamp GPU_PROFILE on the node-agent container

giuliocalzo:
## What This PR Does

Every `nvml-mock` pod now says which GPU it is simulating. The `node-agent` container exposes a `GPU_PROFILE` env var, joining the `GPU_COUNT` and `DRIVER_VERSION` it already advertised:

```console
$ kubectl exec -n nvml-mock-system ds/nvml-mock -- printenv GPU_PROFILE
gb300
```

A release configured inline via `gpu.customConfig` reports `custom`, since no named profile is in effect. Nothing in the agent reads the value; it is there for people and scripts. Set on the `node-agent` container only.

## Why

Answering "what hardware does this pod simulate?" from inside the container currently means parsing the mounted config, which every demo script and debugging session reinvents.

## Checklist

- [x] Tests pass — `make helm-tests` green at 176, up from 173. No Go code changed
- [ ] Linter passes (`make lint-fix`) — `golangci-lint` and `make gen` are clean; the run exits non-zero only at `govulncheck`, on go1.26.5 standard library advisories fixed in go1.26.6. Pre-existing and unrelated to this PR
- [x] New code has SPDX license headers — no new files
- [x] Documentation updated — `docs/tools/node-agent.md`
- [x] CHANGELOG.md updated

## Testing

Three Helm unit tests cover the default profile, an explicit `gpu.profile`, and the `customConfig` sentinel. The daemonset snapshot diff is nothing but the added env entry.


ArangoGutierrez (commented):
isn't this something we want the Mokka controller to own? instead of a hard coded from Helm ?

giuliocalzo:
> hing we want the Mokka controller to own? instead of a hard coded from Helm ?

when the controller will be ready, is going to inject such kind of values as well, but until then it's handy

roma-glushko (commented):
How would this behave with control plane? GPU profile is a dynamic thing. One and the same agent may have different profiles if the user repoint them after the daemon was deployed.

So static helm setting is not a way to go here. Node agent should own setting this information (because it knows what's the profile at any point of time) and likely it should not be env var but maybe a file?

> when the controller will be ready, is going to inject such kind of values as well, but until then it's handy

The Control plane is not going to create node agent daemonset pods. As I said, this is on node agent to set if we need it.

PR #884, opened 2026-09-18
feat: enable allocation-aware NRI by default

### LINKED ISSUE #814: NRI should be a default option for mokka injection
Right now NRI is described to an optional component that you need to enable to get working. But in reality if you don't do that there is nothing that would enable mokka for custom workloads, so they will be broken.

We should aim for just work experience for our users which means enabling NRI by default as the first-class Mokka component.

Update documentation, helm chart, local env, demos in order to enable NRI by default. Remove NRI demo and anything else that makes NRI look like an addon.

olegleyz:
## Summary

- enable the Helm chart NRI plugin by default and align Kind and EKS containerd runtime configuration
- make delivery allocation-aware per container: unallocated workloads and sidecars remain untouched, while GPU allocations receive the mock userspace driver without widening the selected GPU set
- preserve an explicit `nvml-mock.nvidia.com/devices: "true"` management path for all-GPU visibility without scheduler accounting
- gate InfiniBand and IMEX independently, and complete raw GPU allocations with the common NVIDIA control devices
- extend NRI and DRA coverage for zero-GPU workloads, exact allocation visibility, sidecar isolation, control nodes, and DRA claim visibility
- update the quickstart, architecture, operator documentation, and MEP-0002 follow-up contract

Closes #814.

## Validation

- `make test`
- `make helm-tests` (200 tests)
- `make docs` with the pinned local MkDocs environment
- `GOTOOLCHAIN=go1.26.6 make lint-fix` (generation, tidy, lint, vet, and govulncheck all pass)
- `GOTOOLCHAIN=go1.26.6 go test -tags=e2e -run "^$" ./tests/e2e/go/...`
- `git diff --check`
- `make e2e-nri` was attempted locally but could not start because this host has no `kind-mokka` context; the scenario is covered by PR CI


deployments/nvml-mock/helm/nvml-mock/templates/NOTES.txt
```diff
@@ -13,6 +13,19 @@ build-from-source instructions.
...
+DEFAULT: Ordinary workload through NRI
+
+  The chart deploys the NRI plugin by default. On a containerd node with NRI
+  enabled, a pod needs no GPU request, hostPath, or Mokka-specific environment:
+
+    kubectl run mokka-check --image=debian:bookworm-slim \
+      --restart=Never --command -- nvidia-smi -L
+    kubectl wait --for=jsonpath='{.status.phase}'=Succeeded \
+      pod/mokka-check --timeout=120s
+    kubectl logs mokka-check
```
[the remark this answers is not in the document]
olegleyz: Fixed in 84194252. The default NRI workload section and its trailing separator are now conditional on `nri.enabled`, and the Helm notes suite covers the disabled path. `make helm-tests` passes with 199 tests.

deployments/nvml-mock/helm/nvml-mock/templates/nri-daemonset.yaml
```diff
@@ -60,6 +60,7 @@ spec:
             # warns and injects everything else.
             - --imex-channel-annotation={{ .Values.nri.imexChannelAnnotation }}
             - --imex-channel-host-path={{ .Values.nri.overlay.hostPath }}/driver/dev/nvidia-caps-imex-channels
+            - --infiniband-annotation={{ .Values.nri.infinibandAnnotation }}
```
giuliocalzo, line 63: @olegleyz can we deploy as secondary container the NRI bin instead of having a second DS ?

ArangoGutierrez (commented):
Thanks for this. Tying the NRI overlay to a container's actual GPU allocation instead of injecting every pod on the node is the right default for both device-plugin and DRA clusters.

With `nri.enabled` defaulting to true, a container now gets the driver overlay only through a per-GPU device node, an NVIDIA CDI device (which covers DRA claims) or an explicit annotation, and the new E2E checks pin exact per-allocation visibility in both device-plugin and DRA mode. The open items sit at the edges of that selection rule and in the node image smoke test. Before merge, 6818fea6, 30b1a13d and 25f63748 need a Signed-off-by for DCO, and the branch conflicts with main in gpu-operator-values.yaml and VERSION-MATRIX.md. Main's #860 also changed how the GPU Operator reaches the driver tree, so the gpu-operator E2E needs a fresh run after the rebase.

internal/nri/inject/adjust.go
```diff
@@ -42,6 +48,55 @@ func Adjust(cfg Config, container Container) (Adjustment, bool) {
...
+	}
+	if selected.gpu() {
+		attachGPUs(cfg, container, adjustment)
+		attachAllocatedGPUControls(cfg, container, adjustment)
+	}
+}
+
+func hasAllocatedInfiniBand(container Container) bool {
+	for _, device := range container.Devices {
+		if filepath.Dir(device.Path) == "/dev/infiniband" {
```
ArangoGutierrez: containerd copies every host device node into a privileged container, so on a node with real RDMA hardware (the CI runners expose mlx5, see e308cc2e) `hasAllocatedInfiniBand` classifies privileged pods such as the DRA kubelet plugin, the GPU Operator validator or the documented device plugin run outside kube-system as IB-only and sets `MOCK_NVML_VISIBLE_DEVICES=none`, so they find zero GPUs. That is what dropped the DRA ResourceSlices to [0, 0] at c96cabbb, and it reproduces on kind by creating /dev/infiniband/uverbs0 on a worker, so the `devices: "true"` annotations added afterwards work around it without fixing it. Counting only the IB annotation (or an IB node the mock staged) as evidence, and never forcing `none` from inherited device nodes, keeps unannotated DRA, GPU Operator and device-plugin installs working on RDMA hosts.

deployments/kind-nvidia-cdi/Makefile
```diff
@@ -85,12 +86,17 @@ smoke-verify-runtime:
...
 	config=$$(mktemp); \
-	trap 'rm -f "$$config"' EXIT; \
+	nri=$$(mktemp); \
+	trap 'rm -f "$$config" "$$nri"' EXIT; \
 	$(DOCKER) exec "$$node" containerd config dump > "$$config"; \
 	grep -Eq 'enable_cdi[[:space:]]*=[[:space:]]*true' "$$config"; \
 	grep -Eq "default_runtime_name[[:space:]]*=[[:space:]]*['\"]nvidia['\"]" "$$config"; \
 	grep -Eq 'runtimes\.nvidia' "$$config"; \
 	grep -Eq "BinaryName[[:space:]]*=[[:space:]]*['\"]/usr/bin/nvidia-container-runtime['\"]" "$$config"; \
+	awk '$$0 == "[plugins.\"io.containerd.nri.v1.nri\"]" { active = 1; next } \
```
ArangoGutierrez: containerd 2.2.0's `config dump` prints this table indented as `[plugins.'io.containerd.nri.v1.nri']`, with `socket_path = '/var/run/nri/nri.sock'` single-quoted, so this exact-match awk extracts nothing and both greps fail. kind-node-publish runs this target on the merge push and `publish` needs `smoke-test`, so the NRI-enabled node image would not publish; allowing leading whitespace and either quote style, as the `default_runtime_name` check does, fixes it.

docs/guides/compute-domain/run.sh
```diff
@@ -350,7 +350,6 @@ helm upgrade --install "${RELEASE_NAME}" "${REPO_ROOT}/${CHART_PATH}" \
   --set image.repository=nvml-mock \
   --set image.tag=compute-domain \
   --set gpu.profile=gb200 \
-  --set nri.enabled=true \
   --set imex.mockChannels.enabled=true \
```
ArangoGutierrez: The ComputeDomain demo workload opts in only with `nvml-mock.nvidia.com/imex-channels`, and an IMEX-only container now gets its channel nodes but no overlay, PATH or topology env. `assert_clique` then fails at `kubectl exec ... check-fabric`, because that binary exists only in the overlay and the demo image does not ship it. Adding `nvml-mock.nvidia.com/devices: "true"` to demo-workload.yaml, and updating the README passages that say NRI supplies the overlay and topology identity, keeps the guide runnable.

pkg/gpu/mocknvml/engine/engine_test.go
```diff
@@ -509,6 +509,11 @@ func TestEngine_DeviceGetHandleByPciBusIdInvalid(t *testing.T) {
 
 // --- Visibility filtering tests ---
 
+func TestDetectVisibleDevices_ExplicitNone(t *testing.T) {
+	t.Setenv("MOCK_NVML_VISIBLE_DEVICES", "none")
+	require.Empty(t, detectVisibleDevices(&Config{NumDevices: 4}))
```
ArangoGutierrez: `require.Empty` accepts nil, and `detectVisibleDevices` returns nil (every GPU visible) on a host with no /dev/nvidiaN, so this test still passes with the `none` guard in engine.go removed. `require.NotNil` plus `require.Len(t, got, 0)` pins the difference between none and all.

docs/guides/with-gpu-operator/gpu-operator-values.yaml
```diff
@@ -19,7 +19,7 @@
 # Prerequisites:
 #   - the nvml-mock chart already installed (it stages the mock NVML and CUDA
 #     libraries and labels the node nvidia.com/gpu.present=true)
-#   - containerd on every node in CDI mode with the nvidia runtime handler
+#   - containerd on every node in CDI and NRI mode with the nvidia runtime handler
```
ArangoGutierrez: The header says this file matches tests/e2e/gpu-operator-values.yaml and local/gpu-operator/gpu-operator.values.yaml, but only those two got the `daemonsets.annotations` block. If the annotation stays, this file, the Step 3 values in gpu-operator.md and the nv-sentinel mirror need it too; with the IB detection above fixed, none of the three should need it.

ArangoGutierrez:
One more that turned up while running this on kind, and it predates this PR: `mountOverlay` binds the overlay and its config dir with no propagation option, so an injected container that also has a Bidirectional volume (the DRA kubelet plugin's `gpus` container) pushes the config bind back to the node, and the next container's rbind copies the stack. Node mounts at /var/lib/nvml-mock/driver/config went 0, 1, 3, 7, 15 over four such pod creations, the same on main with nri.enabled=true, and they stay after the pods are gone. With NRI on by default and the kubelet plugin annotated, every plugin restart now doubles it, so `rslave` on those two mounts (untested) may be worth folding in here or tracking separately.

PR #885, opened 2026-09-18
feat(imex): stage userspace through the node agent

### LINKED ISSUE #821: Unify Mokka OCI images
There are two Mokka images right now:

- https://github.com/NVIDIA/k8s-test-infra/blob/main/deployments/nvml-mock/Dockerfile
- https://github.com/NVIDIA/k8s-test-infra/blob/main/deployments/nvml-mock/Dockerfile.compute-domain-daemon

There should be no reason for Dockerfile.compute-domain-daemon to exist. Mokka should provide a unified image that includes all capabilities that you can potentially enable in Mokka, so it's easy to use and spin up without thinking which docker image should I use in this case.

olegleyz:
## What This PR Does

- extends the node-agent staging model to install pinned, checksum-verified IMEX userspace in the host-mounted mock driver tree
- runs the upstream DRA v0.5.0 image unchanged; its CDI edits supply the driver footprint, while NRI adds only the real IMEX executable and mock topology they do not provide
- downloads the architecture-specific archive at runtime and caches it on the node, keeping those files out of Mokka's published image
- removes both ComputeDomain-specific Dockerfiles and uses the standard Mokka image for regular and ComputeDomain installations
- adds peer join/loss coverage and an upstream DRA readiness, minimal-injection, and restart-recovery CI leg

## Why

ComputeDomain simulation should preserve the same component boundary as a real
cluster. On a real node, IMEX userspace is node software, commonly installed
alongside the NVIDIA GPU driver stack. The DRA `compute-domain-daemon` consumes
that host-provided userspace and starts `nvidia-imex`; it should not require a
specialized daemon image with another copy of IMEX.

Mokka already follows this node-level model for the mock GPU driver: its
privileged node agent materializes a driver tree on a host-mounted path. This PR
stages IMEX in that same tree. The node agent obtains IMEX from the pinned
official archive at runtime instead of embedding it in Mokka's image, and the
small Mokka shim adds `--nogpu` for the simulated node.

The DRA-generated CDI edits already expose the mock driver, IMEX shim and CLI,
domain config, and device nodes to `compute-domain-daemon`. NRI therefore adds
only `/usr/bin/nvidia-imex.real`, the mock topology document, and its environment
pointer. It identifies the target using both the DRA-owned
`resource.nvidia.com/computeDomain` pod label and the
`compute-domain-daemon` container name. A namespace-wide rule would alter
unrelated NVIDIA containers; the resource-claim CDI reference is not visible to
NRI until after `CreateContainer`.

The specialized adjustment requires both node-staged files before admitting
the daemon. If either is missing, only that container creation is rejected and
kubelet retries after the node agent converges. Other workloads retain Mokka's
existing fail-open behavior.

The result is closer to real-cluster startup and recovery: node software lives
in the node driver tree, the upstream DRA image remains unchanged, and Mokka
uses one OCI image instead of separate demo and DRA variants.

The first userspace staging needs access to the pinned archive URL. Subsequent
staging uses the checksum-addressed node cache.

Closes #821

## Review Map

The commits are intentionally separated by concern:

1. final node-agent staging and container-scoped NRI composition
2. removal of specialized images and local/Tilt wiring
3. peer lifecycle and upstream DRA end-to-end coverage
4. user and contributor documentation

## Validation

- `make test`
- `make helm-tests` (199 tests)
- `make docs`
- `go vet ./...`
- `golangci-lint run ./...`
- `make actionlint`
- `make test-nvidia-imex-shim`
- `make test-imex-userspace-download` (amd64 and arm64)
- `go test -tags e2e -run '^$' ./tests/e2e/go/...`
- targeted `imex-lifecycle` E2E: first peer not UP, second peer joins to UP, peer loss detected
- upstream DRA v0.5.0 E2E: 4/4 IMEX daemons ready, only the missing IMEX/topology files injected, and recovery after concurrent Mokka/daemon restart

## Checklist

- [x] Linter passes
- [x] Documentation updated
- [x] CHANGELOG.md updated


internal/nri/inject/computedomain.go
```diff
@@ -0,0 +1,56 @@
...
+		adjustment.Mounts = append(adjustment.Mounts, readOnlyFileMount(source, realIMEXContainerPath))
+	}
+
+	if topologyInjectable(cfg) {
+		adjustment.Mounts = append(adjustment.Mounts,
+			readOnlyFileMount(cfg.TopologyHostPath, cfg.TopologyContainerPath))
+		env := newEnvSet(container.Env)
+		env.setDefault("MOCK_TOPOLOGY_CONFIG", cfg.TopologyContainerPath)
+		adjustment.Env = append(adjustment.Env, env.changed()...)
+	}
```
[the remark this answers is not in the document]
olegleyz: Agreed. The specialized adjustment now requires both the staged real IMEX executable and topology before it returns any OCI edits. If either prerequisite is missing, NRI rejects only the recognized ComputeDomain daemon creation so kubelet retries after the node agent converges; unrelated containers retain fail-open behavior. The unit test covers neither, IMEX-only, and topology-only states, and the local upstream DRA acceptance test still passes 4/4 readiness plus concurrent restart recovery.

deployments/nvml-mock/helm/nvml-mock/values.yaml
```diff
@@ -322,6 +322,15 @@ imex:
...
     capsMajor: 236
+  # Stage the real IMEX userspace in Mokka's node driver tree. The node daemon
+  # resolves the current architecture through a repository lock, downloads the
+  # pinned archive from NVIDIA, verifies its SHA-256 checksum, and keeps a
+  # verified node-local cache for restart recovery. NRI recognizes the upstream
+  # DRA compute-domain-daemon by its pod label and container name and adds only the staged real
+  # executable and mock topology absent from the CDI edits, without changing
+  # its image. Off by default so ordinary Mokka installs perform no download.
+  userspace:
+    enabled: false
```
roma-glushko, line 325: I think it should be on by default at some point. What's the plan to flip it?

internal/agent/imex/imex.go
```diff
@@ -51,16 +94,51 @@ func (s *Simulator) Ready() bool { return s.ready.Load() }
 
 // Stage materializes the IMEX capability surface under host.Root.
 // It is a no-op (but marks ready) when state.IMEX.Enabled is false.
-func (s *Simulator) Stage(_ context.Context, state *agent.State) error {
+func (s *Simulator) Stage(ctx context.Context, state *agent.State) error {
 	s.ready.Store(false)
 	zap.L().Info("staging simulator", zap.String("simulator", name))
+	if !state.IMEX.UserspaceEnabled {
```
roma-glushko: Why that flag is called userspace? You can have kernel space IMEX? I don't think helps to understand that that flag tries to guard.

internal/agent/imex/userspace.go
```diff
@@ -0,0 +1,323 @@
...
+
+type userspaceInstaller struct {
+	root       string
+	shimPath   string
+	arch       string
+	lock       Lock
+	httpClient *http.Client
+}
+
+func (i *userspaceInstaller) stage(ctx context.Context) error {
```
roma-glushko, line 58: How big are IMEX artifacts?

roma-glushko (commented):
Checked very quickly and looks clean to me ✅ 

Thanks for rectifying how we deal with IMEX.

internal/agent/imex/userspace.go
olegleyz: For the pinned 595.91.07 artifacts, each node downloads only its architecture: 7,984,000 bytes (7.61 MiB) on amd64 or 7,367,088 bytes (7.03 MiB) on arm64. The three extracted files total 25.66 MiB on amd64 and 25.13 MiB on arm64. Because the verified compressed archive remains in the node cache, the steady cache-plus-staged footprint is about 33.27 MiB or 32.15 MiB per node, respectively. The archive is downloaded once per node and reused across node-agent restarts.

deployments/nvml-mock/helm/nvml-mock/values.yaml
olegleyz: Agreed. Common AWS accelerated AMIs and GPU Operator-managed driver installations include IMEX in the node driver footprint regardless of whether a ComputeDomain is active. I changed imex.nodeSoftware.enabled to default to true, with false as the explicit opt-out for restricted or no-egress installations.

internal/agent/imex/imex.go
olegleyz: Agreed. userspace described an implementation boundary rather than the behavior being controlled. I renamed the setting to imex.nodeSoftware.enabled, along with the corresponding node-agent signal, to make it clear that it downloads and stages IMEX node software.

deployments/nvml-mock/helm/nvml-mock/values.yaml
olegleyz: Follow-up: after discussing the runtime dependency with Eduardo and Giulio, I switched the default back to `false`. Enabling it globally would make every Mokka installation download IMEX at node-agent startup, including installations that do not use ComputeDomain. The ComputeDomain guide, local profile, and relevant E2E scenarios enable it explicitly. We can reconsider the default if IMEX can be provided without adding that runtime network dependency.

roma-glushko: > I switched the default back to false. Enabling it globally would make every Mokka installation download IMEX at node-agent startup, including installations that do not use ComputeDomain.

@olegleyz I have got the concern here, but I'm still thinking whether this flag is a way to go. Node agent knows already whether we need to mock IB/IMEX based on the provided configuration. So technically we could automatically infer whether to fetch it right now or not. That would be the best middle ground from my side.

>  I changed imex.nodeSoftware.enabled to default to true, with false as the explicit opt-out for restricted or no-egress installations.

Yes, airgapped is an interesting case to support. We want federal clusters to use Mokka, so at this point I'm just generally curious what GPU operator does/expects in airgapped envs. I think in that we can expect IMEX binary to be already downloaded and shared with out somehow e.g. a shared PV.

 In any case, we should split airgapped support into another, separate, focused task.

internal/agent/imex/imex.go
```diff
@@ -36,11 +39,51 @@ type Simulator struct {
...
 
 // New returns an imex Simulator.
-func New(h *host.Host) *Simulator {
-	return &Simulator{host: h, procDevicesPath: "/proc/devices"}
+func New(h *host.Host, options ...Options) *Simulator {
+	opt := Options{
+		Architecture: runtime.GOARCH,
+		ShimPath:     "/usr/local/bin/nvidia-imex-shim",
+		Lock:         mustDefaultLock(),
+		HTTPClient:   &http.Client{Timeout: 5 * time.Minute},
```
roma-glushko, line 59: This timeout is worth externalizing as a env var. Plus, we don't have 5min for sure. K8s would kill/restart it much earlier because of lack of readiness.

internal/agent/imex/userspace.go
roma-glushko: it would probably take much less than a second to pull it in. That's great.

internal/agent/imex/userspace.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+func (i *userspaceInstaller) downloadArchive(
+	ctx context.Context, artifact Artifact, cacheDir, archive string,
+) (string, error) {
+	downloadURL, err := url.JoinPath(i.lock.BaseURL, artifact.Path)
+	if err != nil {
+		return "", fmt.Errorf("build download URL: %w", err)
+	}
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, downloadURL, nil)
```
roma-glushko, line 142: While doing external, cross-cluster requests like this I would worry about retries and interim conn issues.

internal/agent/imex/imex.go
roma-glushko: I think we still have userspace entry used all over the places like internal/agent/imex/userspace.go. To me, it's just IMEX binary installation. So I would call it `internal/agent/imex/installer.go` or `internal/agent/imex/downloader.go`, for example. It might convey more direct meaning for what it's responsible for.

internal/nri/inject/adjust.go
```diff
@@ -17,18 +17,25 @@ import (
...
 	cfg = withDefaults(cfg)
 	if reason, skipped := skip(cfg, container); skipped {
 		zap.L().Debug("skipping container injection", zap.String("namespace", container.Namespace), zap.String("reason", reason))
-		return Adjustment{}, false
+		return Adjustment{}, false, nil
 	}
 
 	var adjustment Adjustment
+	if computeDomainDaemon(container) {
+		if err := adjustComputeDomain(cfg, container, &adjustment); err != nil {
```
ArangoGutierrez, line 34: Under chart defaults both prerequisites are absent: imex.nodeSoftware.enabled is false, so nvidia-imex.real is never staged (the agent actively discards it), and topology.enabled is false too. A compute-domain-daemon pod is then rejected on every CreateContainer rather than transiently, which contradicts docs/components/nri-plugin.md line 146 promising kubelet retries "after the node agent converges" - convergence never comes. The gate was written in 4fe12be3 when staging was unconditional, and d0ef01df flipped the default without revisiting it. Plumbing the staging flag into inject.Config and hard-failing only when node software is expected keeps the retry semantics the doc describes.

local/compute-domain/scenarios/check-dra-imex.sh
```diff
@@ -0,0 +1,131 @@
...
+# Kubernetes must converge after Mokka restores the stable driver tree; no
+# custom DRA image or manual node preparation participates in the recovery.
+printf 'Restarting Mokka and the per-domain daemon fleet together\n'
+k -n "${MOKKA_NAMESPACE}" delete pod \
+  -l app.kubernetes.io/name=nvml-mock --wait=false >/dev/null
+k -n "${DRIVER_NAMESPACE}" delete pod \
+  -l "resource.nvidia.com/computeDomain=$(k get computedomain "${DOMAIN}" -o jsonpath='{.metadata.uid}')" \
+  --wait=false >/dev/null
+
+k -n "${MOKKA_NAMESPACE}" rollout status daemonset/nvml-mock --timeout="${TIMEOUT_SECONDS}s"
```
ArangoGutierrez, line 126: delete pod --wait=false leaves the DaemonSet generation and spec untouched, so the rollout status calls below it can succeed on their first Get against status that still describes the pre-delete pods, and domain_ready and minimal_nri_adjustment can both be satisfied by the old pods while they are still Terminating and still exec-able. The recovery half of this gate can therefore pass without any restart having happened, and it is what the new e2e-compute-domain-dra job checks. Waiting for the old pods to disappear, or capturing their UIDs and asserting they changed, would make it real.

internal/agent/imex/userspace.go
```diff
@@ -0,0 +1,323 @@
...
+	if err != nil {
+		return "", fmt.Errorf("create download request: %w", err)
+	}
+	resp, err := i.httpClient.Do(req)
```
ArangoGutierrez, line 146: This is a single unretried GET, and any transport error fails imex Stage, which marks the whole node agent Unhealthy and skips the Apply wave for the cdi, gpudriver and pcibus simulators. One network blip therefore stops CDI spec publication and takes GPU mocking down on that node until the liveness probe restarts the pod; FileSource.poll only re-emits when the config bytes change, so nothing retries in-process. A bounded retry with backoff, and scoping the failure to the IMEX surface, would keep the blast radius on the feature that needs the network.

ArangoGutierrez (changes requested):
Thanks for this, staging the IMEX userspace on the node is what lets the upstream compute-domain-daemon image run unmodified and retires the two forked daemon Dockerfiles.

The node-staged model is the right boundary here, and the acquisition path is careful with it: the digest is checked before the archive is promoted into the cache, the cache is re-verified on every reuse, and extraction is allowlist-first. Flipping the default in d0ef01df needs one follow-up in the NRI gate, inline.

- The pinned archive also carries LICENSE and usr/share/doc/third-party-notices.txt next to the three members staged here, and neither is copied into the driver tree. Staging them alongside the binaries keeps the notice files with the software they cover. (internal/agent/imex/userspace.go:37)
- validateBaseURL accepts http as well as https for the URL that yields a binary the node later executes. The embedded lock is https today, so narrowing the accepted scheme costs nothing. (internal/agent/imex/lock.go:78)

internal/nri/inject/computedomain.go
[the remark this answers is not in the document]
ArangoGutierrez: Both prerequisites are hard preconditions here: adjustComputeDomain returns an error when nvidia-imex.real is not a regular file, and again when the topology is not staged, with the first mount append happening only after both checks. Adjust propagates that error out through CreateContainer, so containerd fails the request and kubelet retries instead of latching a partial adjustment. TestAdjustComputeDomainRejectsPartialNodeStaging covers the real-IMEX-only and topology-only cases, asserting an error and an empty adjustment in each.

internal/nri/inject/adjust.go
```diff
@@ -17,18 +17,25 @@ import (
...
+func Adjust(cfg Config, container Container) (Adjustment, bool, error) {
 	cfg = withDefaults(cfg)
 	if reason, skipped := skip(cfg, container); skipped {
 		zap.L().Debug("skipping container injection", zap.String("namespace", container.Namespace), zap.String("reason", reason))
-		return Adjustment{}, false
+		return Adjustment{}, false, nil
 	}
 
 	var adjustment Adjustment
+	if computeDomainDaemon(container) {
```
roma-glushko, line 33: Does it make sense to keep Adjust() for regular workloads and do checks for specific workloads like compute domain daemon on one level above where Adjust is applied?

Compute Domains are potentially not the only specific domain that we might need to intercept.

